PR #1893, opened 2023-01-23
brian.lynn_auth0_606245e5d39a5a00700d27bc-CollateralInterestContribution

regnosys-prod-user:


tomhealey-icma (approved):

tomhealey-icma:
As these changes are limited to the csa and isda-create I have no requested changes but I would suggest avoid using attribute names that include specific domain references, 

enum AlternativeToInterestAmountEnum: <"If there is an alternative to interest amounts, how is it done?">
    CSA_Standard   <" TBD ">
    CSA_NonStandard   <" TBD ">
    Not_Applicable  <" TBD ">

It would be better to not include CSA in the attribute but if necessary use a Source Attribute /ENUM reference.

rosetta-source/src/main/rosetta/mapping-isdacreate-synonym.rosetta
lolabeis, line 1: I see that we are removing ISDA Create synonyms for `DistributionAndInterestPayment`, which makes sense given that we are replacing this type with a new structure. We should address the following questions:

- Do we know what document type(s) will be affected, and does the Collateral WG mind?
- What about CreateIQ and its users?
- If users do mind, do all the attributes that we were mapping to in the old structure have a target in the new one?

Based on the answer to those questions, we can formulate a plan. An interim delivery would maintain both the old and new types in parallel, and also maintain all the synonyms for the former, while marking the old type as `deprecated`. In a second step, we would remap ISDA Create attributes to the new type and then retire the old.

rosetta-source/src/main/rosetta/legaldocumentation-csa-type.rosetta
```diff
@@ -529,7 +536,7 @@ type DisputeResolution: <"A class to specify the election terms under which a pa
...
 	recalculationOfValue RecalculationOfValue (0..1) <"The elections to specify terms for recalculation of the market value of posted collateral.">
 
-type DistributionAndInterestPayment: <"A class to specify the Distributions and Interest Payment provisions applicable to the Japanese Law ISDA 2016 CSA for Initial Margin and the New York Law ISDA 2016 CSA for Variation Margin. ISDA 2016 Japanese Law Credit Support Annex for Initial Margin, paragraph 13, General Principles, (n): Distributions and Interest Payment (IM). | ISDA 2016 Credit Support Annex for Variation Margin, paragraph 13, (i): Distributions and Interest Payment (VM).">
+type DistributionAndInterestPaymentOld: <"A class to specify the Distributions and Interest Payment provisions applicable to the Japanese Law ISDA 2016 CSA for Initial Margin and the New York Law ISDA 2016 CSA for Variation Margin. ISDA 2016 Japanese Law Credit Support Annex for Initial Margin, paragraph 13, General Principles, (n): Distributions and Interest Payment (IM). | ISDA 2016 Credit Support Annex for Variation Margin, paragraph 13, (i): Distributions and Interest Payment (VM).">
```
lolabeis, line 539: See comment re synonyms

rosetta-source/src/main/rosetta/legaldocumentation-csa-type.rosetta
```diff
@@ -529,7 +536,7 @@ type DisputeResolution: <"A class to specify the election terms under which a pa
...
 
 	interestRate EligibleCurrencyInterestRate (1..*) <"The interest rate associated with initial or variation margin collateral, depending upon the type of credit agreement that this election is associated with. ISDA 2016 Japanese Law Credit Support Annex for Initial Margin, paragraph 13, General Principles, (n)(i): Interest Rate (IM). | ISDA 2016 Credit Support Annex for Variation Margin, paragraph 13, (i)(i): Interest Rate (VM).">
```
lolabeis, line 541: There are other types to be deprecated, that are only used as attributes in this type:

- `EligibleCurrencyInterestRate`
- `InterestAmount`
- `InterestAdjustment`

lolabeis (changes requested):

iansloyan:
A redesign is to be proposed by @brianlynn2 - I think it is appropriate to close this PR now

PR #1953, opened 2023-02-15
ben.aubry_auth0_60be7cbc9d409a0069cd0f72-BondReference

regnosys-prod-user:


iansloyan (changes requested):
Hi @ben-aubry thank you for your contribution - was there any benefit considered in this design to using the _Price_ type (or even _FixedRateSpecification_ type) instead of _number_ type?

ben-aubry:
I think we could use _FixedRateSpecification_ type instead of number.

iansloyan:
Sounds like a plan to me. Thanks @ben-aubry

iansloyan:
superseded by https://github.com/finos/common-domain-model/pull/1999

PR #1965, opened 2023-02-21
Update CVE allow list to exclude false positives for Eclipse libraries

hugohills-regnosys:


maoo:
LGTM, great work @hugohills !

minesh-s-patel (approved):

PR #1987, opened 2023-02-26
jbziade_auth0_5fe1be69c40973007616a588-TradeExtraordinaryEvents

regnosys-prod-user:


rosetta-source/src/main/rosetta/base-staticdata-asset-common-enum.rosetta
```diff
@@ -5,8 +5,9 @@ enum AssetClassEnum: <"The enumerated values to specify the FpML asset class cat
 	Commodity <"Commodity.">
 	Credit <"Credit.">
 	Equity <"Equity.">
-	ForeignExchange <"ForeignExchange.">
-	InterestRate <"InterestRate.">
+	ForeignExchange <"Foreign Exchange.">
+	InterestRate <"Interest Rate.">
+    digitalAsset <"Digital Asset.">
```
rvincentISDA, line 10: Let's keep the addition of digitalAsset in a future contribution.

JBZ-Fragmos: image

JBZ-Fragmos: looks fully consistent to me

rvincentISDA: Pointing to the existing Digital Assets Definitions document is ok. Introducing the nomenclature of digitalAsset as an asset Class in the model is not something that has been discussed yet. For example a BTC forward is a FX asset class or a digital asset class ? Note that it is not featured in FpML scheme either.  I agree this is important but it is premature to tag here.

JBZ-Fragmos: I beleive the question "is BTC an FX asset class or a digital asset class" is totally another question than "is digital asset per se an asset class" - first question is about whether an item belongs to a set (e.g. is "2x3" a pair number" ?) the second one is about the set itself (for instance the set made of all "pair numbers", defined as "2n")

and you could revert your own example : notwithstanding the fact we may not know whether BTC an FX asset class or a digital asset class", does it make you have a doubt about the fact FX is a type of Observable ? I guess no, so why would it cause a doubt about the fact Digital Asset is an... "Asset" as its own name says it... 

and what else do we need to decide if "X" is an asset class, else than dedicated ISDA Definitions which header is explicitly refering to "X" as being an "(Digital) Asset" ?

JBZ-Fragmos: that being said, I agree with you this topic is not the same as Additional Terms..

so may I suggest we list this one as a distinct item for the next Committee, ask if there is any objection in that regards i.e. to add Digital Asset to the Enum list of Asset Class, and if we could move forward to ensure minimum adjustments to start including it inside the model ?

for now the only impact I can see elsewhere in the model is that we would also need to add it as a new attribute of Product type e.g. as below (in terms of attribute, only using ProductBase type would be sufficient to beging with) : for instance Index is defined the same way, with a minimum of atttributes

 - for clarity there is no mapping impact, if you try to make this change in your workspace you can see no "break point" happens anywhere..

image

rvincentISDA: Reviewing the  rest of the contribution I can see that it is 393 releases in the past. This is going to be a struggle.
Next steps:
i) Resubmit a fresh contribution with the feature for extraordinary events. That part was agreed at the ARC. (I will connect offline)
ii) Bring the discussion on the requirements for features related to digital asset to the working groups for further discussion

rvincentISDA:
I will close this PR as any further change will be addressed elsewhere.

PR #2006, opened 2023-03-03
DSL update: remove Tycho

SimonCockx:


hugohills-regnosys:
Replaced by #2009 and #2010

PR #2021, opened 2023-03-09
substitution-event

tomhealey-icma:
This PR adds func Create_SubstitutionPrimitiveInstruction, func Create_SubstitutionInstruction and func Qualify_Substitution for repo trades. A substitution event occurs when a counterparty  replaces the collateral on existing repo trade. Substituting collateral may result in a new quantity of bonds and also a new cash amount. All other terms shall remain the same. As such, the proper processing of a substitution event is to close the original trade and recreate one with the new collateral.


tomhealey-icma:
Removing PR. To be resubmitted.

PR #2040, opened 2023-03-14
g.melpignano_auth0_64104580fb05da1a3254ae73-Test_Reply

regnosys-prod-user:


rosetta-source/src/main/rosetta/event-common-type.rosetta
```diff
@@ -257,6 +257,7 @@ type Trade: <"Defines the output of a financial transaction between parties - a
 	tradeIdentifier TradeIdentifier (1..*) <"Represents the identifier(s) that uniquely identify a trade for an identity issuer. A trade can include multiple identifiers, for example a trade that is reportable to both the CFTC and ESMA, and then has an associated USI (Unique Swap Identifier) UTI (Unique Trade Identifier).">
 	tradeDate date (1..1) <"Specifies the date which the trade was agreed.">
         [metadata id]
+    dateOfTrade date (1..1)
 	tradableProduct TradableProduct (1..1) <"Represents the financial instrument The corresponding FpML construct is the product abstract element and the associated substitution group.">
```
chrisisla, line 261: Why do we need dateOfTrade when we have tradeDate? Plus it is mandatory?

PR #2048, opened 2023-03-15
Repo partial-delivery

tomhealey-icma:
This PR adds partial delivery function and qualification from ICMA Phase 2 CDM for Repo and Bonds.

### DIFF
--- a/rosetta-source/src/main/rosetta/event-common-func.rosetta
+++ b/rosetta-source/src/main/rosetta/event-common-func.rosetta
@@ -1661,6 +1661,87 @@ func Qualify_Shaping: <"The qualification of a shaping event from the fact that 
 		and packageRef count = openTradeNoExecutionDetails count
 		and packageRef distinct count = 1
 
+
+func Create_PartialDeliveryPrimitiveInstruction: <"Creates the primitive instructions for a substitution of collateral by replacing the assetpayout of the trade.">
+	inputs:
+		//ICMA-P2 Instructions
+		tradeState TradeState (1..1) <"The original trade to be closed.">
+		effectiveDate AdjustableOrRelativeDate (1..1) <"The date to close and open a new position.">
+		terminationDate AdjustableOrRelativeDate (1..1) <"The new termination date.">
+		deliveredPriceQuantity PriceQuantity (1..*) <"The price and quantity of the remaining trade. ">
+		undeliveredPriceQuantity PriceQuantity (1..*) <"The price and quantity of the remaining trade. ">
+	output: 
+		instruction PrimitiveInstruction (1..1)
+
+	condition ContractualProduct: <"Only a contractual product can be rolled.">
+		tradeState -> trade -> tradableProduct -> product -> contractualProduct exists
+
+	set instruction -> split -> breakdown: <"Sets the first part of the split to be a termination instruction for the existing trade.">
+		[ Create_PrimitiveInstruction(
+			empty, empty, empty, empty,
+						Create_QuantityChangeInstruction(
+						deliveredPriceQuantity, 
+						QuantityChangeDirectionEnum -> Replace, 
+						tradeState->trade->tradableProduct->tradeLot->lotIdentifier
+			),
+			empty, empty, empty, empty,	empty, empty, empty)]
+
+	add instruction -> split -> breakdown: <"Sets the second part of the split to be a new contract with a new asset payout.">
+		[
+		Create_PrimitiveInstruction(
+			empty, empty, empty, empty,
+			Create_QuantityChangeInstruction(
+				undeliveredPriceQuantity,
+				QuantityChangeDirectionEnum -> Replace,
+				empty
+				),
+			empty, empty,
+			Create_PartialDeliveryInstructionTermsChange(
+				tradeState -> trade -> tradableProduct -> product -> contractualProduct,
+				effectiveDate, terminationDate
+				),
+			empty, empty, empty, empty
+			)
+		]
+
+func Create_PartialDeliveryInstructionTermsChange:
+	inputs:
+		contractualProduct ContractualProduct (1..1)
+		effectiveRollDate AdjustableOrRelativeDate (1..1) <"The date to close and open a new position.">
+		terminationDate AdjustableOrRelativeDate (1..1) <"The new termination date.">
+	output:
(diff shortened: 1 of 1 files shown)


tomhealey-icma:
Additional documentation on the business scenario. 
[CWG Meeting 2023-03-21.pptx](https://github.com/finos/common-domain-model/files/10983429/CWG.Meeting.2023-03-21.pptx)

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -1661,6 +1661,87 @@ func Qualify_Shaping: <"The qualification of a shaping event from the fact that
 		and packageRef count = openTradeNoExecutionDetails count
 		and packageRef distinct count = 1
 
+
+func Create_PartialDeliveryPrimitiveInstruction: <"Creates the primitive instructions for a substitution of collateral by replacing the assetpayout of the trade.">
+	inputs:
+		//ICMA-P2 Instructions
```
iansloyan, line 1667: What are ICMA-P2 Instructions? Is this an ICMA specific function?

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -1661,6 +1661,87 @@ func Qualify_Shaping: <"The qualification of a shaping event from the fact that
 		and packageRef count = openTradeNoExecutionDetails count
 		and packageRef distinct count = 1
 
+
+func Create_PartialDeliveryPrimitiveInstruction: <"Creates the primitive instructions for a substitution of collateral by replacing the assetpayout of the trade.">
```
iansloyan, line 1665: If the function is related to the substitution of collateral - why is it called "PartialDelivery"? Generally perhaps a more fulsome definition of the function's purpose would help - ready through the steps and inputs alone it seems to do something quite sophisticated?
_E.g. this instruction is to split a trade in two and change the first bit this way and second bit that way to effect this change overall_

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -1661,6 +1661,87 @@ func Qualify_Shaping: <"The qualification of a shaping event from the fact that
...
+			empty, empty,
+			Create_PartialDeliveryInstructionTermsChange(
+				tradeState -> trade -> tradableProduct -> product -> contractualProduct,
+				effectiveDate, terminationDate
+				),
+			empty, empty, empty, empty
+			)
+		]
+
+func Create_PartialDeliveryInstructionTermsChange:
```
iansloyan, line 1707: What is this function for? How is it different from any other terms change? Also it has no comments to explain what it is doing

iansloyan (commented):
Not sure I understand exactly what is happening here. Though there is a lot of new code and I am probably not reading correctly. I cannot see how the AssetPayout is specifically targeted to make this function only relevant to it.
Looking forward to discussing

rosetta-source/src/main/rosetta/event-common-func.rosetta
tomhealey-icma: A partial delivery occurs when the collateral giver is unable to deliver the full amount on settlement. The market best practice recommends buyers accept the amount available for delivery and may negotiate the remaining. So the workflow in the function is to modify the existing contract to accept the available delivered collateral with an option to open a new contract on the undelivered amount.

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -1661,6 +1661,87 @@ func Qualify_Shaping: <"The qualification of a shaping event from the fact that
...
+
+func Create_PartialDeliveryPrimitiveInstruction: <"Creates the primitive instructions for a substitution of collateral by replacing the assetpayout of the trade.">
+	inputs:
+		//ICMA-P2 Instructions
+		tradeState TradeState (1..1) <"The original trade to be closed.">
+		effectiveDate AdjustableOrRelativeDate (1..1) <"The date to close and open a new position.">
+		terminationDate AdjustableOrRelativeDate (1..1) <"The new termination date.">
+		deliveredPriceQuantity PriceQuantity (1..*) <"The price and quantity of the remaining trade. ">
+		undeliveredPriceQuantity PriceQuantity (1..*) <"The price and quantity of the remaining trade. ">
+	output:
```
chrisisla, line 1673: Do we need undeliveredPriceQuantity? If this is a partial delivery then you would just specify what has been delivered and the remainder would be calculated surely? Just wondering if we need both amounts to be specified.

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -1661,6 +1661,87 @@ func Qualify_Shaping: <"The qualification of a shaping event from the fact that
...
+		deliveredPriceQuantity PriceQuantity (1..*) <"The price and quantity of the remaining trade. ">
+		undeliveredPriceQuantity PriceQuantity (1..*) <"The price and quantity of the remaining trade. ">
+	output: 
+		instruction PrimitiveInstruction (1..1)
+
+	condition ContractualProduct: <"Only a contractual product can be rolled.">
+		tradeState -> trade -> tradableProduct -> product -> contractualProduct exists
+
+	set instruction -> split -> breakdown: <"Sets the first part of the split to be a termination instruction for the existing trade.">
+		[ Create_PrimitiveInstruction(
```
chrisisla, line 1680: Is this comment correct? Looks like a quantity change not a termination?

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -1661,6 +1661,87 @@ func Qualify_Shaping: <"The qualification of a shaping event from the fact that
...
+		contractualProduct
+	set termsChangeInstruction -> product -> contractualProduct -> economicTerms -> effectiveDate:
+		effectiveRollDate
+	set termsChangeInstruction -> product -> contractualProduct -> economicTerms -> terminationDate:
+		terminationDate
+		
+
+
+func Qualify_PartialDelivery: <"Qualification of a partial delivery which constitutes a contract close and reopen with a difference quantity and termination date.">
+	[qualification BusinessEvent]
```
chrisisla, line 1725: "different"
Again though I don't think this event should expect a close/open

Issue #2029, opened 2023-03-10
Release Schedule


nigelcobb:
A release schedule should be provided for the CDM giving detail to implementors on current supported versions of the model and when future versions are expected to be released.

eteridvalishvili:
Hi @nigelcobb The PRs linked to this issue have both been resolved so I am closing. Please feel free to reopen if necessary. Thank you!

eteridvalishvili:
Hi @nigelcobb and @lolabeis Closing this issue. Please reopen if necessary.

PR #2056, opened 2023-03-17
Proposal for Release Schedule communication

nigelcobb:
Re issue #2029 - proposal for communicating a CDM to release schedule to members. @minesh-s-patel @chrisisla as discussed.


tomhealey-icma:
Do we still need this PR?

chrisisla:
Commenting on this issue as requested at the last SWG, Issue #2341 

I'm going to assume we are talking about Production Releases here i.e. N.N.N versions.

The main questions I have on the release schedule are:
- Who decides when a Production Release should be created?
  - Should this be the maintainers, or the members of the SWG, or both (assuming there are people who are in the SWG who are not maintainers or vice-versa)?
- What determines whether the release is a major or a minor release? 
  - Is it just when there is a breaking change we have a new major version?

At the moment there does not appear to be any communication/decision made by the maintainers about when a production release is required.

chrisisla:
Questions from the powerpoint in the previous comment:

> Should the release schedule be criteria driven (e.g. a certain number of PRs, or a certain number of PRs of a specific complexity) or time driven (e.g. aim to release twice a year)?

I do not recommend time driven releases. Major releases should be created as infrequently as possible i.e. only when there is a breaking change. Minor releases should be created when functionality in the current development version is required for production implementations of the model. 

> What should the cadence/relationship be between dev releases and production releases?

Think this is related to the first question really? Dev releases can be created for every commit made to the model, or can include several commits in one. There is no real relationship between a dev or a production release, except that a production release is made up of a 1 to many dev releases. 

> What should be the governance process, timeline and responsible actors for the release process?

My feeling is that the SWG should decide in conjunction with the maintainers when a production release should be created. This would probably need to be raised as an agenda item at the SWG meeting, but when a new release is required sooner than the next meeting then an ad-hoc meeting could be requested. In general it would be anticipated that a major release would be known well in advance of when it needed to be created. If we start defining roadmaps then we should know several months in advance of releases.
Currently releases are physically executed by REGnosys as a trusted technology partner. This could be expanded to include other CDM maintainers as well, but they would have to have the correct skillset - the release process has to be handheld and if something fails the person who is handholding the process needs to know how to fix it. This level of knowledge will need some time to acquire.

> What should the governance process be should an ad-hoc release be required?

Escalating to the maintainers, followed by an ad-hoc meeting of the SWG to approve the release. This should only be required for patch releases, not major or minor releases.

> What are the necessary steps and proposed timeline to share knowledge of, configuration of and responsibility for the maintenance aspects of the release and build processes? E.g movement from CodeFresh to Github Actions

First step would be to setup a small taskforce to help migrate the release process from Codefresh to Github Actions. Once that is done then the process and troubleshooting steps can be documented by the same group. After that we can look into expanding the roles and responsibilities a bit more. 

> What is the process to determine/add which code distributions are included in the Build/Release process?

This should be decided between the TAWG and the SWG. The TAWG can help determine the impact of including specific artefacts in the distribution, and then the SWG can decide whether said items should be in the distribution going forward. As part of this question we also need to consider the ongoing maintenance of anything that is put into the distribution i.e. if something added to the distribution fails during the build process who is responsible for fixing it?

tomhealey-icma (changes requested):
I'm not in favor of long release cycles especially where no significant value is added in the production release process.

tomhealey-icma:
Closing the PR. It's outdated.

PR #2060, opened 2023-03-21
Task 5015: Addition of Clearing Receipt Date Time in EventTimestampQualificationEnum in CDM

hugohills-regnosys:


tomhealey-icma (changes requested):
I question the use of enumerations for the purpose of qualify event workflow timestamps. In this specific case, the presence of a clearing event receipt and its timestamp should be sufficient. Already there's a growing list a workflow specific timestamp enums that appear to be duplicative if the event also exists.

tomhealey-icma:
I still have a concern over expanding lists of enumerations to define events. In this scenario, there should be clearly defined events for each step in the workflow that would include date and timestamp.

hugohills-regnosys:
Thanks @tomhealey-icma.  Yes, agreed on both points.  

I have raised two corresponding issues:
- https://github.com/finos/common-domain-model/issues/2072 - Enums should be in UpperCamelCase
- https://github.com/finos/common-domain-model/issues/2073 - Review EventTimestampQualificationEnum

We could discuss briefly on Monday if you have time.

tomhealey-icma (approved):

PR #2063, opened 2023-03-22
valdensmith_auth0_5c8a2a7821f07405a605d820-EligibleCollateralNameChange

regnosys-prod-user:


tomhealey-icma (commented):
There's a function, EligibleCollateralScheduleHelper that contains an attribute schedule EligibleCollateralSchedule. It looks like this change would impact backward compatibility?

iansloyan:
> There's a function, EligibleCollateralScheduleHelper that contains an attribute schedule EligibleCollateralSchedule. It looks like this change would impact backward compatibility?

this is a good point - @valdensmith - should this renaming propagate to other elements of the model?

tomhealey-icma (approved):

hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2090

PR #2064, opened 2023-03-23
Making the transferExpression field optional

minesh-s-patel:
Changing the cardinality of transferExpression from (1..1) to (0..1) - making this field optional when creating a transferInstruction


tomhealey-icma (commented):
What was the reason for qualifying a transfer type as an Expression when it holds a distinct list of enumeration types? Why not just TransferType?

ben-aubry:
@tomhealey-icma - not sure why this was initially done this way. Currently the way it is implemented, is that transferExpression has priceTransfer and scheduledTransfer (which both contain enums) - wouldn't it make sense to consolidate everything under something called "transferType" which would contain one consolidated enum (containing all the values from the other enums). This would replace the current transferExpression field under "Transfer" ?

iansloyan:
PriceTransfer is not a great name at all is it. It is of type FeeType which makes me wonder if the naming went awry.

I think the idea was to separate transfers which were related to some transaction/negotiated payment.....from.....those that are lifecycle payments (_scheduled_) including some external market factor transfer like a corporate action or credit event (which are not always scheduled!).....then to identify with more precision if the type was corp action.

As discussed @ben-aubry I think that approach of adding a value of "unknown" to the enum is the most appropriate.

tomhealey-icma:
I also question why it's referred to as an Expression. It contains a enumeration list and schedule. Is there a definition of the modifier "Expression" and what it means?

PR #2071, opened 2023-03-24
Update DSL version to 7.3.1 and bundle version 5.9.0

hugohills-regnosys:
- No model or test expectation changes
- DSL version updated, all versions backwards compatible:
  - `7.3.1` Fix Java code-gen bug related to extracting `date` from `zonedDateTime` record type
  - `7.3.0` Add support for external rule reference
  - `7.2.1` Code-gen generated Java that does not contain keyword clashes
- Maven dependency cleaned up - transitive dependencies removed (especially xtext)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,17 +1,15 @@
-# Product Model - FpML synonym mappings for price and quantity
-
-_Background_
-
-This release updates and extends the FpML mapping coverage for the product model.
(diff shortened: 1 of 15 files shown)


minesh-s-patel (approved):

iansloyan (approved):
Having reviewed the code and purpose I believe this can be released!

PR #2080, opened 2023-03-30
Add executionDateTime to Execution Details

tomhealey-icma:
ExecutionDetails type currently does not include an execution date and time, and Trade type only includes date type. This is needed for various regulatory reports and booking systems.

### DIFF
--- a/rosetta-source/src/main/rosetta/event-common-type.rosetta
+++ b/rosetta-source/src/main/rosetta/event-common-type.rosetta
@@ -472,7 +472,8 @@ type ExecutionDetails: <"Defines specific attributes that relate to trade execut
 	executionType ExecutionTypeEnum (1..1) <"Identifies the type of execution, e.g. via voice, electronically...">
 	executionVenue LegalEntity (0..1) <"Represents the venue on which a trade was executed.">
 	packageReference IdentifiedList (0..1) <"A reference to the package linking the trade with other trades, in case the trade was executed as part of a package (hence this attribute is optional).">
-		
+	executionDateTime zonedDateTime (0..1) <"The zoned date and time of the execution.">
+
 	condition ExecutionVenue: <"When the execution type is set to 'Electronically', the execution venue must be specified.">
 		if executionType = ExecutionTypeEnum -> Electronic
 		then executionVenue exists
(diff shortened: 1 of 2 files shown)


nigelcobb:
@tomhealey-icma this issue was discussed at a special session of the DRR Working Group a few weeks ago.
It would be useful to refer back to those materials as the conclusion reached at that discussion is not in line with your proposal.
The regulatory specific concept is served through the DRR model.  An Execution BusinessEvent has a timestamp associated with the WorkflowStep that creates that event so the Core CDM was also felt to be sufficiently expressive.

tomhealey-icma:
Thanks, @nigelcobb for the feedback but my view is that a trade is representative of an event and the attribute tradeDate date (1..1) is insufficient to capture what is a legal and regulatory element of that event. Most regulations today require date, time in microseconds and timezone. Rather than modify trade date I suggest adding zonedDateTime to execution details.

chrisisla:
I think this warrants a bit more conversation. 

I've only had a brief look but if you use Create_BusinessEvent for an execution then I don't think you get back an execution date/time. As Tom suggests it is a requirement for a lot of regulations (and applications) these days that the date, time to the ms and the timezone are all reported, but I don't see you get that currently from this function.

If you used WorkflowStep instead then there is an EventTimeStamp in it, but you'd have to use the WorkflowStep functions to get that returned...

Realistically we should consider how we can get the date/time/timezone returned back for any transaction during a trade lifecycle, not just for the initial execution (I think this is required for MiFID?). If the recommendation is that we always use the WorkflowStep functions so that "timestamp" can be used then we need to consider how this will impact existing usage of the Create_BusinessEvent function.

So in general I am more aligned with Tom on this, I think a generic zonedDateTime element (i.e. not just for executions) within the BusinessEvent itself would be beneficial.

@tomhealey-icma , @nigelcobb  - thoughts?

tomhealey-icma:
Just adding to my earlier notes, when thinking about the trade date/time it should not be the same as when a business event or workflow event took place. A workflow event is when something took place in the workflow process, but that may not be the same as the trade date/time as agreed by the counterparties which I believe becomes part of the trade contract.

PR #2084, opened 2023-03-31
Change collateralType CollateralTypeEnum from (1..1) to (0..1)

tomhealey-icma:
type CollateralProvisions attribute collateralType CollateralTypeEnum (1..1) is required but I this is better located in the collateral criteria. CollateralTypeEnum also only applies to security lending transactions:

enum CollateralTypeEnum: <"Specifies the types of collateral that are accepted by the Lender">
    Cash <"Security Lending Trades against Cash collateral">
    NonCash <"Security Lending Trades against NonCash collateral">
    CashPool <"Security Lending Trades against CashPool collateral">

I don't thing it's good design to have a generic data type apply to one specific market. For now I would just like to modify it's multiplicity from 1..1 to 0..1


valdensmith:
This structure CollateralProvison I believe was contributed by ISLA, the related enumerations CollateralTypeEnum is SFT specific. I understand the need that this collateral provisions added to a product could be generic. This structure and any related changes will require a wider discussion with ISLA and could be added to the agenda for the next CDM Collateral WG meeting. The related data type/attributes and enumeration may also need updating

rvincentISDA:
What is the rational for introducing the executionDateTime as part of the Execution Details. Is this duplicative of the timestamp of a WorkflowStep ?

rvincentISDA:
> 

In addition, If every attribute is optional in a type one should inspect if additional logical data constraints are necessary. If none this may imply that the grouping under that type is artificial and should be inspected.

tomhealey-icma:
Based on the discussion in the collateral working group on 4/25 the cardinality of collateralType CollateralTypeEnum (1..1) will be changed to (0..1). ISLA is to add their proposed conditional clause.

valdensmith:
In reference to last comment, it was not agreed on the Collateral WG that the cardinality of the type should be changed. The consensus was to explore is there is an alternative way to infer the collateral type from the details of the trade. This would need to be confirmed prior to changing any cardinality as there may be other impacts/ considerations for SFT.

tomhealey-icma:
Apologies if I misunderstood. I agree that a condition and qualification should be added to classify the collateral type but the requirement to include a security lending enum on all repo trades is too restrictive and blocking other work at the moment. Why not remove the requirement and add a condition for sec lending trades? You can re-use the sec lending qualification and convert it into a product type condition. Regarding the actual enum, it's not clear to me if it's a qualification of what collateral was provided or an eligibility of what collateral is allowed. If it's the latter it should be moved to eligibleCollateral. If it's a qualification, then I believe it can be determined from the assets and cash on the trade.

chrisisla:
I don't think the securities lending qualification will work anymore now that it uses assetPayout. I think it will not be able to categorically differentiate between a repo and stock loan. This is why in order to add a condition to make collateralType mandatory for securities lending trades we need to modify the qualification.
The CollateralTypeEnum describes the actual collateral used on the trade - so it qualifies the collateral that was provided. I don't think that the assets and/or cash on the trade are able to derive this, which is why I will be raising this at the next ISLA CDM WG meeting.

tomhealey-icma:
I believe a sec lending product should use 2 or more assetpayouts (collateral vs sec). InterestRatePayout is primarily for cash loans  and debt payouts. That said, you will need additional qualifiers, such as the legal agreement, taxonomy or separate from repo, sec lending doesn't use purchase date/price and repurchase date/price. As it stands the qualification would also match a repo:
func Qualify_SecurityLendingAgreement: <"Qualifies a product as a Security Loan.">
	[qualification Product]
	inputs: economicTerms EconomicTerms (1..1)
	output: is_product boolean (1..1)
	set is_product:
		economicTerms -> payout -> assetPayout exists and
		economicTerms -> payout -> interestRatePayout exists and
		economicTerms -> collateral exists

PR #2090, opened 2023-04-04
STORY-2078: Eligible Collateral Schedule root data type name change

hugohills-regnosys:
EligibleCollateralScheduleNameChange


valdensmith:
Hi I am trying to add release notes to this pull request but it will not allow me to process the changes to the release.md

tomhealey-icma (approved):
I approve the change as specification is a better description than schedule but I would add that a function name ...Helper is not the best name for a function. I think we should stick to create, add, set, get, update. Also we should further the design principles with definitions for the use of type modifiers like criteria, information, terms, specification.

minesh-s-patel (approved):

PR #2123, opened 2023-04-26
Representation of extraordinary events

hugohills-regnosys:
This proposal repositions in the legaldocumentaion the description of any additional terms for extraordinary events that may affect a trade and the related contractual rights and obligation of the parties when this happens


tomhealey-icma (changes requested):
My only comment and request would be that enum ContractualDefinitionsEnum should be called ISDAContractualDefinitionsEnum

rvincentISDA:
This PR can be closed as the change will now be addressed via the following: https://github.com/finos/common-domain-model/pull/2401

PR #2046, opened 2023-03-15
Repo substitution function and qualification

tomhealey-icma:
Added repo substitution function and qualification from ICMA CDM Phase 2 for Repo and Bonds.

PR #2135, opened 2023-05-04
Substitution : new value in EventIntentEnum

PayalKhanna:
Substitution : new value in EventIntentEnum

### DIFF
--- a/rosetta-source/src/main/rosetta/event-common-enum.rosetta
+++ b/rosetta-source/src/main/rosetta/event-common-enum.rosetta
@@ -48,6 +48,7 @@ enum EventIntentEnum: <"The enumeration values to qualify the intent associated 
 	PrincipalExchange <"The intent is to pay or to receive a cash transfer, in accordance with Principal Exchange features.">
 	Reallocation <"The intent is to reallocate one or more trades as part of an allocated block trade.">
 	Repurchase <"The intent is to close a repo transaction through repurchase.">
+    Substitution <"The intent is to substitute an underlier to an existing one, which might ; this is intentionaly not considered as a sub-type of a CorporateActionAdjustement, since the business reasons that might cause the intent of either party to substitute are various, beyond the scope of CorporateActionAdjustment triggers.">
 
 enum RecordAmountTypeEnum: <"The enumeration of the account level for the billing summary.">
     AccountTotal


tomhealey-icma (approved):

rvincentISDA:
The description of the Substitution value is unclear  in the PR. What would be a typical business scenario ? An equity basket with one item being replaced ? Is this specific to a type of products ?

JBZ-Fragmos:
@rvincentISDA 

yes, as you say, 

typical scenario is for **Equity Swap on Basket** (from functional perspective, other cases exist, but for sure this is the most obvious business case) : when either party would activate their "right to substitute" e.g. say you want **stockA** to step **out**, to be replaced by **stockB** to step **in**

chrisisla (changes requested):
As per Ruddy's comment I think the comment against the new intent is incomplete. 

Substitution is not necessarily associated to corporate actions either, for example on a repo or securities lending trade you can perform a collateral substitution, so adding "Substitution" as an intent is actually a bit ambiguous. What primitiveInstruction are we expecting this new intent to be used in conjunction with?

JBZ-Fragmos:
some answers to Ruddy and Chrisisla comments above : 

- Chrisisla comment that _"Substitution is not necessarily associated to corporate actions"_ i fully agree with you ; the "substitution" is a functional action which can be made for different business reasons (corp actions, collateral substitution, basket rebalancing, others.)... 
- at this stage, what I had in mind was the kind of substitution involved in vanilla equity return swap, whether basket or single underlier... so if you want we could be more explicit and have _**"UnderlierSubstitution"**_ instead of just "Substitution" and in such case I beleive the description proposed would be correct, no more ambiguous ? 
- taking into account these comments and also the fact there is indeed a typo inside current description (actually the words _", which might ;"_ should be **removed**) I propose we have this description, that is intentionally left short with no further description of business triggers behind it : **_"The intent is to substitute one or several underlier(s) to an existing one"_** 

- and then, depending if you feel mature enough in terms of description to be associated with such type, you may want to add another one in the list of intentEnum for any other distinct business case ; as an example, you may want to add "CollateralSubstitution" for the kind you mentioned that relates to Repo or SLB trades ?

- 2nd part of Chrisisla comment about what _"primitiveInstruction"_ to use : as regards the 1st business case I want to push (that is "UnderlierSubstitution") we use **"quantityChange"** instruction to apply to a given lotIdentifier
> -  there must be multiple change[*] objects inside, in fact we need at least 2 or more, with the following features :
> - one of these change must relate to the "old" observable that the one we want to **step out** AND for this one quantity must be equal to 0
> - at least one (or several)  other change object(s) must exist with quantity<>0 AND with new observable(s), thus the one(s) we want to **step in** to replace the one that is to step out
> AND directionEnum must be "REPLACE" as the trigger for implementors functions required to build the update of the related tradeLot in regards to this quantityChange instruction

- for clarity, I'm still wondering how such quantityChange could then properly affect the referenced values inside Underlier objects in Payout legs, but this is another point, not related to "substitution" per se, that is to **clarify the functional relation** between **quantityChange** and corresponding resolved values in **priceQuantity** which reference the **tradeLot** being updated by a **quantityChange** : the generla question being how to update resolved value ? shall we keep track of related references and re-run the resolving process in regards of the same in quantityChange ? else we cannot resolve again so we should use termsChange primitive instruction combined with quantityChange in order to update in parallel both tradeLot(=quantityChange) and the product(=termsChange) ? but that looks a bit weird to me...

chrisisla:
Thanks for the explanation JB!

Adding more granularity to the name of the intent would probably be better yes, so ```UnderlierSubstitution``` would be clearer for sure. The only problem then is that we would eventually end up with several new intents, one for each different type of substitution... hmmm. So this is not ideal either... but would be a better solution.

In relation to your question about whether the substitution should perform both a ```quantityChange``` and a ```termsChange``` then I recommend having a look at the Pull Request put in by @tomhealey-icma a few weeks ago. You can find it here: #2046 

This performs both events and it is a good approach (the PR has some open questions about when to close the contract which is why it is still under review). It may help you to visualise how it could work if you look at the **Files Changed** for that PR.

Seeing as there are so many cases for substitution is there perhaps scope for a new substitution ```primitiveInstruction```?

JBZ-Fragmos:
@hugohills-regnosys 
hello, why are you closing this issue ? related topic is still open in my undertsanding, no ?

PR #2154, opened 2023-05-19
Collateral Interest Calculation Changes

PayalKhanna:
Includes redesigned types for collateral interest calculation, including a refactored floating rate type


PayalKhanna:
replaced with https://github.com/finos/common-domain-model/pull/2177

PR #2167, opened 2023-05-26
Python code gen build - changes to support integration of Python generator

hugohills-regnosys:


codefresh.yml
```diff
@@ -113,6 +113,19 @@ steps:
...
+        fail_fast: false
+        image: python:3.10-alpine
+        working_directory: ./rosetta-source/target/classes/cdm/python
+        shell: sh
+        commands:
+          - python3 -m pip install pydantic
+          - python3 -m pip install jsonpickle
+          - python3 -m pip install rosetta_runtime-1.0.0-py3-none-any.whl
+          - |-
+            python3 -m pip wheel --no-deps --only-binary :all: .
```
SimonCockx, line 128: What's the `|-` symbol for?

hugohills-regnosys: The all word needed to be escaped 
https://codefresh.io/docs/docs/pipelines/what-is-the-codefresh-yaml/#escaping-strings

SimonCockx: Got it!

PR #2175, opened 2023-06-06
icma-demo-v1

tomhealey-icma:
Version 1 of ICMA demo of repo execution and lifecycle events. Only execution, roll and early termination are  supported in this version.


tomhealey-icma:
closing due to change in file structure

PR #2181, opened 2023-06-09
Lender Availability Model Update from Chris Rayner

minesh-s-patel:


chrisisla:
@iansloyan , @tomhealey-icma , @dschwartznyc - thank you for your feedback on this topic at the CRWG yesterday. I will create a discussion where we can consider some of the other use cases that were pointed out for this type of structure in the CDM.

If nothing is forthcoming over the next couple of weeks then I think we could go with Tom's suggestion of adding "SecurityLending" or just "Security" in front of the `Availability` and `AvailabilityRecord` types, as I agree that the term "Availability" on its own is a little generic.

chrisisla:
Discussion created here: #2189 - please add some comments!

tomhealey-icma (changes requested):
As discussed in the CRWG the type name Availability is too general when it's domain and purpose are specific to securities lending pre-trade offering lists sent to buyers. I would propose, SecuritiesLendingInventoryOfferingList. If you were to separate the offeringList from the specific SL attributes then you could use OfferingList{..} and SecuritiesLendingOffering{offeringList, seller, buyer,...}

chrisisla:
Labels added - I will be updating these changes though based upon feedback and additional use cases.

This structure will also work for borrower locates, which is where a borrower asks a lender if they have a security (so the reverse of availability). I'm thinking a base class of Inventory which we can then extend with new types of SecuritiesLendingAvailability and SecuritiesLendingLocates. The reason for using Inventory as the base class is due to the other possible use case of describing collateral inventory.

chrisisla:
Closing this PR in favour of #2296 which addresses the queries from this PR and extends the use cases.

PR #2200, opened 2023-06-20
ICMA demo and examples of repo execution and lifecycle events

tomhealey-icma:
Adding a set of repo examples and repo lifecycle demo.


tomhealey-icma:
This work does not include any model changes. It's all new, self-contained demo. Maintainers please approve.

chrisisla:
As per the consensus on #2312 it looks like this demo now needs to be put into a separate repository and not included with the model itself.

Leo also asked about the copyrights in the header comments on the files. I think the Apache licence will override the copyrights so they would not be enforced (i.e. the model is open source and the licence states that anyone can use and update the content as they see fit, regardless of copyrights). I think it would be good to seek more guidance on this from FINOS (@maoo is this something you have encountered before, copyrights in files included in a FINOS open source project?) or from the LinuxFoundation (@mcleo-d)

maoo:
Hi @chrisisla !

As any other code hosted in FINOS, the CDM examples will be released under Apache 2.0 license, whose definition already mentions a copyright statement, as you can see on https://github.com/finos/software-project-blueprint/blob/main/LICENSE#L189

In this case `{name of copyright owner}` should be Regnosys and/or ICMA.

For file headers, I'd recommend using [SPDX IDs](https://spdx.dev/ids/), as they're short and not invasive; alternatively, you can opt for the [ASL file header version](https://www.apache.org/legal/src-headers.html#headers), which does not mention copyright, and therefore relies on what mentioned in the root `LICENSE` file (as discussed above).

I hope this clarifies things!

@minesh-s-patel @tomhealey-icma - could you please follow https://community.finos.org/docs/collaboration-infrastructure#finos-project-blueprint and apply the Software Project Blueprint content on that repo? It would make the code transfer process much easier.

Thanks!

tomhealey-icma:
ICMA decided to move this out of the CDM build.

PR #2201, opened 2023-06-20
FIx FraPayoutSplitter Mapper post Collateral interest calculation contribution

PayalKhanna:


rosetta-source/src/main/java/cdm/product/template/processor/FraPayoutSplitterMappingProcessor.java
```diff
@@ -107,10 +107,10 @@ private void addFloatingLegQuantityReference(Mapping mapping, InterestRatePayout
 
 	private void updateFloatingRateIndexReference(Mapping mapping, InterestRatePayoutBuilder floatingLeg) {
 		Reference.ReferenceBuilder reference = Optional.of(floatingLeg)
-				.map(InterestRatePayoutBuilder::getRateSpecification)
-				.map(RateSpecificationBuilder::getFloatingRate)
-				.map(FloatingRateBuilder::getRateOption)
-				.map(ReferenceWithMetaFloatingRateOptionBuilder::getReference)
+				.map(b -> b.getRateSpecification())
```
hugohills-regnosys, line 110: These are no-op changes, but the previous code does not compile is certain java compilers

PR #2207, opened 2023-06-22
Add TripartyAgentEnum to PartyRoleEnum

tomhealey-icma:
Adding TripartyAgent to PartyRoleEnum.

### DIFF
--- a/rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
+++ b/rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
@@ -112,6 +112,8 @@ enum PartyRoleEnum: <"The enumerated values for the party role. The enumerated v
 	TradeSource <"The organization that originally supplied the record of the trade. In the context of regulatory reporting, it is the submitter of the trade record to a regulator or TR.">
 	TradingManager <"The entity responsible for managing the assets/investments of this party.  Synonym:  Asset Manager, Investment Manager, Trading Advisory.">
 	TradingPartner <"An entity with which this party trades from time to time, ie. with which it acts as a counterparty on some transactions.   This role is used for static reference data, not individual transactions.">
+	TripartyAgent <"A third party entity that acts as an intermediary and provides automated clearing, settlement, allocation, valuation and reporting services.">
+	ThirdPartyCustodian <"A third-party providing custody, settlement, segregation and reporting services.">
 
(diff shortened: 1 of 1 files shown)


chrisisla (changes requested):
Comments please change "An third..." to "A third..." otherwise this is all good.

iansloyan:
@tomhealey-icma Good addition - surprised it was not there already

@valdensmith do you think we could or should add "Third Party" to this change for completeness? https://www.isda.org/a/978ME/Triparty-vs.-Third-Party-8.12.19.pdf

valdensmith:
This looks like an addition to the PartyRoleEnum list, where Tri Party Agent is a new one Tom has added correct? 
In that enumeration list there is already 'Custodian' which is not as specific and could mean a TriParty custodian or a third party custodian which have different roles. I am not against adding Tri Party Agent and its description, but users may get confused between the 2. Can @tomhealey-icma  confirm the use case to which the addition is required?

valdensmith:
> @tomhealey-icma Good addition - surprised it was not there already
> 
> @valdensmith do you think we could or should add "Third Party" to this change for completeness? https://www.isda.org/a/978ME/Triparty-vs.-Third-Party-8.12.19.pdf

Responded in his conversion

tomhealey-icma:
I hadn't considered the ThirdParty role and I can why it makes sense to add it as well since it acts in a different capacity. From what I understand a triparty agent should be a (I)CDS and maybe always is, but it's not clear if it must be a (I)CSD. 

So before adding the question is, should we use:
Triparty, TripartyAgent or TripartyCustodian and same for Thirdparty, ThirdpartyAgent or ThirpartyCustodian?

tomhealey-icma:
For clarification, in the repo execution workflow all parties are attached to the trade object with their respective partyrole that is needed for sftr reporting, clearing and settlement:

Here's a settlement agent example:

                              {
                                    "partyReference": {
                                          "externalReference": "549300OZ46BRLZ8Y6F65",
                                          "globalReference": "6cb3f258"
                                    },
                                    "role": "SETTLEMENT_AGENT"
                              }

tomhealey-icma:
Based on market terminology I'm proposing to add:

	TripartyAgent <"A third party entity that acts as an intermediary and provides automated clearing, settlement, allocation, valuation and reporting services.">
	ThirdPartyCustodian <"A third-party providing custody, settlement, segregation and reporting services.">

valdensmith:
All good for me now

valdensmith (approved):

chrisisla (approved):

lolabeis:
Please provide release notes in .md format as a comment in the PR before it is merged, so that notes can be added to the release tag.

You can follow these guidelines:
https://cdm.docs.rosetta-technology.io/source/contribution.html#content-of-release-notes

Issue #2205, opened 2023-06-22
Change tradeDate from date type to zonedDateTime


tomhealey-icma:
The Trade type does not include the ability to define the trade date with a time and timezone. A Product type could be created with relative dates. Example: An overnight repo can be defined in terms of a relative start date and the next business day. The actual start date will depend on the agreement between the counterparties, venue, or other third parties when the trade is executed and where it is executed.

Workflow and business event dates refer to other types of events. Workflows are actions that take place throughout the execution and lifecycle process but remain separate from the trade date which is agreed to by the counterparties. The same is true with a business event. An execution event may take place at one date and time while the trade date agreed to by the counterparties may be different.

The proposed solution is to change the Trade->tradeDate to zonedDateTime with reference to a business center or add a separate Trade->tradeDateTime with type zonedDateTime and deprecate tradeDate in the future.

chrisisla:
I agree that a trade time and timezone would be beneficial on the trade, but I think this should be separate to the trade date. I have not heard of counterparties agreeing a trade time, but this is not to say this is always the case. My suggestion would thus be to not change the usage or structure of tradeDate but add new items for the trade time/timezone as required.

lolabeis:
I agree with the comments above re separating the trade date and time (and having that second one optional only).

But I would also point-out that the direction of travel for the CDM generally was to _do away_ with trade date as an attribute of the trade itself. The reason is that if trade date is about:
1) recording when the trade event took place, it belongs to the event date on the execution event
2) a date that matters for the contract mechanics (e.g. calculating accruals or settlement), it belongs to the product's effective date

Therefore, further entrenching trade date at trade level by adding time / location would work against that objective.

If it's about needing time or location for case 2), could you include some term sheet language that explicitly references those?

tomhealey-icma:
I have separated the tradeDate and tradeTime. Please see, [PR2210](https://github.com/finos/common-domain-model/pull/2210). 

I'll give you a couple scenarios why this is necessary:

One scenario is where a repo security (security + term) is defined as a spot-next has a start-date equal to the trade-date plus 2 days and a end date equal to the next business day after the spot date. These dates are not known until the trade is agreed to, so the effective date is relative to the trade date and the end date is relative the start-date. You could assume the trade date and event date are the same but this is not always the case either, which leads to second scenario.

[...]

PR #2210, opened 2023-06-22
Add tradeDateTime to Trade type and Execution Instruction

tomhealey-icma:
Reference issue,[2205](https://github.com/finos/common-domain-model/issues/2205)

### DIFF
--- a/documentation/source/event-model.rst
+++ b/documentation/source/event-model.rst
@@ -67,6 +67,8 @@ The ``Trade`` data type defines the outcome of a financial transaction between p
    tradeIdentifier TradeIdentifier (1..*)
    tradeDate date (1..1)
      [metadata id]
+   tradeTime TimeZone (0..1)
+     [metadata id]	 
    tradableProduct TradableProduct (1..1)
    party Party (0..*)
    partyRole PartyRole (0..*)
@@ -356,6 +358,8 @@ Therefore, the execution function does not take any before state as input and al
    partyRoles PartyRole (0..*)
    executionDetails ExecutionDetails (1..1)
    tradeDate date (1..1)
+       [metadata id]
+   tradeTime TimeZone (0..1)
        [metadata id]
    tradeIdentifier TradeIdentifier (1..*)
    collateral Collateral (0..1)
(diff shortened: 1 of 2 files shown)


documentation/source/event-model.rst
```diff
@@ -63,20 +63,22 @@ The ``Trade`` data type defines the outcome of a financial transaction between p
...
+	tradeIdentifier TradeIdentifier (1..*)
+	tradeDate date (1..1)
+        [metadata id]
+	tradeDateTime zonedDateTime (0..1)
+        [metadata id]
+	tradableProduct TradableProduct (1..1)
+	party Party (0..*)
+	partyRole PartyRole (0..*)
+	executionDetails ExecutionDetails (0..1)
+	contractDetails ContractDetails (0..1)
```
chrisisla, line 76: Is contracttDetails being added in this PR as well?

rosetta-source/src/main/rosetta/event-common-type.rosetta
```diff
@@ -103,6 +103,8 @@ type ExecutionInstruction: <"Specifies instructions for execution of a transacti
 	executionDetails ExecutionDetails (1..1) <"Specifies the type and venue of execution, e.g. via voice, or electronically.">
 	tradeDate date (1..1) <"Denotes the trade/execution date.">
 	    [metadata id]
+	tradeDateTime zonedDateTime (0..1) <"Denotes the trade execution date and time with timezone.">
```
chrisisla, line 106: As per my comment on the associated Issue I think we should have separate trade date and trade time/timezone items.

chrisisla (changes requested):

tomhealey-icma:
I agree with this approach. We'll keep tradeDate as is and add tradeTime TimeZone (0..1) which adds the other two attributes time and timezone that make up zonedDateTime.

documentation/source/event-model.rst
tomhealey-icma: No that is probably just the result of indenting.

lolabeis:
Please provide release notes in .md format as a comment in the PR before it is merged, so that notes can be added to the release tag.

You can follow these guidelines:
https://cdm.docs.rosetta-technology.io/source/contribution.html#content-of-release-notes

PR #2211, opened 2023-06-23
Ian Sloyan Adding party specific eligibility criteria

regnosys-prod-user:


rosetta-source/src/main/rosetta/product-collateral-type.rosetta
```diff
@@ -111,6 +111,7 @@ type CollateralCriteriaBase: <"Represents a set of criteria used to specify and
 type EligibleCollateralCriteria extends CollateralCriteriaBase: <"Represents a set of criteria used to specify eligible collateral.">
 
 	treatment CollateralTreatment (1..1) <"Identifies the treatment of specified collateral, e.g., haircuts,holding limits or exclusions.">
+    partyCriteriaAppliesTo Party (1..*) <"Identifies the parties to which the specified eligible collateral and treatment applies to. e.g. Party A or Party B (Pledgor or Receiver)">
```
chrisisla, line 114: Is this referring to a Party by name or by Role? From your description it looks like by role in which case would this be type partyRole instead? Happy adding the party at this level but I don't think it should be mandatory i.e. cardinality 0..* rather than 1..*, as there may be no need to identify a party.

valdensmith:
Hi @iansloyan @chrisisla. I think partyRole directs to Party and does the same thing so not too sure?
On the cardinality topic if its left mandatory there would be requirement for the user to specify that the eligible collateral they are specifying is eligible/applicable to both party 1 and 2 or for either independently. Or if it was optional the user would need to specify if it applies to one or the other, if not used there should be a caveat that it defaults to both parties ? I am in two minds on this but an swaying toward it being mandatory to ensure it is clear which parties are included 1 or both. Happy to discuss further

rosetta-source/src/main/rosetta/product-collateral-type.rosetta
valdensmith: replied as a comment

tomhealey-icma (changes requested):
My comments on this are:
1. The party itself doesn't seem to fit within criteria. Instead it may better fit within collateralProvisions. 
2. Collateral agreements applies to both parties. I think what this is suggesting is the party responsible for delivery. 
3. On product term sheets the party responsible for collateral delivery is not know only their role as buyer or seller. So I think it would be better to use a party reference instead of the actual party. I think it's important to always separate a product from a trade and products should stand on their own.

lolabeis:
I agree that the positioning of `Party` here is odd.

In the CSA model, the `Party` is positioned higher-up in `PostingObligationsElection`. Having an alternative position would make the model confusing. Could we add this attribute to the `CollateralProvisions` object, as per Tom's suggestion?

Also like the CSA model, it should use `CounterpartyRoleEnum` to refer to one of the parties rather than re-specifying it.

image
image

lolabeis:
Supersded by #2436

Issue #2213, opened 2023-06-23
Deprecate SettlementCenterEnum


tomhealey-icma:
The SettlementBase attribute SettlementCenter is a enum of two options, EuroclearBank or ClearstreamBankingLuxembourg. The use of enums for this purpose is not necessary and actually more complicated to match settlement parties to a specific enum. Instead there should be settlement parties with various roles needed to support the settlement process. The proposal is to deprecate SettlementCenter and add settlementParty Party(0..*) and settlementRole(0..*) with the option to reference the parties on the trade.

lolabeis:
Reopening as #2214 has been reverted by #2326

lolabeis:
> We could use AncillaryParty that includes AncillaryRoleEnum and partyReference Party (1..*) <"Specifies the party, or parties, associated to the ancillary role.">
[metadata reference]
We would mark settlementCentre SettlementCentreEnum (0..1) for deprecation.

Thanks @tomhealey-icma. Some implementation considerations:
- `SettlementBase` should only include the `AncillaryRoleEnum` attribute, no `Party` (or reference)
- The link between `Party` and `AncillaryRoleEnum` should be done at the `TradableProduct` level using the existing `AncillaryParty` attribute
- `AncillaryRoleEnum` needs a new enum value for settlement centre
- A complete contribution for v7 should include a remapping of the old to the new attributes for existing trade samples that enables regression tests to pass, rather than just marking the old attribute as "deprecated"

tomhealey-icma:
If we proceed in adding AncillaryParty to AssetBase then we can just add settlementPartyRole AncillaryPartyRole (0..*) since parties linked to the product and not the trade will be available in assetBase. We can then deprecate settlementCenterEnum.

tomhealey-icma:
When splitting parties into counterparties and ancillary parties the question is, if a party is acting as an ancillary party can it also be a counterparty or would it's role change from being a ancillary party to a counterparty even though it's the same party. Examples, executing broker acting as principal, clearing firm acting as counterparty, issuer acting as buyer or seller. My initial list for PartyRoleEnum is:

BeneficialOwner
Beneficiary
Borrower
Broker
Buyer
Client
ContractualParty
CounterPartyAffiliate
CounterPartyUltimateParent
Counterparty
DisputingParty
ExecutingBroker
HedgingParty
Lender

CDM-ReleaseManagement-OT:
[...]

PR #2214, opened 2023-06-23
Add settlement party and role to settlement base

tomhealey-icma:
This model change will deprecate settlementcenterenum and add settlement party and party role. Reference, [2213](https://github.com/finos/common-domain-model/issues/2213).


chrisisla (changes requested):
I think this change makes a lot of sense. 
Looking at the rosetta file there is a comment just below this change of "//ICMA-P2", can this be removed as well please?
There is also a very large [docReference] next to the "settlementCurrency" just above this change. I know we decided to use [docReference] to _reference_ external text and not include big swathes of text in the model. While we're in this file could we reduce the amount of text in this [docReference]?

tomhealey-icma:
I can remove the //ICMA-P2 comment and change the doc reference to:
[docReference ICMA GMRA namingConvention "Contractual Currency" provision "As defined in GMRA paragraph 2(k)/ paragraph 7(a)"]

chrisisla (approved):

lolabeis:
I do agree with the driver and purpose of the change, however it should be implemented differently.

The Party should not feature "in hard" in Settlement Terms, which is a low-level object inside Trade. Instead, the SettlementCenter should be added as an enum value in AncillaryRoleEnum. Then the actual Party entity should be specified in the Trade's list of Parties, and then tied to that enum value in the ancillaryParty attribute of TradableProduct.

Please see `ExerciseNotice` type for a template implementation.

Also could you articulate your plan for actual deletion of the existing attribute (whih you're simply marking as `deprecated`)? Existing ingested samples where this attribute is populated will need to be remapped to the new structure.

PR #2218, opened 2023-06-27
Added Python unit tests to confirm that generated code performs as expected 

hugohills-regnosys:


hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2219

PR #2293, opened 2023-07-31
Ruddy Vincent Repositioning of Extraordinary Events

regnosys-prod-user:


tomhealey-icma (changes requested):
As we discussed the changed files should be cleaned up and we need to address adding contract information to the product definition.

rvincentISDA:
This PR will be closed as the change will now be processed via: https://github.com/finos/common-domain-model/pull/2401

PR #2296, opened 2023-08-03
ISLA: Modelling: Lender Availability and Borrower Locates

regnosys-prod-user:
**Modelling Lender Availability and Borrower Locates**

_Background_
In securities lending a lender will broadcast their availability to lend specific securities to potential borrowers. This process allows borrowers to easily identify and select lenders who have the desired securities available for lending and can help facilitate trades more quickly and efficiently. 

Borrowers will also send out requests to lenders looking to locate specific securities. This allows borrowers to confirm that they can acquire the security before proceeding with a short sale of that security.

As part of the ongoing work into standardising pre-trade processes by the ISLA Trading Working Group, enhancements are being introduced to the model to support the representation of both lender availability and borrower locates.

_Model Changes_
The structures of lender availability and borrower locates are extremely similar. It has also been suggested that the structure of collateral inventory could be represented the same way. To this end new base types have been added that can be extended to support multiple use cases.

The `InventoryBase` type can be used to hold a list of securities. This base type has been extended into the `Inventory` type. The term “Inventory” in the financial markets refers to the securities that a party has on hand and thus works well as a base class name.  

The details of the securities held in the inventory are described in the `InventoryRecordBase` type. This base type has been extended into the `InventoryRecord` type. 

The similarity between the lender availability and borrower locate use cases led onto the creation of two new types, `AvailableInventory` and `AvailableInventoryRecord`.

The `AvailableInventoryRecord` type extends `InventoryRecord` which allows it to inherit the security details. Additional criteria that are applicable to both availability and locate processing are held here too, including `collateralProvisions`, `expirationDateTime`, `partyRole` (to allow cross referencing back to parties in the main `AvailableInventory` type described below) and `priceQuantity` which is where the quantity of securities associated to this available inventory record can be specified.

The `AvailableInventory` type is where the `AvailableInventoryRecord` is used, and describes a set of inventory that is held by a party. The `AvailableInventoryRecord` array is where the individual rows of available inventory are held. In addition to the available securities this type also contains the `party` and `partyRole` types. The `party` here should define all the parties that are held across all the elements in the `AvailableInventoryRecord` array. The `partyRole` can also be used at this level to define the role each party has in relation to the inventory. 

For the lender availability use case, the `SecurityAvailability` type is a new type that will hold a list of all the securities that a lender has available. Each available security is then held in an element in the `availableInventoryRecord` array, which can hold details of the security, the quantity of shares available and some optional additional criteria that is often used by the borrower to determine if they want to pursue a new trade with the lender for a specific security.

For the borrower locates use case, the `SecurityLocate` type is a new type that will hold a list of all the securities that a borrower is looking to locate from a lender. Each requested security is then held in an element in the `availableInventoryRecord` array, which can hold details of the security, the quantity of shares required and some optional additional criteria that is often used by the lender to determine if they want to lend the shares to the borrower.

The `SecurityLocate` type also contains a condition to ensure that the `availableInventoryRecord` item exists, as a locate must always request availability for at least one security.

_Data Types_
- Added `InventoryBase` type:
  - Includes new `inventoryRecord` type
- Added `Inventory` type:
  - Extends `InventoryBase`
  - There is nothing else in this type at the moment, it has been created for consistency and further extension in the future
- Added `InventoryRecordBase` type:
  - Includes `identifier` and `security`
- Added `InventoryRecord` type:
  - Extends `InventoryRecordBase`
  - There is nothing else in this type at the moment, it has been created for consistency and further extension
- Added `AvailableInventory` type:
  - Includes `party`, `partyRole` and new type `availableInventoryRecord`
- Added `AvailableInventoryRecord` type:
  - Extends `InventoryRecord` and thus inherits `identifier` and `security` from it
  - Additionally includes `expirationDateTime`, `collateral`, `partyRole` and `priceQuantity`
- Added `SecurityAvailability` type:
  - Extends `AvailableInventory` and thus inherits `party`, `partyRole` and `availableInventoryRecord` from it
  - Additionally includes `messageInformation`
- Added `SecurityLocate` type:
  - Extends `AvailableInventory` and thus inherits `party`, `partyRole` and `availableInventoryRecord` from it
  - Additionally includes `messageInformation`
  - There is a condition `RequestOneSecurityMinimum` which ensures `availableInventoryRecord` is populated


chrisisla:
This continues the work originally contributed in PR #2181 - that pull request has now been closed as this one supersedes it.

The original modelling has been extended to support two use cases (and with a third possible one in mind too).

chrisisla (dismissed):

tomhealey-icma (changes requested):
The terms used here, InventoryBase and Inventory are very generic terms used to describe a very specific workflow element. It seems that InventoryBase already includes attributes related to the securities lending process. Inventory base should be generic like a list of securities, Inventory extends InventoryBase should add more specific details like portfolio, fund, or type, and then add a new type, AvailableLendingInventory extends Inventory could include the parties and other attributes.

chrisisla:
Hi @tomhealey-icma 
So the suggestion is essentially to move the party details out of InventoryBase and put them into the types associated to the use cases we intend to support, so Availability and BorrowerLocate instead. That makes sense yes, I can make that change.
Thanks!

chrisisla:
Hi @tomhealey-icma,
I've updated the structure as we discussed the other day, so we now have AvailableInventory and AvailableInventoryRecord, where AvailabilityInventoryRecord is used by both the SecurityAvailability and SecurityLocate types:

SecurityAvailability:
image

SecurityLocate:
image

The SecurityLocate type has an additional condition in it ensuring that availableInventoryRecord exists, as you must have at least one security being requested.

Let me know if you have any further questions, this change should resolve all the outstanding issues we discussed.

Thanks!

rosetta-source/src/main/rosetta/event-position-type.rosetta
```diff
@@ -42,3 +48,35 @@ type Portfolio: <" A Portfolio represents an aggregation of multiple Positions,
...
+    identifer AssignedIdentifier (1..1) <"Unique identifier for this record. This can be used to uniquely identify a specific piece of inventory.">
+    security Security (1..1) <"The security details.">
+
+type Inventory extends InventoryBase: <"Describes an inventory of securities.">
+
+type InventoryRecord extends InventoryRecordBase: <"Describes an individual piece of inventory.">
+
+type AvailableInventory: <"A data type that can be used to describe the inventory of securities that a party holds. ">
+    party Party (0..*) <"Defines all parties involved for the list of inventory records in this set of inventory. For example, when used to describe securities lending availability, this could hold the sender of the availability, the intended recipient, the beneficial owner(s), the lender (which may differ from the sender as the lender may have the same piece of availability going through multiple agents), an agent or a venue.">
+    partyRole PartyRole (0..*) <"Defines the role(s) that party(ies) may have in relation to the inventory.">
```
lolabeis, line 65: I understand the description of `PartyRole` in `AvailableInventoryRecord`, that it needs to be defined at the security level. However in this case we shouldn't duplicate `PartyRole` in `AvailableInventory` as well. Or are we saying that there are certain party roles that apply at the `AvailableInventory` level, and some at the `AvailableInventoryRecord` level? In which case you may want to restrict the set of enum values that would be allowed at each level, or at least make sure that they are mutually exclusive.

rosetta-source/src/main/rosetta/event-position-type.rosetta
```diff
@@ -42,3 +48,35 @@ type Portfolio: <" A Portfolio represents an aggregation of multiple Positions,
...
+    collateral CollateralProvisions (0..*) <"The type of collateral can often be required when determining if the piece of availability being described is suitable for a party.">
+    partyRole PartyRole (0..*) <"An individual security may be held by several agents. Including the party role at this level allows us to reference the party holding this specific item.">
+    priceQuantity PriceQuantity (0..*) <"Primarily used to hold the quantity of the security. An additional element can be used to hold an optional rate associated to this piece of availability.">
+
+type SecurityAvailability extends AvailableInventory: <"SecurityAvailability describes a list of availability records, with each record being an individual security and its associated criteria. Criteria can include the quantity available, the rate at which the security is available to borrow at, as well as other details that can affect the decision as to whether a party wants to borrow the securities being offered.">
+    messageInformation MessageInformation (0..1) <"Allows details related to the availability messaging to be defined">
```
lolabeis, line 75: If the `messageInformation` attribute is optional in both cases, I would put it down into `AvailableInventory` directly. This means you could model a security availability as an `AvailableInventory` type directly. Less types in the model = good.

rosetta-source/src/main/rosetta/event-position-type.rosetta
```diff
@@ -42,3 +48,35 @@ type Portfolio: <" A Portfolio represents an aggregation of multiple Positions,
 
 	aggregationParameters AggregationParameters (1..1) <"Describes the portfolio by describing how to aggregate all its relevant Events.">
 	portfolioState PortfolioState (1..1) <"Describes the state of the Portfolio as a list of Positions resulting from the aggregation.">
+
+type InventoryBase: <"A data type that can be used to describe an inventory of securities.">
+    inventoryRecord InventoryRecord (0..*) <"An array holding the list of inventory being described. Each element in the inventoryRecord array represents an individual piece of inventory i.e. a security.">
+
+type InventoryRecordBase: <"An individual piece of inventory. This represents a single security.">
+    identifer AssignedIdentifier (1..1) <"Unique identifier for this record. This can be used to uniquely identify a specific piece of inventory.">
+    security Security (1..1) <"The security details.">
```
lolabeis, line 57: Do you want to make inventory exclusive to `Security` only? Inventory could apply to assets you hold more broadly, so may be more aligned to `Product`. For instance this could apply to `Commodity` as well?

Note: this is related to the broader discussion we're going to have regarding the different layers of "Products".

rosetta-source/src/main/rosetta/event-position-type.rosetta
```diff
@@ -42,3 +48,35 @@ type Portfolio: <" A Portfolio represents an aggregation of multiple Positions,
...
+    availableInventoryRecord AvailableInventoryRecord (0..*) <"An array holding the list of inventory being described. Each element in the inventoryRecord array represents an individual piece of inventory i.e. a security.">
+	
+type AvailableInventoryRecord extends InventoryRecord: <"An individual piece of available inventory. This represents a single security and its associated criteria. The criteria are used to describe any restrictions on the securities.">
+    expirationDateTime zonedDateTime (0..1) <"There may be a set period/time restriction associated to the security.">
+    collateral CollateralProvisions (0..*) <"The type of collateral can often be required when determining if the piece of availability being described is suitable for a party.">
+    partyRole PartyRole (0..*) <"An individual security may be held by several agents. Including the party role at this level allows us to reference the party holding this specific item.">
+    priceQuantity PriceQuantity (0..*) <"Primarily used to hold the quantity of the security. An additional element can be used to hold an optional rate associated to this piece of availability.">
```
lolabeis, line 72: Why `PriceQuantity` and not just `Quantity`, if it's only to hold the quantity?

And why `(0..*)`? If only 1 security, max cardinality should be 1. And should attribute really be optional?

Also I would have expected this quantity attribute to be in `InventoryRecord` rather than in the extended type. Does it make sense to talk about inventory without talking about a quantity?

rosetta-source/src/main/rosetta/event-position-type.rosetta
```diff
@@ -42,3 +48,35 @@ type Portfolio: <" A Portfolio represents an aggregation of multiple Positions,
 
 	aggregationParameters AggregationParameters (1..1) <"Describes the portfolio by describing how to aggregate all its relevant Events.">
 	portfolioState PortfolioState (1..1) <"Describes the state of the Portfolio as a list of Positions resulting from the aggregation.">
+
+type InventoryBase: <"A data type that can be used to describe an inventory of securities.">
```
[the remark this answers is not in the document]
chrisisla: Agreed. I have renamed ```InventoryBase``` to ```Inventory```, and ```InventoryRecordBase``` to ```InventoryRecord```. The original types have been removed

rosetta-source/src/main/rosetta/event-position-type.rosetta
chrisisla: I’ve moved ```messageInformation``` into ```AvailableInventory```, and as you say, this now means that the ```SecurityAvailability``` type is no longer required. I’ve updated the comments in ```AvailableInventory``` accordingly.

rosetta-source/src/main/rosetta/event-position-type.rosetta
chrisisla: The ```PriceQuantity``` was used as we need to support the entry of a quantity and potentially a rate associated to the holding too. We also need to support the use case where a lender does not want to include either of these details, only the security code itself. This is why the cardinality was set to (0..*).
I’ve changed this to now use a ```Quantity``` type with cardinality (0..1) for the available quantity and a ```Price``` type with cardinality (0..1) for an optional rate against that quantity. I have also added a condition to confirm that if the price (rate) is present then it is set as an Interest rate in the priceType.

rosetta-source/src/main/rosetta/event-position-type.rosetta
```diff
@@ -38,6 +44,49 @@ type AggregationParameters: <" Parameters to be used to filter events that are r
...
+    messageInformation MessageInformation (0..1) <"Allows details related to the availability messaging use case to be defined"> 
+    party Party (0..*) <"Defines all parties involved for the list of inventory records in this set of inventory. For example, when used to describe securities lending availability, this could hold the sender of the availability, the intended recipient, the beneficial owner(s), the lender (which may differ from the sender as the lender may have the same piece of availability going through multiple agents), an agent or a venue.">
+    partyRole PartyRole (0..*) <"Defines the role(s) that party(ies) may have in relation to the inventory.">
+    availableInventoryRecord AvailableInventoryRecord (0..*) <"An array holding the list of inventory being described. Each element in the inventoryRecord array represents an individual piece of inventory i.e. a security.">
+	
+    condition ValidPartyRole: <"Restrict roles to be only those that would be associated to the overall list of availability">
```
lolabeis, line 63: Don't you also want to test that if, say, `Lender` is specified at the `AvailableInventory` level, then `Lender` is absent from `partyRole` at the `AvailableInventoryRecord`? (to avoid the risk of inconsistency)

rosetta-source/src/main/rosetta/event-position-type.rosetta
```diff
@@ -38,6 +44,49 @@ type AggregationParameters: <" Parameters to be used to filter events that are r
...
+    condition InterestRate: <"Where an Interest Rate has been specified then the Price Type must be set to interest rate">
+        if interestRate exists then
+            interestRate -> priceType = PriceTypeEnum -> InterestRate
+
+    condition ValidPartyRole: <"Restrict roles to be only those that would be associated to an individual piece of availability">
+        if partyRole exists then
+            partyRole -> role any = PartyRoleEnum -> AgentLender 
+            or partyRole -> role any = PartyRoleEnum -> BeneficialOwner            
+            or partyRole -> role any = PartyRoleEnum -> Custodian
+            or partyRole -> role any = PartyRoleEnum -> Lender
```
lolabeis, line 87: The condition as written doesn't achieve the objective, because you're only testing that `any` is in the allowed list. So if 1 is in and 1 is out, the condition still passes.

You actually need to write (with only 2 enum values for simplicity):

```
condition ValidPartyRole:
        partyRole
            extract ( role = PartyRoleEnum -> Lender or role = PartyRoleEnum -> Custodian )
            then all = True
```

Which, if you have more allowable values, could be written more compactly as:

```
condition ValidPartyRole:
        partyRole
            extract ([ PartyRoleEnum -> Lender, PartyRoleEnum -> Custodian ] contains role)
            then all = True
```

rosetta-source/src/main/rosetta/event-position-type.rosetta
lolabeis: See also other comment below re the semantics of the condition

rosetta-source/src/main/rosetta/event-position-type.rosetta
chrisisla: Thanks for the tip Leo! I have updated the conditions as suggested.

rosetta-source/src/main/rosetta/event-position-type.rosetta
chrisisla: I did raise this at the ISLA CDM WG last Thursday but the consensus was that there could be occasions where the roles were not mutually exclusive. It is also true to say that a party may fulfil multiple roles (thus ```partyRole``` at the record level is 0 to many) which again could lead to the same role being required at the top level and the individual record level.

lolabeis (approved):

lolabeis:
Superseded by #2469

PR #2319, opened 2023-08-21
Fix contents of release note docs page

hugohills-regnosys:


iansloyan (approved):
I assume this adds a dynamic link to within the docs page rather than weblink - approving

PR #2325, opened 2023-08-22
Fix doc snippets for SettlementBase

hugohills-regnosys:
Type was changed in PR https://github.com/finos/common-domain-model/pull/2214 but docs were not


hugohills-regnosys:
Closing PR as https://github.com/finos/common-domain-model/pull/2214 is reverted

PR #2335, opened 2023-08-24
Vineet Kumar Task6023_IndexNamesChange

regnosys-prod-user:

PR #2336, opened 2023-08-24
index names place holder

PayalKhanna:
original contribution #2335


winitall01 (approved):
changes inline with expectations

lolabeis:
> changes inline with expectations

What is the driver for the validation failure increase?

winitall01:
> > changes inline with expectations
> 
> What is the driver for the validation failure increase?

Hello @lolabeis - the expectation errors occured due to a require choice clause between index name and id. we've modified it now.

lolabeis (commented):
The release note needs to be changed to reflect what has been implemented. `IndexReferenceInformation` is not, in fact, a new type. It already existed, but the problem was that it contained all the information related to a credit index. So your work was to split it into two and create a more basic `IndexReferenceInformation` type that could be encapsulated within a rate option. The release notes also need to mention `CreditIndexReferenceInformation` and how it's been implemented as an extension.

lolabeis (commented):
@winitall01 Please proceed with this PR following today's conversation between CDM maintainers. The release note still needs to be adjusted as per my earlier comment.

iansloyan (approved):
Approved based on other issues being addressed in future

chrisisla (dismissed):

hugohills-regnosys (approved):

PR #2348, opened 2023-08-30
Open source reporting

SimonCockx:


examples/pom.xml
```diff
@@ -78,7 +78,7 @@
             <plugin>
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <configuration>
-                    <additionalJOption>--no-module-directories</additionalJOption>
+                    <additionalJOption>${maven-javadoc-plugin.option}</additionalJOption>
```
SimonCockx, line 81: I've refactored this a bit to make it possible to build the CDM with Java 17 (as well as Java 11). The `--no-module-directories` option was removed in Java 13 (it is assumed implicitly from then onwards), so I've factored this option out in a Maven property that is only set when Java is 12 or below (using a profile).

rosetta-source/src/main/resources/cdm-sample-files/functions/fpml-5-10/processes/msg-ex52-execution-advice-trade-partial-novation-C02-00-func-output.json
```diff
@@ -706,26 +706,6 @@
...
-              "meta" : {
-                "scheme" : "http://www.fpml.org/coding-scheme/external/iso17442"
-              },
-              "value" : "549300L7VWW3G1702646"
-            },
-            "identifierType" : "LEI",
-            "meta" : {
-              "globalKey" : "1454247a"
-            }
-          } ]
```
SimonCockx, line 728: This party was present twice in the function output due to a bug having to do with the builder pattern of the generated Java code. This bug has been resolved, so this duplicate was rightfully removed from the expectation.

The bug is actually a symptom of a design issue in the generated build pattern. For details on the root issue, see https://github.com/REGnosys/rosetta-dsl/issues/631.

hugohills-regnosys (dismissed):

hugohills-regnosys (dismissed):

hugohills-regnosys (approved):

PR #2353, opened 2023-08-31
Manuel Carrera Commodity spread price type prod

regnosys-prod-user:


PayalKhanna:
replaced with https://github.com/finos/common-domain-model/pull/2354

PR #2403, opened 2023-09-26
Story-2691: add instruction on how to use the download

PayalKhanna:
projection of the CDM


chrisisla:
Hi @PayalKhanna,
Not convinced this is the right place to put this comment but I've tried to use the link and I can download the zip file okay (it looks like it is from a unix gzip file as windows extraction did not work but 7zip did) but the resultant file does not open with Excel. The unzipped file does not have an extension on it but when I add ".xlsx" and open it using Excel it complains that the format is not correct. Could be that 7zip is not extracting it properly - have you extracted and viewed successfully on windows? If so what applications did you use?
Thanks!

PayalKhanna:
Hi @chrisisla 
i just got this tested on windows 7zip worked just fine. please see this
[screen-recorder-tue-sep-26-2023-15-39-22.webm](https://github.com/finos/common-domain-model/assets/107400406/d202b8a1-dbf6-481f-a0cd-38b78d3529be)

chrisisla:
> Hi @chrisisla i just got this tested on windows 7zip worked just fine. please see this [screen-recorder-tue-sep-26-2023-15-39-22.webm](https://github.com/finos/common-domain-model/assets/107400406/d202b8a1-dbf6-481f-a0cd-38b78d3529be)

Ah okay got it. So this is a zipped directory in a zip file i.e. if I am using 7zip command line I need to unzip twice to get to the excel folder where the "model.xlsx" file is held. 

Is this going to be documented anywhere? It might be worth it as not everyone will be using a zip manager to open the zip file.

minesh-s-patel (approved):

hugohills-regnosys (approved):

PR #2409, opened 2023-09-29
Remove legacy Rosetta Project module

Chuks-Ajeh:
This is to remove the Rosetta-project project within the CDM repo that is no longer in use or being maintained. It is part of a larger effort where the project is also being removed from other repos.


tomhealey-icma:
Please add labels. If you cannot add them to the PR, put them in a comment and one of the maintainers will add them.

Issue #2216, opened 2023-06-26
Add contractDetails to Product


tomhealey-icma:
The current model can only support contract details at the Trade level after execution and during the contract formation phase. For most financial products a legal agreement and governing law are set on the product during the creation of the product before execution. Consider bond issuance, structured products with term sheets, swap agreements based on ISDA agreements, repo transaction executed with a GMRA vs a private agreement. I propose adding a new attribute to ProductBase -> productContract ContractDetails (0..*). I'm open to this being a required field as well.

JBZ-Fragmos:
- if to **add** it as an **optional info** to Product, I would **agree**
- if to **replace** the current place of this object (that is an attribute of Trade) or to make it additional but **required** at Product level, then I would **disagree**

hereunder reasons to document my position, hope will make sense for you and clarifies :

eventually the Product object is mostly used from business point of view as the Observable &/or the Underlier of Payout legs involved in a given Trade, so attaching ContractDetails to Product actually means you would attach Legal Terms to an Observable/Underlier which honestly looks a bit unusual 

(also would lead to obvious corner cases e.g. imagine a Trade on a Basket = multiple Products you would have multiple ContractDetails=LegalAgreements... whereas obviously you want only 1 set of legal terms for the whole trade... that is why you want it at Trade level... not Product level...)

accordingly, i beleive the ContractDetails are attached in the correct place as of today that is at Trade level which present the "agreement" whether on Economics (roughly speaking "EconomicTerms") or on Legal terms as well ("ContractDetails=LegalAgreement")

besides, the reasons you provide e.g. _"structured products with term sheets, swap agreements based on ISDA agreement"_ for me are look like the reasons why proper place for ContractDetails is precisely at Trade level, and not at Product/Underlier level :

- in practice, traders/middleOfficers will first reconcile and get agree on a termsheet during execution phase, usually they do not check legal terms (since termsheets mostly focus on core economics required for front/risk booking) at this stage so only EconomicTerms in Trade need to be populated via Execution Instruction that will path trough EconomicTerms, at **Trade level**
- and then as 2nd step one of the party will draft the Legal Form to be sent for further agreement, that is when ContractFormation Instruction will be populated to path through related additional terms to ContractDetails, **again at Trade level**

[...]

PR #2419, opened 2023-10-02
Add legalAgreement and governingLaw to ProductBase

tomhealey-icma:
This PR addresses a gap in the product model that did not include a reference to the legal agreements that define the product. The CDM only included contract details at the trade level which can only be added after execution. Product offerings, especially primary issuance, exchange traded products and swap execution facilities offer products to be traded with pre-defined legal agreements.


lolabeis:
See comment to issue #2216. The proposed change would contravene the product concept and break the CDM's namespace hierarchy as a result.

tomhealey-icma:
We should then move the contract details to a separate namespace. Which do you recommend?

PR #2436, opened 2023-10-09
Vineet Kumar Task6239_EligibleCollateralAddition

regnosys-prod-user:


valdensmith:
Adding CounterpartyRoleEnum does not give you the ability to specify the details of the counterparty it only says if it is party 1 or 2 which is meaningless if not specified along side this. We need the ability to specify party details and (party1/2) in eligible collateral specification as its a root type, it will be used independently and not limited to attaching to a legal agreement or trade where party details can be pulled through. This was the whole objective

Oblongs:
Superceded by PR #2453

lolabeis:
Superceded by PR #2453

PR #2439, opened 2023-10-10
Minesh Patel Added CheckEligibilityByDetails function

regnosys-prod-user:


lolabeis:
My over-arching comment is that I'm not convinced about the query object having optional attributes. I would expect the query to represent a *single*, well identified asset for which all the attributes can be populated.

Instead, having optional attributes suggests that a query can represent a *range* of assets (e.g. all the government bonds of a given country).  But then when we are running that query against the eligibility criteria, we are comparing a set (the query) against another set (the criteria). The problem is that in this context, "comparison" is ambiguous: does it mean "inclusion" or "overlap"?

Take the "all the government bonds of a given country" query example, and say that the criteria included government bonds from that country but also had a maturity restriction of 1 to 5 years, such that not all assets in the set may be eligible. In this scenario, if the comparison means "included" it's false, but if it means "overlap" it's true. In fact, it's worse than that: because the full set of maturities in the queried set is not accessible from the query, "included" may be true (if all those bonds had a maturity between 1 and 5 years), or "overlap" may be false (if all those bonds had a maturity greater than 5 years). In other words: this query has no answer.

This example illustrates why, if we're after a single yes/no answer to whether a query is eligible, that query can only be of a single asset for which all attributes are known. To run with multiple assets, the query should be run multiple times with a single asset each (and some assets may be included, others excluded).

minesh-s-patel:
> My over-arching comment is that I'm not convinced about the query object having optional attributes. I would expect the query to represent a _single_, well identified asset for which all the attributes can be populated.
> 
> Instead, having optional attributes suggests that a query can represent a _range_ of assets (e.g. all the government bonds of a given country). But then when we are running that query against the eligibility criteria, we are comparing a set (the query) against another set (the criteria). The problem is that in this context, "comparison" is ambiguous: does it mean "inclusion" or "overlap"?
> 
> Take the "all the government bonds of a given country" query example, and say that the criteria included government bonds from that country but also had a maturity restriction of 1 to 5 years, such that not all assets in the set may be eligible. In this scenario, if the comparison means "included" it's false, but if it means "overlap" it's true. In fact, it's worse than that: because the full set of maturities in the queried set is not accessible from the query, "included" may be true (if all those bonds had a maturity between 1 and 5 years), or "overlap" may be false (if all those bonds had a maturity greater than 5 years). In other words: this query has no answer.
> 
> This example illustrates why, if we're after a single yes/no answer to whether a query is eligible, that query can only be of a single asset for which all attributes are known. To run with multiple assets, the query should be run multiple times with a single asset each (and some assets may be included, others excluded).

@lolabeis - Thank you for the review and feedback - I totally agree.

I have made the Query arguments mandatory and amended the functional logic so that if the if a query contains some data (let's say Upper Maturity Period) that is not set in the specification, the eligibility will return true.

valdensmith (approved):

hugohills-regnosys (approved):

PR #2448, opened 2023-10-12
Lionel Smith-Gordon rename Qualify_RepurchaseAgreement

regnosys-prod-user:

PR #2456, opened 2023-10-17
BUG-557 sec-lending qualification with updated Expectations

PayalKhanna:
original contribution https://github.com/finos/common-domain-model/pull/2448


lolabeis:
I was expecting to see changes in the qualification output for the sec-lending examples, from "SecurityLendingAgreement" to "SecuritiesFinance".

Also the custom FIS Mapper needs to be adjusted so that the sample maps to the new structure, with `assetPayout` positioned under `collateral`. This is the reason why qualification appears to have dropped from the FIS example.

rosetta-source/src/main/resources/cdm-sample-files/functions/sec-lending/reallocation/reallocation-pre-settled-func-input.json
```diff
@@ -1,759 +1,665 @@
...
+            "meta" : {
+              "globalKey" : "3f2255"
+            }
+          },
+          "tradableProduct" : {
+            "product" : {
+              "contractualProduct" : {
+                "productTaxonomy" : [ {
+                  "source" : "ISDA",
+                  "productQualifier" : "SecurityLendingAgreement"
```
lolabeis, line 166: This one still says "SecurityLendingAgreement"? (There may be other ones, I haven't checked them all)

rosetta-source/src/main/resources/result-json-files/fis/isla1.json
```diff
@@ -28,10 +28,6 @@
         "tradableProduct" : {
           "product" : {
             "contractualProduct" : {
-              "productTaxonomy" : [ {
-                "source" : "ISDA",
-                "productQualifier" : "SecurityLendingAgreement"
-              } ],
```
lolabeis, line 34: These ones should continue to qualify (but as "SecuritiesFinance") once you rework the FIS Mapper.

rosetta-source/src/main/resources/cdm-sample-files/functions/sec-lending/reallocation/reallocation-pre-settled-func-input.json
Oblongs: Three files updated

tomhealey-icma (changes requested):
A couple of issues related to repurchase agreements:
1) The source says ISDA instead of ICMA
2) Repurchase agreement qualify_product does not match current repo product structure.
3) Test cases need to change to use correct repo product object.

lolabeis:
@tomhealey-icma Thank you for your review

> The source says ISDA instead of ICMA

As previously responded on the issue linked to this PR, there is no way to implement this taxonomy source switch currently. This will have to come from the strategic enhancement to the qualification feature.

lolabeis:
> 2. Repurchase agreement qualify_product does not match current repo product structure.

The new `Qualify_SecuritiesFinance` function has been modelled exactly on the former `Qualify_RepurchaseAgreement` and matches the current repo structure.

> 3. Test cases need to change to use correct repo product object.

We haven't modified the repo test cases. They use the repo product object that was built and validated by ICMA last year, the only difference being that they now qualify as "SecuritiesFinance".

If you have further changes that you would need to bring to those objects, those are outside the scope of this fix.

lolabeis:
@chrisisla @tomhealey-icma @dshoneisda 

Reflecting on our discussion on Friday 20/10, we have 2 paths forward for this PR and related issue:

1) Either we go with the current repo representation in prod (with `assetPayout` in `collateral`), while bringing sec-lending samples to be consistent with that approach. This is what this PR assumes for the `SecuritiesFinance` qualification logic.
2) Or we modify the repo representation in prod (moving `assetPayout` to `payout`) and keep sec-lending samples as they are. The qualification logic for `SecuritiesFinance` in this PR would no longer work and needs to be changed.

Option 1) is ready, we could merge and release this PR if you're ok to approve it. We could even relax the qualification logic to make it work with `assetPayout` in either `collateral` or `payout`, although this wouldn't be ideal because it means CDM users would have 2 ways of specifying the same product. But it means that samples could be modified in a subsequent minor release as it wouldn't break backward-compatibility.

If you prefer to go with Option 2), @tomhealey-icma I would ask that you supply that PR from your side with the fix and updated samples for all existing repo events.

chrisisla:
Hi @lolabeis,
I would go with Option 1 as this option is more expedient and will definitely allow us to proceed with the intended release this Friday.  I've looked at the changes and will do a proper review now so that this change is ready to go when required.
Thanks! Chris

chrisisla (dismissed):

davidalk:
Hi @tomhealey-icma given @lolabeis responses above are you ok to approve this PR?

tomhealey-icma (dismissed):
Approving as a temporary fix for qualification functions to work. All agreed to address the qualification issue immediately after 5.0 release.

PayalKhanna:
@lolabeis @Oblongs can you please view the latest changes I have updated expectations and a rosetta file to reflect that the assetPayout as moved its location

lolabeis (commented):
There are some `observable` and `settlementOrigin` attributes that seem to have disappeared from some of the sec-lending samples - e.g.:
`full-return-settlement-workflow-func-input.json`

lolabeis (approved):

lolabeis (approved):

hugohills-regnosys (approved):

Issue #2465, opened 2023-10-23
Commodity Delivery structures for Commodity Forwards and Options


JanBaserba:
Commodity delivery was added into CommodityPayout. This has been reported as insufficient for the case of commodity forwards and options, which are represented with the ForwardPayout and OptionPayout respectively instead.

lolabeis:
Closing as completed by #2519

PR #2475, opened 2023-10-31
Jan Baserba Commodity Delivery Option Forward

regnosys-prod-user:


JanBaserba:
LABELS
Category: Modelling
Complexity: Minor Change
Criticallity: High
Target: Development

JanBaserba:
Issue https://github.com/finos/common-domain-model/issues/2465

JanBaserba:
[commodity_delivery_forwards_options_release_notes.md](https://github.com/finos/common-domain-model/files/13215989/commodity_delivery_forwards_options_release_notes.md)

rvincentISDA:
@tomhealey-icma good point on the naming convention used for CommodityDeliveryInformation. It is true that other asset classes e.g. equity and credit can also have delivery features. For now we should focus on the commodity use case and let other use cases address the other asset class when they arise. @JanBaserba will adjust with the prefix Asset.

JanBaserba:
> There is a problem with adding a generic attribute like delivery that is actually defined for a specific domain, in this commodities. This leads back to another problem, in the type OptionPayout, the schedule attribute is of type CommoditySchedule, that means that OptionPayout is really CommodityOptionPayout and not a generic Option. So before going through all the changes, one of two things needs to change, OptionPayout to CommodityOptionPayout or schedule needs a new type DeliverySchedule that support different assets other than just commodities. I would recommend the latter. Referring to underlier of type Product, it should not include delivery CommodityDeliveryInformation but a generic deliveryInformation since underliers cannot be any product.

@tomhealey-icma please check the new commit with the changes @rvincentISDA mentioned that should address these concerns. Thank you.

rvincentISDA (approved):

tomhealey-icma (approved):

davidalk:
Update expectation PR #2516

hugohills-regnosys:
Deleting branch as a new version will be contributed

hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2517

PR #2478, opened 2023-11-06
Jan Baserba Ordinal condition update

regnosys-prod-user:


JanBaserba:
Issue: https://github.com/finos/common-domain-model/issues/2479

JanBaserba:
LABELS
Category: Modelling
Complexity: Minor Change
Criticallity: Low
Target: Development

JanBaserba:
Release notes: 
[ordinal_exists_release_notes.md](https://github.com/finos/common-domain-model/files/13269184/ordinal_exists_release_notes.md)

rosetta-source/src/main/rosetta/base-staticdata-asset-common-func.rosetta
```diff
@@ -0,0 +1,19 @@
...
+version "${project.version}"
+
+func OrdinalExists: <"Requires that, if multiple classification attributes are present, they contain an  ordinal so that they can be sorted. The condition checks that the number of classfication attributes is the same as the number of ordinals. Since each classification attributes can only contain one ordinal attribute, the logic effectively enforces that whenever more than one classification attribute exists, all classification attributes will have an ordinal.">
+    inputs:
+        productTaxonomy ProductTaxonomy (0..*)
+    output:
+        result boolean (0..1)
+    set result:
+        productTaxonomy
+            filter source = TaxonomySourceEnum -> ISDA or source = TaxonomySourceEnum -> EMIR
```
lolabeis, line 11: Why do you need this filter? It looks ad-hoc...
Is that because in other cases, `ordinal` does not exist and so the condition fails? If that is the case, you should instead test whether 'ordinal' exists. See below.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-func.rosetta
```diff
@@ -0,0 +1,19 @@
...
+            then extract 
+                if value -> classification count > 1
```
lolabeis, line 13: There will be certain taxonomy classifications where ordinal is absent, and that's ok (that's why `ordinal` is optional). But when it's present, then **all** classes should have it. You should add a:
```
and value -> classification -> ordinal exists
```

JanBaserba: Do you think we need to add this? To check existence of at least one ordinal we already have:
`then 
                    (value -> classification 
                    extract ordinal exists
                    then all = True)
                else True
            then all = True`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-func.rosetta
JanBaserba: I think the idea here was more to restrict applicability to the classification systems we know that work this way. I agree it may be too ad-hoc... We can remove the filter, which will imply that we assume that in all classification systems never mix a hierarchical classification (with `ordinal`) and a non-hierarchical one (without `ordinal`). I think it could be a reasonable assumption but we added the filter for safety.

LionelSG-REGnosys:
Closed as a part of the GitHub clean-up of stale items for 2025.  If you want to re-open this item, please do so or reach out.

PR #2473, opened 2023-10-27
Roger Guitart RogerGuitart-ETD-Positions-Contribution

regnosys-prod-user:

PR #2510, opened 2023-11-10
Rguitart etd position model with updated expectations

PayalKhanna:
original contribution PR: https://github.com/finos/common-domain-model/pull/2473


PayalKhanna:
@finos-admin 
please add label as below
Category: Modelling
Complexity: Minor Change
Criticality: High
Rosetta
Target: Development

PayalKhanna:
@rogerguitart can you please review these changes and also get the relevant reviewers to review and approve to get this released

rogerguitart:
@PayalKhanna There seems to be an issue with the CDM samples resulting from the FpML mappings in the folders `fpml-5-10/incomplete-processes` and `fpml-5-10/processes`. The problem is that the `counterpartyPositionBusinessEvent` field is being filled for these samples, and more specifically, the `after -> counterpartyPosition -> party -> account` inside it. This should not be happening since my contribution only consists of adding the position model fields to the CDM model (this `counterpartyPositionBusinessEvent` field being one of them), and the mappings rosetta folder has not been touched, so the samples should not change like this. Do you know why this might be happening? Thanks in advance.

rosetta-source/src/main/rosetta/event-workflow-type.rosetta
```diff
@@ -41,6 +41,7 @@ type WorkflowStep: <"A workflow step represents the state of a business event. T
     [rootType]
 
     businessEvent BusinessEvent (0..1) <"Life cycle event for the step. The businessEvent is optional when a proposedEvent or rejection are present.">
+    counterpartyPositionBusinessEvent CounterpartyPositionBusinessEvent (0..1) <"Documents the life cycle event for a position.">
```
PayalKhanna, line 44: @rogerguitart new property counterpartyPositionBusinessEvent is bring added here under counterpartyPositionBusinessEvent. my gues is its because of this its being added to samples

rogerguitart: Okay, but there is also the businessEvent inside WorkflowStep, but that one doesn't appear in every sample under these folders, since both businessEvent and counterpartyPositionBusinessEvent are optional (for example, see the sample `msg ex19 cds execution allocations` under the fpml-5-10/incomplete-processes folder). Maybe I am missing technical knowledge...

hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2513

PR #2512, opened 2023-11-13
Roger Guitart Adding support for Exchange Traded Positions

regnosys-prod-user:


hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2513

PR #2518, opened 2023-11-14
Adding UPI mapping for ProductIdTypeEnum from scheme

hugohills-regnosys:
Adding Support for UPI as a productIdScheme where UPI is expected to be specified with following scheme: http://www.fpml.org/coding-scheme/external/iso4914


hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2521

PR #2531, opened 2023-11-16
Bataller García Eric PTRR - Portfolio Rebalancing

regnosys-prod-user:

PR #2538, opened 2023-11-20
Bataller García Eric PTRR - Portfolio Rebalancing 

davidalk:


davidalk:
Original PR #2531

ebataller (approved):

hugohills-regnosys (approved):

hugohills-regnosys (dismissed):

chrisisla (changes requested):
I don't think this is the correct use of the ```intent```. There has been a lot of discussion about how to use ```intent``` and the usage here is more about the purpose of an execution i.e. the intent of this ```businessEvent``` is an execution, the _reason_ for performing the execution is to rebalance the portfolio but the intention is still an execution.

mgratacos:
> I don't think this is the correct use of the `intent`. There has been a lot of discussion about how to use `intent` and the usage here is more about the purpose of an execution i.e. the intent of this `businessEvent` is an execution, the _reason_ for performing the execution is to rebalance the portfolio but the intention is still an execution.

The primitiveInstruction is Execution but the event is a Portfolio Rebalancing event. It is equivalent to a primitiveInstruction Execution for a Novation business event or for a Clearing business event.

The use of Intent is documented in CDM as: The intent attribute is meant to be specified when the event qualification cannot be programmatically inferred from the event features. 

This is exactly the case for Portfolio Rebalancing event that is supported by an Execution primitive instruction.

ESMA is identifying Portfolio Rebalancing and Portfolio Compression as Post-Trade Risk Reduction events. These events are reportable for EMIR reporting.

chrisisla:
Thanks for the clarification @mgratacos. I know we're trying to reduce the reliance on ```intent``` but your explanation makes sense.

chrisisla (approved):

PR #2545, opened 2023-11-23
Bataller García Eric Natural Person - Circular reference fix

regnosys-prod-user:

Issue #2443, opened 2023-10-11
Pull Request Template


chrisisla:
In order to improve the quality and consistency of Issues and Pull Requests the suggestion has been made that we make use of the Templates available in github.

I propose creating a new general Pull Request template that will be built around the principles and suggestions outlined in the Steering WG on 10th October. The requirements are that a Pull Request must:
- Have a Release Note included within it describing the changes
- Have a link to the Issue describing the work, with the relevant background and documentation
- Include proof of sign off from an active CDM WG e.g. TAWG, a domain WG
- Include proof of community review period being met
- Include proof of having been reviewed for compliance with CDM Design and Contribution Principles
- Have no build errors and is up to date with the master i.e. can be merged into model

Where possible the use of checkboxes is recommended to make it easier for contributors to confirm that they have performed the required tasks. Where a checkbox is ticked then evidence that the work/item has been undertaken will be required. 

For example, if the checkbox for "Release Notes" is ticked then the owner of the Pull Request must ensure that the Release Notes are attached. In some cases the Build and Release process can automatically confirm the existence of these items, but this will not be possible in all cases.

This Issue is only for an initial general Pull Request template. Additional templates can be created in the future if we feel we have different requirements for different types or Pull Requests e.g. model changes, technology changes, documentation changes.

Please provide any feedback you have as a comment to this Issue.

dshoneisda:
I think this works. We should finalise in the steering group and note that some of these may not be mandatory, but expedient e.g. the domain wg sign off

chrisisla:
The PR documentation should also include a section that details and explains any changes that are backwardly incompatible. This will greatly assist people working with the model, and can be accumulated into the production level release notes.

Thanks @lolabeis for the suggestion!

lolabeis:
> The PR documentation should also include a section that details and explains any changes that are backwardly incompatible.

By "PR documentation", we propose that these backward-incompatible changes be documented in a specific _Backward- incompatible changes_ section in the PR's release notes. The existing detailed _What is released?_ section could be broken-down between that section and a section that focuses on new additions.

The backward-incompatible section should include documented examples for implementors to remap CDM objects that may have been persisted using the prior version, with references to sample files included in the CDM distribution.

A good retrospective example to illustrate this approach is the recent pair of "price refactoring" PRs which contain fundamental structural changes to the core CDM component `Price`:
#2292 
#2329 

Under the proposed framework, the release notes would have contained the following section:

_Backward-Incompatible Changes_

[...]

PR #2546, opened 2023-11-23
Bataller García Eric Natural Person - Circular reference fix

JayasriR:
The NaturalPerson type contains a NaturalPersonRole, which references back to the containing type of NaturalPerson, causing a circular reference in the model. This issue is fixed with this contribution.


JayasriR:
Original PR: https://github.com/finos/common-domain-model/pull/2545

JayasriR:
@finos-admin

Please add the following labels to this PR:
Category: Modelling
Complexity: Minor Change
Criticality: Medium/Low
Target: Production

lolabeis (changes requested):
This PR contains a backward-incompatible change (removal of an attribute) so should be rolled into the next dev version of CDM. I have labelled it with `Target: Development`.

Also backward incompatible changes are now required to be specifically documented in the release notes - See #2443

In the meantime, to indicate that an attribute is meant to be retired and should not be used in the prod version without breaking backward compatibility, please use the `[deprecated]` annotation instead.

lolabeis (dismissed):

eacunaISDA (approved):

hugohills-regnosys (approved):

PR #2550, opened 2023-11-28
DSL Update

SimonCockx:


rosetta-source/src/main/resources/cdm-sample-files/functions/repo-and-bond/partial-delivery-input.json
SimonCockx, line 1: Formatted these files so it is easier to review expectation changes.

eacunaISDA (approved):

Issue #2543, opened 2023-11-22
Simplifying valuationHistory record into current valuation and associated  create_  function


gtarres:
The `valuationHistory` defined within the `TradeState` and `CounterPartyPositionState` are designed to track all the updates on the valuation.

However, the accumulation of valuations series within the `valuationHistory` must be reconsidered since daily updates through extended periods can cause a significant increase in the size of CDM objects. This will create significant difficulties in term of managing and processing the corresponding serialised CDM message.

To solve this issue, the proposal is to adjust the `valuationHistory` into a single cardinality `valuation` element. This single `valuation` object will represent the latest valuation associated to the trade state or the counterpartyPosition The previous valuation records will be retrieved by the lineage - from the `before TradeState`.

That is:
1. Adjust from `valuationHistory (0..*)` to `valuation (0..1)` for both the `TradeState` and the `CounterpartyPositionState`.

![image](https://github.com/finos/common-domain-model/assets/118831133/f0182395-8548-4c15-820a-d977b61c9052)

![image](https://github.com/finos/common-domain-model/assets/118831133/6a276dd8-c2a0-47c8-8805-07b53e51680d)

2. Provide a primitive operator that sets/replaces, instead of adding, the updated valuation provided in the `ValuationInstruction` primitive instruction as the current one in the `TradeState` or `CounterpartyPositionState`.

![image](https://github.com/finos/common-domain-model/assets/118831133/eab1612c-87c7-44bd-993e-ba77e90ed733)

![image](https://github.com/finos/common-domain-model/assets/118831133/4949fdee-72c4-4e75-b199-e80b1af268e8)

gtarres:
As per the CDM Maintainers Review of PR/Issues meeting, hold on the 28th of November, the valuationHistory will remain unchanged. The Create_Vauation function will support both approaches by having an additional input allowing the user either adding the new updated valuation to the history (keeping the old tracks) or replacing all the history by the new value (removing the old tracks).

![image](https://github.com/finos/common-domain-model/assets/118831133/67a1654e-b1c1-4d95-9d8e-d15bf6ed0305)

gtarres:
Contributed in PR #2552

lolabeis:
Completed by #2552

PR #2552, opened 2023-11-29
Georgina Tarres TH_Valuation_Instruction

regnosys-prod-user:


gtarres:
Linked to ISSUE #2543  

### Release notes:
[CDM_release_notes_Valuation.md](https://github.com/finos/common-domain-model/files/13502637/CDM_release_notes_Valuation.md)

### Proposed labels:
Category: Modelling
Complexity: Minor change
Criticality: Medium/Low
Target: Developement

gtarres:
Required review from @lolabeis. To be added in the prod release.

tomhealey-icma (changes requested):
I don't believe the addition of a valuation object should constitute a change in trade state. I would suggest adding ValuationState without replicating history before to after on every change.

lolabeis:
@tomhealey-icma in response to your comments:

> I don't believe the addition of a valuation object should constitute a change in trade state.

The `TradeState` object was introduced specifically to separate information that **does not** change the `Trade` object (cashflows, observations), but which may nonetheless impact a trade, from information that **does** change the `Trade` object (any sort of amendment).

In an extension of this design, valuation was added to `TradeState` as part of an earlier approved change. It follows that changing / updating it means changing the `TradeState` object, a role that is currently fulfilled by a `Primitive`. The change proposed in this PR is consistent with the current design.

> I would suggest adding ValuationState without replicating history before to after on every change.

Can you clarify your suggestion, keeping in mind that any change must be backward-compatible? What would be the difference between `ValuationState` and `Valuation`, and would `ValuationState` be included in `TradeState`?

lolabeis (approved):

tomhealey-icma (approved):
Although this is not a lifecycle event change and shouldn't necessitate a tradeState change this can go ahead for 5.0 but should be revisited in 6.0.

eacunaISDA (approved):

PR #2556, opened 2023-11-30
JEAN BAPTISTE ZIADE FRAGMOS_TermsChange_TransactionAdditionalTerms

regnosys-prod-user:


tomhealey-icma (changes requested):
As discussed on this issue, adding legal agreement to terms change instruction will not work at the product level.

JBZ-Fragmos:
no object in yet defined "at product model" for now, so there can be no link between this PR that is to update an object which already exists and thus requires a way to be updated, and the consideration how to update an object that does not not exist, which prevents any concrete determination how to update it

tomhealey-icma:
Please add labels. If cannot add them to the PR, put them in a comment and one of the maintainers will add them.

JBZ-Fragmos:
ok

i've updated Labels - 

image

rvincentISDA:
See my observations in the issue tkt. It should be possible to amend the additionalTransactionalTerm but the current proprosal won't be deterministic.

tomhealey-icma:
You are missing labels for category and criticality.

JBZ-Fragmos:
@tomhealey-icma thanks, i've completed labels with missing ones

@rvincentISDA sorry not sure why it would not be deterministic ? not sure i understand your point, yet maybe I do then please consider "v2" and compare to initial one : does "v2" answer your need about being "deterministic" in case of multiple root objects at LegalAgreement level ? 
changes i've made = change type + added one new condition

image

JBZ-Fragmos:
taking into account last comment from Ruddy, i've also desifned "v2" below - to be discussed as well 

image

tomhealey-icma:
I cannot support adding transcationalAdditionalTerms to the termsChangeInstruction. It was clearly not the intent of this Type to include changes to objects outside the product. Of the options Leo proposed, option 2 is most preferred and most consistent with the design principles.

JBZ-Fragmos:
ok, 

i will draft a proposal aligned with "option2" (see separate discussion on related GitHub Issue)

then we will review it as new starting point for our discussions

so will also create a new PR, therefore i'm closing this one

JBZ-Fragmos:
i will draft a proposal aligned with "option2" (see separate discussion on related GitHub Issue)

then we will review it as new starting point for our discussions

so will also create a new PR, therefore i'm closing this one

PR #2561, opened 2023-12-01
Fix CVE scanning

SimonCockx:


.github/workflows/cve-scanning.yml
```diff
@@ -3,6 +3,8 @@ name: CVE Scanning for Maven
 on:
   workflow_dispatch:
   push:
+    branches:
+      - master
```
SimonCockx, line 7: Filter `push` events on master to prevent triggering CVE scanning twice.

The triggers now work as follows:
- The `pull_request` event makes sure that PRs must always succeed CVE scanning.
- The `push` event filtered on `master` will run CVE scanning on the master branch every time that it changes, as an additional check.

hugohills-regnosys (approved):

PR #2583, opened 2023-12-11
Nicholas Moger  Create_Execution missing lotIdentifier

regnosys-prod-user:


nicholas-moger:
> There are a few places in the CDM that instantiate an `ExecutionInstruction` object - e.g. in `Create_Exercise`. For completeness, these functions should include populating the `lotIdentifier` attribute as well.

Hi Leo - I am hesitant to start adding it to other functions at least for now, as there would need to be further code within those functions on how to apply the exercise with respect to tradeLots like exists in quantityChange. As per below is looks like Create_Excercise is meant to fully terminate the option, so it would need special handling for Partial_Exercise i think. 

The main issue now is we cannot ascertain what extensions are needed for other functions because the base trade  is lacking the lotIdentifier to begin with. We are not using CDM prod branch yet so if this was merged only to Dev branch we could spend time understanding the related impacts on other functions if needed. Would that work?

func Create_Exercise: <"Defines the process of putting into effect the rights specified in an options contract, such as to buy or sell a security.  Once exercised the option contract is terminated.">

lolabeis:
Closing as superseded by a Pull Request against `dev` instead of `prod`: #2591

PR #2597, opened 2023-12-21
Operand type additions for Spread and others

tomhealey-icma:
Additional operand types are needed for Spread, BorrowFee, MarkUpDown, VenueExecutionFee, ClearingFee and Tax
Added Spread, BorrowFee, MarkUpDown, VenueExecutionFee, ClearingFee, Tax and other enums to `PriceOperandEnum`.

Also added Yield to PriceTypeEnum.


Oblongs:
Reviewed PR with @tomhealey-icma and @hugohills-regnosys.  The technical implementation is fine however we feel that the proposed Enums need syndication and review from a business perspective with the other trade associations.
The ongoing governance and population of enums is a broader topic that needs to be tackled.

PR #2610, opened 2024-01-05
Repo API Functions

tomhealey-icma:
The CDM supports repo products and lifecycle events but integration with existing applications and systems
can be complicated due to the complex nature of the CDM trade and product structure. Accessing basic repo
data element like repo rate or haircut involves constructing data paths deep into the product. To address this
difficulty and make it easier for firms to adopt the CDM this enhancement includes a suite of custom repo functions
built using CDM modelling methods so that the functions are standard and consistent across programming languages and
platforms.

New files with the name product-api-repo- are being added that includes functions for repo dates, parties,
collateral, and tranactions. The file list includes:
- product-api-repo-common-func.rosetta
- product-api-repo-collateral-func.rosetta
- product-api-repo-datetime-func.rosetta
- product-api-repo-parties-func.rosetta
- product-api-repo-trade-func.rosetta
- product-api-repo-product-func.rosetta

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,53 +1,43 @@
-# _Product Model - ISO Country Code Enum Update_
+# _Product model - Repo Functions_
 
 _Background_
 
-Rosetta has a feature for its enumerations that enables, if a certain enumeration is directly related to an FpML or ISO coding scheme, to label that enumeration with the corresponding coding scheme canonical URI, so every time that coding scheme is updated, the enumeration will be automatically updated.
+The CDM supports repo products and lifecycle events but integration with existing applications and systems
+can be complicated due to the complex nature of the CDM trade and product structure. Accessing basic repo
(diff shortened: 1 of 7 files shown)


Oblongs:
@tomhealey-icma , @hugohills-regnosys and I discussed this PR.  Key objective is to help ICMA's community to adopt and use CDM more easily and, in this context, tooling to help is appropriate. But we feel that it is more appropriate that these APIs and capabilities are hosted in a separate repository, rather than the main, shared CDM on Finos.  We do not want the core CDM model to take on too much use case or implementation-specific functions but these can be made available in a standalone companion repository.
There may be a small set of the proposed APIs which are appropriate for inclusion in CDM, particularly where they interact tightly with the Product Qualification logic.
Tom H is going to pursue building a new repo for the ICMA community.

PR #2600, opened 2024-01-03
Gerard Opazo Enhanced Price mappings

regnosys-prod-user:

PR #2630, opened 2024-01-16
Gerard Opazo Enhanced Price mappings with release Note

PayalKhanna:
original Contribution https://github.com/finos/common-domain-model/pull/2600


PayalKhanna:
@finos-admin 
Labels:

Category: Modelling
Complexity: Small/Bug/Mapping
Criticality: Low
Target: Development

dshoneisda (approved):

lolabeis:
@gopazoTH Don't you need that against Prod as well, so that we can use it in DRR?

gopazoTH:
@lolabeis yes, we need it into DRR as well, basically the main sample to test knownAmount is there, so its good to place it in Prod.

PR #2647, opened 2024-01-24
Nicholas Moger Create_Execution Update

regnosys-prod-user:


nicholas-moger:
Re-submitting PR against prod branch as discussed agreed with ISDA/Regnosys. This change is to ensure the originating lotIdentifer can be added to the executionInstruciton such that increase/unwind can be used without automastically creating a new tradeLot (i.e. you can increase/decrease an existing tradeLot rather than using replace).

JayasriR:
Superceded by: https://github.com/finos/common-domain-model/pull/2649

Issue #2669, opened 2024-02-01
CDM Derivatives Products and Business Events Working Group - February 27th, 2024


eteridvalishvili:
# CDM Derivatives Products and Business Events Working Group Minutes 
Meeting Host: David Shone, ISDA

February 27th, 2024 - 11:30 am EST / 4:30 pm GMT

## Untracked attendees
- Fullname, Affiliation, (optional) GitHub username

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies). 

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda
- [x] Convene & roll call (5mins)
- [x] Display [FINOS Antitrust Policy summary slide](https://community.finos.org/Compliance-Slides/Antitrust-Compliance-Slide.pdf) 
- [x] Review Meeting Notices (see above)
- [x] Approve past meeting minutes (N/A)
   - [x] https://github.com/finos/common-domain-model/issues/2666
 - [x] Next planned CDM minor release (`CDM 5.6`)
   - [x] Proposed release date: ASAP (possibly by Tues 27/2 COB if no impediment)
   - [x] Content to be ratified:
     - [x] Master Agreement type #2696 
     - [x] Qualification fix for ETD underlier #2705 
   - [x] Mapping or bugfix update (no model change):
     - [x] SettlementType mapping fixes #2701
     - [x] Commodity Classification mapping fixes: #2715 
[...]

PR #2749, opened 2024-03-06
Manuel Carrera CDM Commodity Physical Option dev

regnosys-prod-user:


manucarreramoreno:
# Product Model - Commodity Physical Options

_Background_

This release provides qualification support for Commodity Physical Option products, which were not supported until now. This kind of products will be represented very similarly to cash settled options: an `optionPayout` with an strike that can be fixed or floating, a commodity `underlier` and the delivery information specified in the attribute `delivery`. The only difference will be that `settlementType` will be `Physical`.

_What is being released?_

* Added two qualifying functions: `Qualify_Commodity_Option_Cash` and `Qualify_Commodity_Option_Physical`.

* Updated the qualifying function `Qualify_Commodity_Option`, which was formerly used to qualify only cash settled options. Now it uses both qualifying functions specified above.

* Simultaneously, updated the qualifying functions `Qualify_Commodity_Swap_FixedFloat` and `Qualify_Commodity_Swap_Basis` so they do not check that the underlier is a commodity. That is redundant information since it is already been checked in `Qualify_AssetClass_Commodity`.

_Review directions_

* In the Rosetta platform, select the Textual Browser and inspect each of the changes identified above.

lolabeis (approved):
Approved at the [CDM Derivatives Products and Business Events Working Group - February 27th, 2024](https://github.com/finos/common-domain-model/issues/2669)

eacunaISDA (approved):

PR #2796, opened 2024-03-20
Update use-cases and sidebars

chrisisla:
Add a title to use-cases document

Add eligible collateral, repurchase agreement and securities lending to the sidebar (otherwise sidebar does not display on these  documents and the default Prev/Next navigation will not work)


maoo (dismissed):
LGTM!

valdensmith (approved):

PR #2801, opened 2024-03-21
[5.x.x] STORY-3174: Enhanced commodity classification mapping

hugohills-regnosys:


hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2803

PR #2810, opened 2024-03-26
Chris Rayner Add new enum to AvailableInventory

regnosys-prod-user:


tomhealey-icma (changes requested):
The enum name, InventoryTypeEnum: is not specific enough for the type of enum values. I would suggest, SecuritiesLendingAvailableInventoryEnum.

chrisisla:
Hi @tomhealey-icma ,
How about ```AvailableInventoryTypeEnum```, as the enum will be used in the ```AvailableInventory``` type, and that is not specifically associated to Securities Lending, as per the discussions at the CRWG where the original PR was presented.

chrisisla:
> The descriptions for the enums explicitly refer to lending or suggest it's purpose is lending: 

Correct, because at this time the use cases I am proposing are for securities lending. When others want to extend the use cases then they can add more options to the enumeration. 

If you prefer I could try to make the enum descriptions more generic?

tomhealey-icma:
@chrisisla - This issue is not just the descriptions, "Available" means many different things depending on the context (available to trade, available to delivery, available to lend). My suggestion was to change the Enum level name so its specific to securities lending but it you want to keep the Enum AvailableInventoryTypeEnum then the enum values should be more specific:
AvailableToLend
RequestToLend
HoldToLend

chrisisla:
Hi @tomhealey-icma ,

Sounds good, I can change the _Available_ and _Request_ enums to _AvailableToLend_ and _RequestToBorrow_ as when a lender broadcasts their availability it is letting the market know what they have available to lend, but when a borrower requests a security they are asking to borrow securities from a lender.

I'll also remove the _Holding_ enum as this is unrelated to availability, it is more for where a party wants to describe a list of the securities that they hold, so it could be renamed to _SecurityHolding_, but I will remove it for now.

I'll reinstate the original comments for the enums as these are now specifically for lending use cases.

tomhealey-icma (approved):

dshoneisda (approved):

hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2872

PR #2830, opened 2024-04-04
Jbziade acc decumulator temporay drr prod release

PayalKhanna:
original contribution https://github.com/finos/common-domain-model/pull/2827


eacunaISDA (dismissed):
code comments should really be in the descriptions

hugohills-regnosys (approved):

PR #2831, opened 2024-04-04
Lionel Smith-Gordon ARTForce initial proposals

regnosys-prod-user:


rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -5,11 +5,74 @@ import cdm.base.*
...
+    security Security (0..1) <"A specific type of Asset that is issued by a party to be held by or transferred to others.">
+    condition OneOfAsset:  <"An Asset must be of one and only one of the following.">
+        required choice assetPool, basket, cash, commodity, derivative, digitalAsset, loan, security
+// Using the 'required choice' condition, rather than 'one-of' as it allows other shared attributes to be added to the data type.
+
+// To improve the model, it is proposed to replace this design with a new Union construct.  The revised modelling:
+//
+// union Asset:  <"An Asset is defined something that can be held and transferred in the financial markets.">
+//      AssetPool       <"An Asset that consists of a pool of other assets that back an asset backed security.">
+//      Basket          <"An Asset that is a collection of other Assets.">
```
JBZ-Fragmos, line 54: **DISAGREE** 
- with Basket description that is saying **Basket Components** shall be of type **Asset**

**PROPOSAL** 
- **Basket Components** shall be of type **Observable**

**RATIONALE** 
cumulative reasons below :
  - Basket of CurrencyPairs may exist (e.g. Currency CorrelationSwap)
  - as well as a Basket of Index (less "vanilla" occurrence, but may exist in some Structured Product) 
  - more generally, there is eventually no theoretical reason no to structure a payout that would observe a Basket of **any** mix...

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -5,11 +5,74 @@ import cdm.base.*
...
 import cdm.base.math.*
 import cdm.base.staticdata.party.*
+import cdm.base.staticdata.identifier.*
+import cdm.base.staticdata.asset.rates.*
 import cdm.observable.asset.*
 import cdm.product.template.*
 
-type AssetPool: <"Characterizes the asset pool behind an asset backed bond.">
 
+// ART-01
```
JBZ-Fragmos, line 14: **AGREE** 

- with ALL attributes proposed for **new type Asset**

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -95,13 +158,43 @@ type IndexReferenceInformation: <"A class defining information related to Index"
     condition IndexAttributes: <"A required choice condition for either a floating rate or inflation rate index.">
         indexName exists or indexId exists
 
-type Index extends ProductBase: <"Identifies an index by referencing a product identifier.">
+// ART-02 Observable and Index refactoring
```
JBZ-Fragmos, line 161: **PARTIALLY AGREE**

- AGREE with attributes listed below for new type Observable
- REMARK current list may need to be extended with missing one that is Fixed Rate 

**PROPOSAL**
- add one addionalt attribute for Observable that would be of type FixedRate
- **RATIONALE** : cumulative reason below :
  - considering that TradeLot->PriceQuantity->Observable is the generic place where "arguments" of Payout shall be defined (with metadata location) for being referenced in Payout (with metadata address), then for any kinds of payout argument all would be an Observable, except the Fixed Rate, that is to say accross the whole CDM model FixedRatePayout would be the unique Payout component which argument would not be an Observable (by "argument" please hear "X" in "F(X)" when "F" is the functional formula of a given Payout)
  - despite the fact it is "already observed" (because it is given as Fixed value), a Fixed Rate is all in all homegneous _in nature_ to floatingRateIndex, that is to say it is same class "Rate", that is to say fact a Rate is already Fixed or not Fixed yet ("Floating") does not make a difference from a payout "F(X") point of view

rosetta-source/src/main/rosetta/observable-asset-type.rosetta
```diff
@@ -184,9 +184,19 @@ type Price extends PriceSchedule: <"Specifies a price as a single value to be as
     condition AmountOnlyExists: <"The amount must exist when the price represents a single value, and steps must be absent.">
         value exists and datedValue is absent
 
+// ART02 Observable and Index refactoring
```
JBZ-Fragmos, line 187: same comment as prevous one above 

**PARTIALLY AGREE**

- AGREE with new type Observable and current list of attributes
- REMARK beleive there is missing one attribute in the list

**PROPOSAL**

- add one addional attribute for Observable that would be of type **FixedRate**
- RATIONALE : see details in previous comment

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
JBZ-Fragmos, line 161: **PROPOSAL**
- whenever/anywhere an Observable type is used somewhere in the model, notably in Payout, as well as any other key payout argument such as Price or Quantity, then it shall be [**metadata address**], so that such Observable originally defined in TradeLot->PriceQuantity->Observable with [**metadata location**] can be referenced
- just rephrasing same idea : **booking method** below shall apply in generic manner accross whole model  : 
  - all Observable, Price, Quantity shall be defined in TradeLot->PriceQuantity with **metadata location**
  - then any usage of such type in Payout shall be referencing them with **metadata address**

**RATIONALE**
- as of today, model is **inconsistent** about "booking method" above described, beacuse it is partially in place for some Payout components (**_examples 1_**), and not in place for other ones (**_example 2_**)
- there are multiple **negative side-effects** for cases where referencing is not place, notably **DRR** e.g. see **_example 2_** below about ASIC new data fields for Barrier Price items 58 and 59
- existence of generic/consistent referencing from TradeLot items to Payout items is MILESTONE prerequisite for preparing **SmartContract** features (= input -> resolvingPayout -> output) which requires that input can be resolved in the place it is used, typically metadata location shall be leveraged for this purpose for producing resolved output in metadata addresse - hence the need to have this structure in place accross all Payout

**_Example 1_** - Referencing TradeLot->PriceQuantity in Payout terms exists
_non-exhaustive list_ 
- Payout->Underlier->[...]->product Identifier
- Payout->rateSpecification->rateOption
- Payout->rateSpecification->spreadSchedule
- Payout->ResolvablePriceQuantity->Quantity
- Payout->ResolvablePriceQuantity->Price (by the way, highly recommand to remove this one, in order to ensure that any Price component is an attribute of core payout, in order for CDM to be aligned ith generic payout schema that is Quantity x CorePayout... in regards of which CDM situation is confusing, because we have PriceQuantity x Payout, where some Price might be out of Payout per se and Payout may not include any core formula component at all, for instance the weird case of fixedPricePayout, which implicit functional formula would be = Quantity x Price hence no Underlier neither Fixed Price in core payout exist...)

**_Example 2_** - Referencing TradeLot->PriceQuantity in Payout terms is **missing**
_non-exhaustive list_ 
- trigger and triggerPercentage are of type "number" in Trigger type : this one is causing issue for DRR
- optionStrike is Price but with no metadata
- volatilityStrike, same comment
- varianceStrike, same comment
- valuationPriceInitial, same comment

Oblongs: Makes sense.

Do you see "fixed rate" as a new attribute of `Observable` or another type of `Index`?

That is, either:

`type Observable: `
`    asset Asset (0..1)  `
`   quotedCurrencyPair QuotedCurrencyPair (0..1)`
`   index Index (0..1)`
`   fixedRate number (0..1)`
`      condition: one-of `

or 

`type Index: <"Identifies an index by referencing a product identifier.">`
`    fixedRate number (0..1)  `  
`    floatingIndex FloatingIndex (0..1) `
`    inflationIndex InflationIndex (0..1) `
`   equityIndex EquityIndex (0..1)`
`    creditIndex CreditIndex (0..1) `
`     condition: one-of`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oblongs: Your proposals will be included in the next version of the Open Contribution (and in the Task Force meeting on 11 April) with designation "ART-04".

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
JBZ-Fragmos: would recommand to define it as a new **attribute of Observable**, because there is no issuer and no other features, per se that is only what it is : a fixed rate... 

yet would also avoid usage of type _number_ 
- that is a "poor" type if i may say, compared to PriceSchedule type where datedValue might be defined if need be, etc. 
- also let's consider that type used in current CDM for fixed rate (when defining a fixed rate via TradeLot->PriceQuantity to be referenced then in Payout leg) is PriceSchedule type, so using poor type such as number instead would deprecate such original type for no good reason in my view...

that being said, proposal would be as below :

type **Observable**: 
   asset Asset (0..1) 
   quotedCurrencyPair QuotedCurrencyPair (0..1)
   index Index (0..1)
   fixedRate **PriceSchedule** (0..1)
     condition: one-of

tomhealey-icma:
First point is that observable is just a reference to how the model is using a product and not a separate product or asset class as is being defined here. A product can be tradeable or observable and being observable doesn't mean it's not a tradable product. 

Second, there is no need to separate asset from product, this is just semantics and not material. 

A basket is a separate product because it may have it's own identifier, price and other factors that make it more than a container but it should exists as a product.

The list of tradable product should be defined as:
Equity
Debt
Commodity
Options
Futures
Structured
Swap
Currency
Financing
Basket
Fund

My previous points;

- DigitalAsset is not a separate product. It represents the form of the product.
- ContractDetails should be included in ProductBase.
- Regarding priceQuantity in tradeLot, I suggest moving Product under tradeLot and setting prices directly in Product for each lot. The current structure requires using attribute to match a price and quality to a specific attribute in the product. This is not flexible or sustainable when creating complex products. With many price components.
- Loans, index, assetPool, Security are defined as products, but they do not include any economics or product base. The attributes of these products are very inconsistent as to what they are supposed to represent. So every product needs to be under contractualProduct and what is the purpose of contractualProduct if there is no non-ContractualProduct?

Oblongs:
> First point is that observable is just a reference to how the model is using a product and not a separate product or asset class as is being defined here. 

We are not saying that `Observable` is separate from `Product` or `Asset` but it is purely modelling things that can be "observed", with supporting data types, including `observation`.  As modelled, foreign exchange rates and indices are examples of things that can be observed but cannot not be held, transferred or traded, hence the different modelling.

> A product can be tradeable or observable and being observable doesn't mean it's not a tradable product.

An `Observable` cannot be traded directly but clearly can be a component of a product that is created through negotiation (and composition in the CDM).

> The list of tradable product should be defined as: Equity Debt Commodity Options Futures Structured Swap Currency Financing Basket Fund

We haven't discussed this yet within the Task Force as we agreed we'd focus first on the base set of asset data types.  Your list reads like a good definition of a set of financial products that might be defined in a dictionary that attempts to document terms to educate participants.   (When I use the word "product", with a small "p", I am referring to the general market term for all financial "things" that market participants deal in.) 
In the CDM, we do not want to discretely model each different type of financial product as a different data type.  We want to follow the [CDM Design Principles](https://cdm.finos.org/docs/contribution#normalisation) of composition, normalisation and modularisation to assemble complex financial products from consistuent parts.  What we are attempting to do in this Task Force is to redefine the lowest level of these parts as `asset` with the base building blocks within it.  For example, equity products can be composed using the `security` asset data type while structured and swap products are composed from a combination of many different `asset` types and `observables`.

> * DigitalAsset is not a separate product. It represents the form of the product.

IMHO, there are different types of digitised (ie cryptographically-represented) financial products:

- the use of blockchain and distributed ledger technologies to create digital representations of real-world assets in order to improve market efficiency, examples being industry initiatives to create tokens to represent securities or cash that are then settled electronically.  I agree that these are not a separate class of asset but a form of an existing asset and would most likely be represented in the CDM as a an attribute of the `Asset`.
- new products, which exist only digitally and are not backed by physical assets; bitcoin and ethereum are the most obvious examples as they exists only as a crytpographic key on a public blockchain, are not backed by and cannot be represented by any other asset.  Such financial products would be modelled using the proposed `DigitalAsset` data type.
- a possible third type would be asset-backed cryptocurrencies such as stablecoin: these are products that have been created digitally, are represented on a blockchain, and are supposedly backed by real-world assets.  Perhaps these might be modelled in the CDM in the future using payouts and underliers to represent the asset relationship.  However, I do not believe participants are seeking to model these products in the CDM at this time.

The following topics will be covered in future ARTForce discussions:
> * ContractDetails should be included in ProductBase.
> * Regarding priceQuantity in tradeLot, I suggest moving Product under tradeLot and setting prices directly in Product for each lot. 
> * what is the purpose of contractualProduct if there is no non-ContractualProduct?

tomhealey-icma:
@Oblongs Regarding your comments:

1.An Observable cannot be traded directly but clearly can be a component of a product that is created through negotiation (and composition in the CDM).

If this is the case then observable should be redefined so as not to identify a asset or product but what is being observed.

Observable is defined as:
    rateOption FloatingRateOption (0..1) <"Specifies a floating rate index and tenor.">
        [metadata location]
    commodity Commodity (0..1) <"Identifies a commodity by referencing a product identifier.">
        [metadata location]
    productIdentifier ProductIdentifier (0..*) <"Comprises of an identifier and a source. The associated metadata key denotes the ability to associate a hash value to the ProductIdentifier instantiations for the purpose of model cross-referencing, in support of functionality such as the event effect and the lineage.">
        [metadata location]
    currencyPair QuotedCurrencyPair (0..1) <"Describes the composition of a rate that has been quoted or is to be quoted, including the two currencies and the quotation relationship between the two currencies.">
        [metadata location]
    optionReferenceType OptionReferenceTypeEnum (0..1) <"The underlying contract which is referenced when determining the final settlement price of the instrument. Eg. Rolling Front Month Future; Spot etc.">
    
   
This list is very inconsistent. Observable should define what is being observed, including price or yield, tenor, method, etc. This list just has commodity, currency pair, product...

2. We haven't discussed this yet within the Task Force as we agreed we'd focus first on the base set of asset data types. Your list reads like a good definition of a set of financial products that might be defined in a dictionary that attempts to document terms to educate participants. (When I use the word "product", with a small "p", I am referring to the general market term for all financial "things" that market participants deal in.)

It seems like your list just cherry picked a few products
basket
cash
commodity
loan
security

These products are defined in the namespace of product.template under TradableProduct because they represent "product templates" of things that are tradable. 

If were going to consider ...all financial "things" that market participants deal in.).. then other products should be considered. 

3. For digital assets we need to consider digital representations separate from native digital assets.

Oblongs:
> observable should be redefined so as not to identify a asset or product but what is being observed.

Yes; in the current proposal, it is:

`type Observable:`
`    asset Asset (0..1)  `
`    quotedCurrencyPair QuotedCurrencyPair (0..1)`
`    index Index (0..1)`
`    condition: one-of` 

I've posed this question to the group:

* Should Observable contain Asset or just an AssetIdentifier?  (Current Observable contains ProductIdentifier).

> These products are defined in the namespace of product.template under TradableProduct because they represent "product templates" of things that are tradable.

Yes, as acknowledged in my proposal, we need to look at the namespaces.

> 3. For digital assets we need to consider digital representations separate from native digital assets.

Yes.

Oblongs:
This PR has been superceded by a new one:  https://github.com/finos/common-domain-model/pull/2860

PR #2850, opened 2024-04-10
Manuel Carrera CDM Qualification of Bond Options and Forwards DEV

regnosys-prod-user:


manucarreramoreno:
# Product Model - Bond Option and Forward Qualification

_Background_

A previous release caused a regression in the qualification of Bond Forwards and Bond Options, that were not qualified anymore due to an update to the `Qualify_AssetClass_InterestRate` function. This release fixes that.

_What is being released?_

* Reintroduced to `Qualify_AssetClass_InterestRate` the possibility to have a security with `securityType = SecurityTypeEnum -> Debt` as underlier of the option or forward.

_Review directions_

In Rosetta, select the Textual View and inspect each of the changes identified above

PR #2860, opened 2024-04-15
Lionel Smith-Gordon ARTForce v2

regnosys-prod-user:
Supersedes #2831


iansloyan:
Hi @Oblongs is there a document that sets out the intentions of this work that I can use as a starting point to understand what the task force is doing please?

tomhealey-icma (changes requested):
Adding my comments from the prior PR that was closed, 2831:

1. ContractDetails should be included in ProductBase since any tradable product will/may have a contract.

2. Observable is just a reference to how the model is using a product and not a separate product or asset class as is being defined here. A product can be tradeable or observable and being observable doesn't mean it's not a tradable product. Also a product is just a reference what is actually observable is a price, yield or other observable and measurable value.

3. There is no need to separate asset from product, this is just semantics and not material or structural. As pointed out in the last meeting, an asset is not tradable except when it becomes a product. 

Loans, index, assetPool, Security are defined as products, but they do not include any economics or product base. The attributes of these products are very inconsistent as to what they are supposed to represent. So every product needs to be under contractualProduct and what is the purpose of contractualProduct if there is no non-ContractualProduct?

A basket is a separate product because it may have it's own identifier, price and other factors that make it more than a container but it should exists as a product.


4. Regarding priceQuantity in tradeLot, I suggest moving Product under tradeLot and setting prices directly in Product for each lot. The current structure requires using attribute to match a price and quality to a specific attribute in the product. This is not flexible or sustainable when creating complex products. With many price components.

lolabeis:
Superseded by #2918

PR #2880, opened 2024-04-22
Hugo Hills Update descriptions on the rounding enum and funcs

regnosys-prod-user:


hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2917

PR #2918, opened 2024-05-09
Lionel Smith-Gordon ARTForce_V3

regnosys-prod-user:


JBZ-Fragmos:
@Oblongs 

- type **_relatedExchange_** is missing

we had agreed not only exchange but also relatedExchange would be added to the model with an optional cardinality, and it is currently missing in PR, see an example of termsheet below that is very classic so please add this component _**relatedExchange**_ we definitly need it...

image

- **_exchange_** and **_relatedExchange_** are missing attribute for creditIndex and for equityIndex

you can see it again obviously with the termsheet : the basketConstituent is of type index i.e. i need to define exchange and relatedExchange as attribute of the type equityIndex
i guess the need is the same for creditIndex, and more generally for other kinds of Index

so all in all i beleive these two attribiutes **should belong to type Index**, eventually (with optional cardinality, of course...)

JBZ-Fragmos:
@Oblongs is that possible / would you agree to included as part of your PR these 2 minor changes in names below, as part of "cleaning" the concept of "observable" ? Else would have t create PR etc. for such minor point  

that is about attributes **pricingTime** and **pricingTimeType** in ObservationTerms

**proposed change :**
- update as **observationTime** 
- and **observationTimeType**

**rationale :** 
- "observation" is more "neutral" concept rather than "fixing"
(which per se could be discussed/challenged wjether appropriate as part of an observation (because one may distinguish obervation and the fact you use it for "fixing" or for other purpose I mean "fixing" may relate the fact a price is being fixed on market screen for instance but this same term may be used a but like "reset" to say that the trade product is being "fixed" etc.)
- also such prefix is more aligned with the one in the name of the parent object that is "**observation**Terms"

... please let me know if you would agree to include this change in your PR...

JBZ-Fragmos:
@Oblongs 
last proposal would suggest to include is to slitghtly refactor **Transfer** 

**proposed change :** 

- add **new attribute Price (0..1)** that is missing for specifying the price of deliverable asset
- upgrade **current cardinality** of **quantity** NonNegativeQuantity (1..1) **up to (1..2)**, that is for the purpose of being able to represent not only the global money amount of the transfer but also the number of units

**rationale :** 

- current representation of transfer is minimal/restricted/too poor, because _**you can just represent 1 single quantity**_ so you are forced to choose between 2 types of quantity - whereas **most systems would record _both_** as a complementary info
   - **_either_** to represent the quantity unit e.g. **10 000 stocks** that is physically transfered... 
   - **_else_** the quantity that correspond to the equivalent money amount, say "notional amount" e.g. **50 000 €** with for instance price per each 5€)
- also would record the price e.g. **5€** which the units are delivered
- and all **the 3 components are consistent with each others**, of course : **Nominal (50 000€) = Nber ofUnit (10 000 stock) x Price (5€/stock)** - hence propsed change so that all this info may optionally exist in Transfer if need kbe for business to represent in most exhaustive manner

Oblongs:
> * **_exchange_** and **_relatedExchange_** are **missing attribute** for **_Index_**
> 
> you can see it again obviously with the termsheet that each basketConstituent is of type equityIndex i.e. need to define _exchange_ and _relatedExchange_ as attribute of _equityIndex_ yet i guess the need is the same for creditIndex, and more generally for other kinds of Index, hence conclusion that these two attributes **should belong to type Index**

I will update the definition of `IndexBase` in the Phase 1 PR:

```
type IndexBase extends Identifier: <"Identifies an index by referencing a product identifier.">
    name string (0..1)  <"A description of the Index.">
    provider string (0..1) <"The organisation that creates or maintains the Index.">
    assetClass AssetClassEnum (0..1) <"The Asset Classs of the Index.">
    exchange string (0..*)  <"If the Index is listed, defines the public exchange(s) of the listing.">
        [metadata scheme]
    relatedExchange string (0..*)  <"Provides the related Exchanges, if applicable.">
```

See https://github.com/finos/common-domain-model/pull/2985

Oblongs:
> about attributes **pricingTime** and **pricingTimeType** in ObservationTerms
> 
> **proposed change :**
> * update as **observationTime**
> * and **observationTimeType**

I have added this to the scope of the next ("Phase 2") PR that will come next week.

JBZ-Fragmos:
thanks @Oblongs for your feedbacks about including 1st and 2nd requests (respectively : "exchange & relatedExchange new attributes for index" and "renaming of pricingTime & pricingTimeType in ObservationTerms")

have you had a chance to review as well the 3rd one about "Transfer" refactoring ("include Price & update cardinality if Quantity") ?

lolabeis:
@JBZ-Fragmos
> **proposed change :**
> 
> * add **new attribute Price (0..1)** that is missing for specifying the price of deliverable asset
> * upgrade **current cardinality** of **quantity** NonNegativeQuantity (1..1) **up to (1..2)**, that is for the purpose of being able to represent not only the global money amount of the transfer but also the number of units

Your use-case is represented by **two** transfers, one for the asset and the other for cash (also an asset). Each is specified with their quantities, respectively the no. units and the currency amount.

The consistent price / quantity information is already the remit of the `buySellPayout` (renaming TBD), which in turn would generate those 2 transfers. Each transfer can specify its `settlementOrigin`, i.e. the payout structure that generated it, so you can build lineage to that `buySellPayout` and the price information that's in it.

JBZ-Fragmos:
thank you @lolabeis for your feedback, 
however it does not match the need at stake : 

- the use-case that is need to have possibly two quantity **cannot be represented by "two transfers" as you suggest** : 
   - quantity of cash amount in 1 transfer describing the cash leg is **one thing**...
   - but the fact same quantity may also exist in second transfer that is describing the physical leg is **second/distinct thing** : for instance, think about a "free of payment" that is **stand-alone 1 transfer physical leg** : you may want to represent not only the quantity of physical units (**quantity 1**) but also the nominal value of this single transfer (**quantity 2**)

- the use-case that is the need to have a Price as well is needed inside the Transfer per se, such an information **cannot be** **indirect** via **settlementOrigin** ===> **Price shall belong to Transfer pe se**  that is to ensure **readability** and **consistency** between _quantity_ transfered and _price_ at which it is being transfered, at Transfer level ; 
   - indeed, **settlementOrigin** is only **high level refecence** to any whole Payout node, so the **"price" attribute** possibly embedded somewhere inside the referenced Payout will be **impossible to extract** for the purpose of being used to represent effectively the price of the transfer,
   - besides, there are a lot of various types of payout and reasons why a transfer would occur, so there is **no reason** why only **"BuySellPayout" would be the only one to reference** with settlementOrigin,

JBZ-Fragmos:
[like]  Jean-Baptiste Ziadé reacted to your message:
________________________________

PR #2914, opened 2024-05-07
Roger Guitart AmericanExercise condition update

regnosys-prod-user:

Issue #2913, opened 2024-05-07
Proposed modification of AmericanExercise condition under ExerciseTerms


rogerguitart:
## Background ##

It has been brought up internally that the conditions under `ExerciseTerms` in the refactored `OptionPayout` (in dev) are not sufficiently restrictive for the american options. More specifically, the `AmericanExercise` condition should not only control that `expirationDate` is present (which it already does through the `CommencementAndExpirationDate` condition), but also that **it is present only once** (not currently implemented in the model).

## Proposal ##

This proposal suggests to add the constraint of making the count of `expirationDate` equal to 1 whenever an `American` style option is present. The change would be as follows:

![image](https://github.com/finos/common-domain-model/assets/141412602/d5da1f2a-2ecc-484a-a2aa-b2b440da2168)

## Compatibility ##

There should not be any compatibility issues with the proposed change, as it is only adding a small modification to a condition.

PR #2920, opened 2024-05-10
Roger Guitart AmericanExercise Condition Update v2

regnosys-prod-user:


rogerguitart:
Linked to issue #2913

rogerguitart:
Hi @mgratacos @lolabeis , here is the new PR regarding the condition change that needed to be resubmitted. See below the updated Release Notes - sent message through Slack to get them approved

rogerguitart:
# _Product Model - Modification of AmericanExercise Condition in ExerciseTerms_
_Background_

The conditions under `ExerciseTerms` in the refactored `OptionPayout` are not sufficiently restrictive for the american options. More specifically, the `AmericanExercise` condition should not only control that the `expirationDate` is present (which it already does through the `CommencementAndExpirationDate` condition), but that it is also **present only once** (not currently implemented in the model). This release aims at modifying the `AmericanExercise` condition to correct this.

It should be noted that this proposal contains a backwards incompatible change, given that a condition has been made stricter, but should not impact any of the actual implementations.

_What is being released?_

- The `AmericanExercise` condition under the `ExerciseTerms` type has been modified to constraint the `expirationDate` field to only one occurrence.

_Review directions_

In the Rosetta Platform, select the Textual View and inspect the change identified above.

mgratacos (approved):
As agreed by the CDM Derivatives Products and Business Events Working Group on May 8th 2024.

lolabeis:
Please do not re-submit the condition change - it has already been merged - only submit the release notes. These notes should be provided as an update to the RELEASE.md file.

Please position the backward-incompatibility comment in a specific *Backward incompatible changes* section (same as *Review directions*). This is designed to aid collection of the release notes for the future 6.0 prod release.

The review direction section must contain a link to the PR (the original one with the condition change in this case).

rogerguitart:
@lolabeis I've updated the release notes above as well as the ones on the original PR (#2914), please double check.

PayalKhanna:
Superseded by https://github.com/finos/common-domain-model/pull/2921

PR #2878, opened 2024-04-19
Gerard Opazo Trigger type refactoring

regnosys-prod-user:

Issue #2849, opened 2024-04-09
Proposed refactoring of level and levelPercentage within Trigger CDM element to represent Barriers


gopazoTH:
Currently we have in the model the lower-barrier and upper-barrier elements represented in barrierFloor and barrierCap within the OptionPayout. 
The items **level** and **levelPercentage** would give us the value of that barrier if it is monetary or percentage respectively, in itself, these values should be represented as if they were of type **PriceSchedule** and not simple **number** type.
![image](https://github.com/finos/common-domain-model/assets/109520965/a8633c34-b72f-4b66-ab99-e61901aa1023)

The proposal is to unify level and levelPercentage in a single level element of type PriceSchedule, since this value itself should be considered as a Price according to Acc/Dec current work.
![image](https://github.com/finos/common-domain-model/assets/109520965/40eab7c1-3c18-4713-b5f8-4aee78e1fe00)
(Update from Peer Review 10/04/24: It was discussed that barrier extraction would be conditioned by the path **OptionPayout->feature->knock->knockIn/knockOut->trigger->level** instead of barrierCap/barrierFloor, this does not affect this issue as refractoring of the Trigger type is still involved)
It would be necessary to be discussed / agreed /solved within related PR :

- mapping impact of Price instead of “level” and “levelPercentage”
- enrichment of existing Enum value list of PriceExpressionEnum or PriceTypeEnum for barriers
- change, if needed, the required choice on Trigger type:
![image](https://github.com/finos/common-domain-model/assets/109520965/52f66525-cbeb-4b9c-9826-cfa2591e2af5)

gopazoTH:
(Update from DPBE 24/04/24): according to this meeting, it was agreed that the proposed cardinality of PriceSchedule (0..*) should be changed to (0..2), those changes were addressed in the corresponding PRs linked to this issue.
![image](https://github.com/finos/common-domain-model/assets/109520965/57b38652-898d-48ef-b869-1eaff7449648)

gopazoTH:
Closed by https://github.com/finos/common-domain-model/pull/2926

PR #2926, opened 2024-05-21
Gerard Opazo Trigger type refactoring Dev release

PayalKhanna:
original PR https://github.com/finos/common-domain-model/pull/2878


PayalKhanna:
Issue https://github.com/finos/common-domain-model/issues/2849

PayalKhanna:
aprroved https://github.com/finos/common-domain-model/pull/2878#issuecomment-2088618295

eacunaISDA (dismissed):

PayalKhanna:
@gopazoTH @dshoneisda can you review expectations changes please there are increase in validation failures

gopazoTH (approved):
dev release looks good, with no validation errors now, good to go

rosetta-source/src/main/rosetta/observable-event-type.rosetta
```diff
@@ -10,16 +10,15 @@ import cdm.observable.common.*
 
 type Trigger: <"Trigger point at which feature is effective.">
 
-    level number (0..1) <"The trigger level.">
-    levelPercentage number (0..1) <"The trigger level percentage.">
+    level PriceSchedule (0..2) <"The trigger level.">
```
hugohills-regnosys, line 13: Do we really want differently named parameters between prod and dev?

hugohills-regnosys: In prod, it is `levelPrice`, but in dev, it is `level`.  So anyone upgrading would need to change that.  If you make them the same, it would just work.

rosetta-source/src/main/rosetta/mapping-fpml-confirmation-tradestate-synonym.rosetta
```diff
@@ -1239,9 +1243,9 @@ synonym source FpML_5_Confirmation_To_TradeState extends FpML
 
     Trigger:
         + level
-            [value "level"]
-        + levelPercentage
-            [value "levelPercentage"]
+            [hint "level"]
+            [hint "levelPercentage"]
+            [value "levelPrice"]
```
hugohills-regnosys, line 1248: Is levelPrice actually part of the FpML schema?

gopazoTH: yes this is part of the FpML schema, I am just covering it to be safe in the future
image

hugohills-regnosys: OK great, thanks for checking

rosetta-source/src/main/rosetta/observable-event-type.rosetta
gopazoTH: yes, that what we discussed this morning to unify it to a single level elemt, but I see your point Hugo, we can put the levelPrice element and delete level/levelPercentage then.

gopazoTH: I just have a chat with Anna and Marc and we concluded that it would be better to use ` level PriceSchedule (0..2)` since we have also a RR in DRR that would be using that level element. We can delete CDM Prod release and keep this one, would be that ok for you Hugo?

hugohills-regnosys: Currently, DRR (Prod and Dev) both use CDM Prod, so if you only make the change in CDM Dev then it can't be picked up in DRR.

hugohills-regnosys: I discussed with @gopazoTH  offline, and understood that the intention is to not release on CDM Prod

hugohills-regnosys (approved):

PR #2963, opened 2024-06-04
Story/882/4606

ja6a-regnosys:
add docs

### DIFF
--- a/docs/cdm-guidelines.md
+++ b/docs/cdm-guidelines.md
@@ -23,7 +23,7 @@ version of CDM. Topics covered are listed below:
 # Setup
 
 In order to use the CDM in a Maven project, the following dependency
-needs to be addeed to the project pom.xml:
+needs to be added to the project pom.xml:
 
 ``` xml
 <dependency>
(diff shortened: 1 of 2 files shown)


docs/cdm-guidelines.md
```diff
@@ -43,6 +43,9 @@ needs to be addeed to the project pom.xml:
 -   The CDM uses [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) for each of the pojos. The
     distribution ships with the json to java object serialisers.
 
+_NOTE:_ For CDM releases prior to version 4.0.0 the dependencies can be found in the ISDA repository: `https://europe-west1-maven.pkg.dev/production-208613/isda-maven`
```
hugohills-regnosys, line 46: Just to be fully explicit, should we state that current releases are available in Maven Central.  Something like:
> All current CDM dependencies are available in Maven Central.  For CDM releases prior to version 4.0.0 the dependencies can be found in the ISDA repository:

hugohills-regnosys (approved):

PR #2970, opened 2024-06-06
[PROD] DSL update

SimonCockx:


eacunaISDA (approved):

chrisisla (changes requested):
This appears to make a change to the DSL that should be approved by the TAWG before the CDM is updated to use it. The TAWG is meeting next week, please present the change and its impact at that meeting.

chrisisla (approved):

PR #2976, opened 2024-06-10
[5.x.x] Update FpML mappings based on DRR and ISDA member mapping fixes

regnosys-prod-user:


hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/2982

PR #3018, opened 2024-06-28
Update ISOCurrencyCodeEnum based on updated scheme ISO Standard 4217

regnosys-prod-user:


llynhiavu (commented):
5 files are included with changes, the PR (about adding a new currency code) only applies to the first file base-staticdata-asset-common-enum.rosetta

the other 4 files only contain formatting syntax changes.  Is it the intention to bundle the extensive formatting changes with the ISO code addition?

dshoneisda (approved):

PR #3051, opened 2024-07-18
JEAN BAPTISTE ZIADE ObservationTermsValuationTerms_DRAFT

regnosys-prod-user:


JBZ-Fragmos:
@lolabeis @Oblongs

**CHANGE PROPOSAL**

proposed changes mainly consists in :
- minor refactoring/enrichment of ObservationTerms _and_ ValuationTerms (= ValuationDates renamed, because does not only contain dates)
- moving these two components in PayoutBase (with an optional cardinality)

will draft short GitHub Issue to be discussed with the group, but you can already have a look at it, ideally would target this changes to be released with all the change package discussed in context of the ARTF...

**RATIONALE**

- i beleive these changes are relevant for improving the model as it is, because it will make it more consistent i.e. there is no good reason why ObservationTerms and/or ValuationTerms are present for some Payout (e.g. performancePayout, OptionPayout) but not for the other ones... from business standpoint, all such "terms" are comparable to "settlementTerms" : they may be required or not depending business context and product wrappers, and cannot be restricted to particular payout more than another one... hence the proposal to have all of them in PayoutBase

- besides, they will anyway be required when i will present the "BespokePayoutFormula" so that would also help if some preliminary components would already be part of the model at that time (would permit to focus only on the set of components coming with root object "BespokePayoutFormula", and not to confuse with other components, which are required, but make sense per se anyway, regardless of "BespokePayoutFormula")

JBZ-Fragmos:
please see related Issue @dshoneisda - would like presenting at next meeting
https://github.com/finos/common-domain-model/issues/3058

JBZ-Fragmos:
overview at target (new items are highlighted in yellow)
image

focus on ObservationTerms : cf. 5 items highlighted below 
image

PR #3056, opened 2024-07-22
JEAN BAPTISTE ZIADE AddFixedPriceInForwardPayout

regnosys-prod-user:


JBZ-Fragmos:
this particular PR is obsolete because forwardPayout has been replaced by settlementPayout

yet the business need to have Price inside the Payout is still valid, meaning we still recommand to add such attribute inside settlementPayout

PR #3085, opened 2024-08-02
Lionel Smith-Gordon 3-ARTRelease.V2

regnosys-prod-user:


Oblongs:
Supersedes https://github.com/finos/common-domain-model/pull/3073

JBZ-Fragmos:
@Oblongs 
## UNDERLIERS ##

2 concerns/questions :

- **multiple distinct type** actually used behind same name “underlier” depending on the payout e.g. 

optionPayout : name optionUnderlier with type **OptionUnderlier** 
performancePayout : name underlier with type **Observable**
settlementPayout : name underlier with type **Underlier**

– is that what we all want, eventually ? it looks complexifying the model, eventually… why not use instead a unique generic type and custom restrictions as need be via conditions ?

- also some attributes seems to be in **overlap** i.e. see **Asset vs. TransferableProduct** in optionPayout and in settlementPayout

–  I believe this opens possibility to represent with two distinct paths the same thing, thus compromising convergence of booking models, right ? how do we handle this ?

For clarity, have highlighted these concerns below in graphical view :
- in blue the items at stake with concern 1. = using distinct type for same concept “underlier”)
- in red the ones at stake with concern 2. = overlaps 

image

JBZ-Fragmos:
@Oblongs 

## SettlementPayout ## 

2 suggestions

- **fixedPrice** shall be added as required attribute, of type PriceSchedule (1..*), with [metadata reference]

•	indeed, the essence of forward=settlement payout consists of straight commitment on fixedPrice (conceptually , that is equivalent to a strike, which obviously cannot be absent from core payout terms)
•	having unbound cardinality is to align with the same in reference source in tradeLot->priceQuantity->price(0…*)

- what about suggestion to **merge** it as well with **fixedPricePayout** ?

•	all in all, fixedPricePayout has the fixedPrice that settlementPayout is missing, 
•	whereas settlementPayout has the underlier that fixedPricePayout is missing
•	that is to say merging eventually consists in (1) removing fixedPricePayout and remap any item as need be with settlementPayout (2) provided settlementPayout has been completed with fixedPrice as suggested in point 1.
•	as an indication, Nick already agreed about this suggestion (in separate email loop), but would be great to get as well other participants advice about this suggestion

hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/3124

Issue #3087, opened 2024-08-08
Proposed CDM amendments for Equity Products


winitall01:
# *Background*
Following proposal is to make amendments for Equity Products 

## Issue 1: Zero Strike option ##
_Problem Statement:_ in PriceSchedule, Positive Price condition requires Cash Price to be greater than 0. This causes a CDM error for Zero Strike Option products:

_Proposal:_ Relax the Positive Price condition to allow to 0
![Screenshot 2024-08-08 at 16 17 22](https://github.com/user-attachments/assets/56cc446f-ffcc-4807-a7e2-90ccf31de569)

## Issue 2: Quantity Condition in Payout ##
_Problem Statement:_ in type Payout, Quantity condition expects a priceQuantity or interestRatePayout legs only. For FX Options, there’re no interestRatePayout legs, hence this condition throws an error for FX Options. 

_Proposal:_ Add `foreignExchange` check in condition `Quantity` 
![Screenshot 2024-08-08 at 16 18 49](https://github.com/user-attachments/assets/8cb36527-edee-4df3-ad37-60e80b548f49)

## Issue 3 cardinality of expirationTime ##
_Problem Statement:_ in type ExerciseTerms, expirationTime has a mandatory cardinality (1..1). For FX Options, there’s no fix expirationTime. 

_Proposal:_ relax `expirationTime` cardinality rule. Also, add a condition to require `expirationTimeType` when expirationTime exists or vice versa
![Screenshot 2024-08-08 at 16 18 49](https://github.com/user-attachments/assets/8cb36527-edee-4df3-ad37-60e80b548f49)

## Issue 4: Currently, there’re no qualification functions for Equity Forwards ##

_Proposal:_ Add following Qualification functions: 
`Qualify_EquityForward_PriceReturnBasicPerformance_SingleName`
`Qualify_EquityForward_PriceReturnBasicPerformance_SingleIndex`
`Qualify_EquityForward_PriceReturnBasicPerformance_Basket`
![Screenshot 2024-08-08 at 16 22 40](https://github.com/user-attachments/assets/ce51cd36-d049-4ff6-8653-ce11f8e80af0)

## Issue 5: Equity Option requires check for `fxFeature`. ##

_Problem Statement:_ Currently, FX Option does not qualify from the existing functions
_Proposal:_ Add `Quanto` and `Composite` check for Equity Option in the following functions: 
`Qualify_EquityOption_PriceReturnBasicPerformance_SingleName`
[...]

PR #3089, opened 2024-08-09
Vineet Kumar CDM%41_09082024

regnosys-prod-user:


winitall01:
on the back of : #3087

minesh-s-patel:
This PR has been closed and replaced with #3092

PR #3124, opened 2024-09-04
Lionel Smith-Gordon ART P3 with mapping

regnosys-prod-user:


hugohills-regnosys:
Replaced. by https://github.com/finos/common-domain-model/pull/3127

Issue #3119, opened 2024-09-03
CDM Technology Architecture WG - September 12th, 2024


eteridvalishvili:
# CDM Technology Architecture Working Group Minutes
Meeting Host: Chris Rayner, ISLA

## Date
September 12th, 2023 - 9 am EST / 2 pm BST

## Untracked attendees
- Fullname, Affiliation, (optional) GitHub username

## Meeting notices
- FINOS **Project leads** are responsible for observing the FINOS guidelines for [running project meetings](https://community.finos.org/docs/governance/meeting-procedures/). Project maintainers can find additional resources in the [FINOS Maintainers Cheatsheet](https://community.finos.org/docs/finos-maintainers-cheatsheet).

- **All participants** in FINOS project meetings are subject to the [LF Antitrust Policy](https://www.linuxfoundation.org/antitrust-policy/), the [FINOS Community Code of Conduct](https://community.finos.org/docs/governance/code-of-conduct) and all other [FINOS policies](https://community.finos.org/docs/governance/#policies). 

- FINOS meetings involve participation by industry competitors, and it is the intention of FINOS and the Linux Foundation to conduct all of its activities in accordance with applicable antitrust and competition laws. It is therefore extremely important that attendees adhere to meeting agendas, and be aware of, and not participate in, any activities that are prohibited under applicable US state, federal or foreign antitrust and competition laws. Please contact legal@finos.org with any questions.

- FINOS project meetings may be recorded for use solely by the FINOS team for administration purposes. In very limited instances, and with explicit approval, recordings may be made more widely available.

## Agenda
- [x] Convene & roll call (5mins)
- [x] Display [FINOS Antitrust Policy summary slide](https://community.finos.org/Compliance-Slides/Antitrust-Compliance-Slide.pdf) 
- [x] Review Meeting Notices (see above)
- [x] Review minutes of last meeting
   - [x] https://github.com/finos/common-domain-model/issues/3025
- [x] Updates (15 mins) 
  - [x] Rune issues #3046 - @chrisisla 
  - [x] Code Generators #2928 - @chrisisla 
  - [x] Serialisation - @chrisisla
  - [x] Ongoing Priorities Survey #3106 - @chrisisla
  - [x] Rune Translate2 - @davidalk
- [x] Rune language enhancements (10 mins) - @davidalk
- [x] TAWG Priority setting POSTPONED DUE TO TIME - @chrisisla
- [x] AOB, Q&A & Adjourn (5mins)

[...]

PR #3125, opened 2024-09-05
[6-dev] Update CDM to use DSL version 9.17.1 and update reference data ISO currency scheme

hugohills-regnosys:


hugohills-regnosys:
Changes approved at CDM Technology Architecture WG - September 12th, 2024
https://github.com/finos/common-domain-model/issues/3119

minesh-s-patel (approved):

PR #3181, opened 2024-10-16
[5xx] Added function RoundToPrecisionRemoveTrailingZero

PayalKhanna:


rosetta-source/src/main/java/cdm/base/math/functions/RoundToPrecisionRemoveTrailingZeroImpl.java
```diff
@@ -0,0 +1,31 @@
+package cdm.base.math.functions;
+
+import cdm.base.math.RoundingDirectionEnum;
+
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+public class RoundToPrecisionRemoveTrailingZeroImpl extends RoundToPrecisionRemoveTrailingZero {
```
hugohills-regnosys, line 8: Missing the Guice configuration.  Also, please can we rename to `RoundToPrecisionRemoveTrailingZeros` (as it may remove multiple trailing zeros)

dshoneisda (approved):

PR #3205, opened 2024-10-29
Paul Hands Added initial skeleton for ISDA master agreement

regnosys-prod-user:


chrisisla (changes requested):
I don't think it would be in keeping with the common nature of the model to have a new ISDA master agreement namespace. I'll add more details to the Issue to explain why.

tomhealey-icma (commented):
I believe this should be in a separate namespace. When you have a domain specific model, in this case the ISDA agreement, and not a general taxonomy for legal agreements, it should exist in a separate space.  That said, it's not clear how this will be used, the CDM does not have a capability to parse, analyze or manipulate text. Maybe some use cases would help. I would also suggest that this should be optional and not part of the core build.

dshoneisda:
@lolabeis @mgratacos @llynhiavu would appreciate your input here too

My view is that this should probably be brought to the CDM Legal WG as a starter, but that it makes sense for this to be contributed after the decomm of Foundations and legal agreement structures brought back into the model.
I think it makes sense to have common attributes where it can be done, but in reality the legal frameworks are by nature different and thus having separate types of document under the general legal agreement structure is inevitable

regnosys-prod-user:
_What is being released?_

Moved the elections into Master as recommended, extended from commonElections.

There are still some namespaces that need deleting which I'm going to initially attempt to do using github dev

> [!NOTE]

lolabeis:
@paulhands 
Please empty all the files that are no longer needed and add a comment:
```
// PLEASE DELETE THIS FILE
```
We will take care of the deletion on our side.

E.g. legaldocumentation > gmraagreement should be deleted but still has content.

PR #3265, opened 2024-11-18
[5xx] Adding "MAS" value in TaxonomySourceEnum

PayalKhanna:


manel-martos (approved):
As per DRR requirement and it is just an addition to the TaxonomySourceEnum

manel-martos (approved):

PR #3250, opened 2024-11-12
Add price reference mapping for SettlementPayout.priceSchedule

hugohills-regnosys:

PR #3274, opened 2024-11-20
Add price reference mapping for SettlementPayout.priceSchedule

hugohills-regnosys:


hugohills-regnosys:
See PR https://github.com/finos/common-domain-model/pull/3250 instead

PR #3294, opened 2024-12-10
Simon Cockx Fix cardinality errors

regnosys-prod-user:


tomhealey-icma (changes requested):
The function, SecurityFinanceCashSettlementAmount does not work for repo for a couple of reasons:
1) The filter, tradeState -> trade -> tradeLot -> priceQuantity -> quantity, FinancialUnitEnum -> Share only works for Shares on not other units. Also, FinancialUnitEnum is optional so it may not exist.
2) The function suggest that the repo repurchase price is the collateral value which is also incorrect. The repurchase price is the cashleg principal + interest. 

If this function is intended to only work for Security Lending then please change the function name and remove the repurchase price comments.

One other point, the condition, condition ProductIdentifiersMatch:

        tradeState -> trade -> tradableProduct -> tradeLot -> priceQuantity -> observable -> productIdentifier = assetPayout -> securityInformation -> security -> productIdentifier

is overly restrictive and fails in many cases where the identifiers actually match. I would change this to:

        tradeState -> trade -> tradableProduct -> tradeLot -> priceQuantity -> observable -> productIdentifier -> identifier = assetPayout -> securityInformation -> security -> productIdentifier -> identifier

SimonCockx:
Hi @tomhealey-icma. Note that I have not changed the behaviour of `SecurityFinanceCashSettlementAmount` - I have just fixed a cardinality warning that the DSL highlighted. See release notes:

> The function `SecurityFinanceCashSettlementAmount` contained a multiplication of which one operand - `securityQuantity` -
   was of multi cardinality. In practice, due to filtering, it should always be a single element, so this was fixed with `only-element`.

On your second comment:

> One other point, the condition, condition ProductIdentifiersMatch:

I have not touched this condition at all, so I don't see how it is relevant.

I have no intention of changing the model in any semantic way. If you have found an existing issue with `SecurityFinanceCashSettlementAmount`, it will need to be looked at in a separate contribution.

If you agree with my remarks above, could you please unblock this PR? If not - feel free to reach out.

hugohills-regnosys (approved):

Oblongs:
@tomhealey-icma 

Tom, you make a couple of points here, namely that
> The function, SecurityFinanceCashSettlementAmount does not work for repo for a couple of reasons

and

> the condition, condition ProductIdentifiersMatch 
> is overly restrictive and fails in many cases where the identifiers actually match. 

These may be valid points however we believe they are beyond the scope of this particular PR which is focused on addressing a specific issue in the DSL on the evaluation of cardinality.  In both this PR #3294 and the related one PR #3295, the only change in the function in question is to address this issue with the inclusion of `only-element` in the evaluation of the `FilterQuantityByFinancialUnit` function.  Your points relate to the business modelling of these functions for some products; you have identified potential issues in the model which have been there for some time.

We would like to proceed with these PRs and fix the underlying DSL issues which occur across the model, as quickly as possible.  Could we propose that we:
- unblock and proceed to release these PRs
- raise issue(s) related to what you have identified
- prioritise these through the appropriate WGs to get the issues addressed.

Thanks
Lionel, acting as Release Manager for the CDM.

tomhealey-icma (dismissed):

eacunaISDA (approved):

PR #3295, opened 2024-12-11
Simon Cockx Fix cardinality errors

regnosys-prod-user:


hugohills-regnosys (approved):

tomhealey-icma (changes requested):
The function, SecurityFinanceCashSettlementAmount does not work for repo for a couple of reasons:

The filter, tradeState -> trade -> tradeLot -> priceQuantity -> quantity, FinancialUnitEnum -> Share only works for Shares on not other units. Also, FinancialUnitEnum is optional so it may not exist.
The function suggest that the repo repurchase price is the collateral value which is also incorrect. The repurchase price is the cashleg principal + interest.
If this function is intended to only work for Security Lending then please change the function name and remove the repurchase price comments.

One other point, the condition, condition ProductIdentifiersMatch:

    tradeState -> trade -> tradableProduct -> tradeLot -> priceQuantity -> observable -> productIdentifier = assetPayout -> securityInformation -> security -> productIdentifier
	
	tradeState -> trade -> tradableProduct -> tradeLot -> priceQuantity -> observable -> productIdentifier -> identifier = assetPayout -> securityInformation -> security -> productIdentifier -> identifier

Oblongs:
@tomhealey-icma 

Tom, you make a couple of points here, namely that
> The function, SecurityFinanceCashSettlementAmount does not work for repo for a couple of reasons

and

> the condition, condition ProductIdentifiersMatch 
> is overly restrictive and fails in many cases where the identifiers actually match. 

These may be valid points however we believe they are beyond the scope of this particular PR which is focused on addressing a specific issue in the DSL on the evaluation of cardinality.  In both this PR #3295 and the related one PR #3294, the only change in the function in question is to address this issue with the inclusion of `only-element` in the evaluation of the `FilterQuantityByFinancialUnit` function.  Your points relate to the business modelling of these functions for some products; you have identified potential issues in the model which have been there for some time.

We would like to proceed with these PRs and fix the underlying DSL issues which occur across the model, as quickly as possible.  Could we propose that we:
- unblock and proceed to release these PRs
- raise issue(s) related to what you have identified
- prioritise these through the appropriate WGs to get the issues addressed.

Thanks
Lionel, acting as Release Manager for the CDM.

tomhealey-icma (dismissed):

dshoneisda (approved):

PR #3352, opened 2025-01-29
Oscar Twomey CDM user can access an extended agreement model

regnosys-prod-user:

PR #3372, opened 2025-02-05
Oscar Twomey CDM user can access an extended agreement model - Release

JayasriR:
Original PR: https://github.com/finos/common-domain-model/pull/3352


rosetta-source/src/main/rosetta/legaldocumentation-master-additionalterms-type.rosetta
```diff
@@ -0,0 +1,12 @@
+namespace cdm.legaldocumentation.master.additionalterms
```
chrisisla, line 1: Why do we need this new namespace? The types being moved are all empty, can they actually be used at all anywhere? Should they not be removed rather than having them in a new namespace? Can you elaborate on the purpose of the new namespace please?

rosetta-source/src/main/rosetta/legaldocumentation-contract-type.rosetta
```diff
@@ -13,3 +13,14 @@ type Agreement: <"Specification of the standard set of terms that define a legal
     transactionAdditionalTerms TransactionAdditionalTerms (0..1) <"Any additional terms which mainly intend to specify the extraordinary events that may affect a trade and the related contractual rights and obligation of the parties when this happens">
     condition:
         one-of
+
+type BrokerConfirmation: <"Identifies the market sector in which the trade has been arranged.">
+    brokerConfirmationType BrokerConfirmationTypeEnum (1..1) <"The type of broker confirmation executed between the parties.">
+        [metadata scheme]
+
+type IssuerTradeId: <"A class for a two-parts identifier, such as a USI.">
```
chrisisla, line 21: Not sure this should be in here? Is this associated to the legal documentation?

chrisisla (changes requested):

rosetta-source/src/main/rosetta/legaldocumentation-master-additionalterms-type.rosetta
lolabeis: The types being moved here are indeed empty in CDM, but not in ISDA Foundations. They are present as attributes in the CDM's `ExtraordinaryEvents` and `TransactionAdditionalTerms` types and can't be removed.

The reason for moving them to a dedicated namespace that only contains these empty types is to prevent the rest of the `legaldocumentation.master` namespace in ISDA Foundations to go out of sync with CDM - which is what actually happened. This way, we only need to over-ride `legaldocumentation.master.additionalterms` in ISDA Foundations. This is similar to what is already done for `legal documentation.csa`, that only contains empty types.

At a later point when we complete the migration, the types in this new namespace will be filled with content from ISDA Foundations - but for now we just keep them empty, as they currently are in CDM.

rosetta-source/src/main/rosetta/legaldocumentation-contract-type.rosetta
lolabeis: Actually this type doesn't look like it's used anywhere, even in ISDA Foundations, so we're simply going to remove it.

rosetta-source/src/main/rosetta/legaldocumentation-master-additionalterms-type.rosetta
chrisisla: Thanks for this explanation Leo :) As we discussed, these attributes may not actually fit under the `master` namespace and so more analysis is required. I will approve these changes now and as agreed Oscar will create a new Issue which we can take to the CRWG for further input too.

rosetta-source/src/main/rosetta/legaldocumentation-contract-type.rosetta
chrisisla: Thanks Leo!

chrisisla (approved):
Approving this Issue as agreed. A new Issue will be created to reconsider whether the empty types really should be under `master` or potentially a new namespace under `legaldocumentation`.

Oscar-Twomey:
> Approving this Issue as agreed. A new Issue will be created to reconsider whether the empty types really should be under `master` or potentially a new namespace under `legaldocumentation`.

I have created a new issue here: #3408

PR #3375, opened 2025-02-05
Fix rune data annotation 2

davidalk:


pom.xml
```diff
@@ -228,6 +228,11 @@
                             <artifactId>com.regnosys.rosetta.lib</artifactId>
                             <version>${rosetta.dsl.version}</version>
                         </dependency>
+                        <dependency>
```
minesh-s-patel, line 231: This is a circular dependency. We can't depend on the `cdm-java` jar as this (`rosetta-maven-plugin`) is used to generate the `cdm-java` in the first place.

PR #3378, opened 2025-02-05
Oscar Twomey Moved ClosedStateEnum, ClosedState to event.common

regnosys-prod-user:


rosetta-source/src/main/rosetta/event-common-type.rosetta
```diff
@@ -683,3 +683,10 @@ type MarginCallResponse extends MarginCallBase: <"Represents common attributes r
     marginCallResponseAction MarginCallResponseAction (1..*) <"Specifies the margin call action details, including collateral to be moved and direction.">
     marginResponseType MarginCallResponseTypeEnum (1..1) <"Indicates the response type, such as, is the margin call response a 'full' 'part' agreement or 'dispute'.">
     agreedAmountBaseCurrency Money (1..1) <"Indicates the amount that posting entity agrees to remit in response to margin call (in base currency).">
+
+type ClosedState: <" A class to qualify the closed state of an execution or a contract through the combination or a state (e.g. terminated, novated) and a set of dates: activity date, effective date and, when relevant, last payment date.">
+
+    state ClosedStateEnum (1..1) <"The qualification of what gave way to the contract or execution closure, e.g. allocation, termination, ...">
```
chrisisla, line 689: Just a comment (as this PR is moving existing attributes rather than updating them) but is `state` the best name for this attribute? This was the action, event or the reason that resulted in the item being closed. Having `closedState->state` seems a bit unintuitive, as surely the state of something under ClosedState would be closed? 🤷

lolabeis: @chrisisla I do concur that this merits revisiting. This should get on the roadmap this year as part of the "Position" work. But it will be a backward-incompatible change in the context of CDM 7 or above and is beyond the scope of the current issue which is just looking to tidy-up the `legalagreement` namespace.

chrisisla: Agreed 👍  As you say this is not the purpose of this PR and that's why I just put this as a comment :)

chrisisla (approved):

JayasriR:
Superseded by https://github.com/finos/common-domain-model/pull/3386

PR #3379, opened 2025-02-06
DSL dependency update and set FpML / ISO sync configuration to AdditiveMatch

JayasriR:

PR #3381, opened 2025-02-07
[6-dev] Use actions/upload-artifact@v4

hugohills-regnosys:


hugohills-regnosys:
Released as part of https://github.com/finos/common-domain-model/pull/3379

PR #3359, opened 2025-01-30
Oscar Twomey STORY-3632_ExtendedFROModel

regnosys-prod-user:

PR #3392, opened 2025-02-11
Oscar story 3632 STORY-3632_ExtendedFRO  missed changes

PayalKhanna:
original PR https://github.com/finos/common-domain-model/pull/3359


lolabeis (commented):
This PR should also adjust the RELEASE.md file to point to itself, in addition to the original PR, in the review directions. The content of the release note also needs to be adjusted slightly to reflect changes.

chrisisla:
Hi @Oscar-Twomey ,
If you are happy to rename the function (yet again, sorry!) and update the comment to reflect the fact this could return multiple indices then I think this is looking good. Thanks again for making the changes!

Oscar-Twomey:
> Hi @Oscar-Twomey , If you are happy to rename the function (yet again, sorry!) and update the comment to reflect the fact this could return multiple indices then I think this is looking good. Thanks again for making the changes!

Thanks Chris, this has been updated.

chrisisla (approved):
Function looks good! 👍 
I think we should review whether we need a new `Administrator` type at all though, as I think this is essentially a `Party`. Let's discuss on follow up PRs please 😃

PR #3563, opened 2025-03-24
[7-dev] Fix cve-scanning and licence-scanning and docusaurus build issues

hugohills-regnosys:


hugohills-regnosys:
The CVE and Licence scanning workflows were failing with the error:

Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run

Fixed as per recommendation on issue https://github.com/actions/upload-artifact/issues/480

hugohills-regnosys:
Docusaurus build was failing due to error:

Error message "error:0308010C:digital envelope routines::unsupported"

Fixed as per recommendation https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported

Issue #3206, opened 2024-10-29
Adding ISDA master agreement specific Agreement typings


paulhands:
## Background ##

While the CDM is leading the path in making a universal language to discuss legal agreements, a huge hurdle to get over is that the legal agreements themselves are written inherently differently. While there is overlap, there is NOT enough overlap to allow for an ISDA master agreement to be represented easily and readily into the system. At least, not without a lot of string values at the end.

 This is something the collateral working groups are moving towards with the ISDA foundations model and I believe a similar, heavily typed approach would work for the CDM for an ISDA master agreement

## Proposal ##

Separate out and add a new section to the Agreement type to include an ISDA master agreement specific type

Populate that with the basic skeleton and enhance further in a piecemeal, agile fashion.

## Compatibility ##

No compatibility issues, as this is adding new types and mappings only

chrisisla:
Hi @paulhands 

Although I agree a typed approach is better for representing legal documentation - as we have discussed in the past - I think it moves away from the "common" ideals that the CDM is attempting to drive. Adding an ISDA specific namespace under legalAgreement would indeed allow for ISDA specific terms to be put into the model, but then the model would also need to have namespaces for ISLA, ICMA, ESMA and indeed any other organisations' agreements that wanted to be represented in the CDM.

ISLA, in conjunction with D2LT, considered this at great length last year, and spent considerable time and resource coming up with a structure that would allow for master agreements to be represented in the CDM without having to have specific terms mapped out within it. 

In the discussions you and I have had ab out this subject we thought about removing some of the enumerated lists that were added and making the identifiers user definable i.e. strings. This should allow ISDA (and other) agreements to be mapped without adding any new namespaces that are specific to any one organisation. I'm happy to continue those conversations whenever you're available.

Thanks!

paulhands:
Hi @chrisisla ,

Happy to discuss this elsewhere, as I know a github issue isn't exactly the ideal place to try and get points across. But my overriding suspicion with this is that, because of the inherent differences in the contracts underpinning what the CDM is trying to represent, _not_ having a dedicated optional namespace in the `agreement` part of the CDM will quickly limit the amount of representation legal documents actually get, which will be reflected on uptake by firms wanting to use this aspect of the model.

As I say, happy to chat about it offline.

paulhands:
https://github.com/finos/common-domain-model/pull/3581

[...]

PR #3581, opened 2025-03-27
Added skeleton framework for master agreements

paulhands:


paulhands:
Due to the limitations of Rosetta and not being able to delete namespaces inside it, I have forked, and then reraised this PR. Not sure if that was the right approach or not.

The previous PR that this is related to is https://github.com/finos/common-domain-model/pull/3205

This also references to https://github.com/finos/common-domain-model/issues/3206

lolabeis:
> Due to the limitations of Rosetta and not being able to delete namespaces inside it, I have forked, and then reraised this PR. Not sure if that was the right approach or not.

Thanks @paulhands . If you don't mind, pls keep the original PR (it means you can keep working in your original workspace and commit from it). I can see the empty files there, we can handle the deletion on our side.

I will close the present PR.

PR #3635, opened 2025-04-09
Georgina Tarres Common Function QuantityDecreasedToZero - QuantityDecreasedToZero function Fix

regnosys-prod-user:


regnosys-prod-user:
# *Common Function QuantityDecreasedToZero - QuantityDecreasedToZero function Fix*

_Background_

The function QuantityDecreasedToZero incorrectly flags as a decrease cases where the quantity goes from zero to zero.

_What is being released?_

# *CDM Event Model - Decrease to zero*

_Background_

The CDM function `QuantityDecreasedToZero` currently assumes that a decrease has occurred when the quantity is reduced to zero by checking that:

1. The quantity in the `after` `TradeState` is zero.
2. The quantity in the `before` `TradeState` is zero or greater.

This leads to incorrectly flagging as decrease cases where both the `before` and `after` quantities are zero, which should not be considered a decrease.

_What is being released?_

This contribution modifies the logic in the `QuantityDecreasedToZero` function so that a decrease is only considered when the before quantity is greater than zero, not equal to zero. 

_Review directions_

In Rosetta, select the Textual View and search for the changes mentioned.

_Review directions_

Changes can be reviewed in PR: [#3635](https://github.com/finos/common-domain-model/pull/3635)

> [!NOTE]

PayalKhanna:
Superseded by https://github.com/finos/common-domain-model/pull/3757

PR #3710, opened 2025-05-16
Imanol Jurado CDM Model - Addition of Product Grade Enumeration

regnosys-prod-user:


manel-martos:
@LionelSG-REGnosys - similar to the 5.x version, it appears that the ingestion expectations under `tests` module are causing the build failure. Please instruct the team to generate an update expectations PR for verification.

Issue #3004, opened 2024-06-21
Update CDM Build Process to use Java 22 


minesh-s-patel:
## Background ##

The Common Domain Model currently uses Java 11 for compilation. The Java 22 compiler is faster and more reliable than java 11 and is the latest supported Java version.

Note  - The generated Java code is compiled to Java 8 for community support. This will remain unchanged.

## Proposal ##

Update the GitHub Actions Build Process to use Java 22.
Update the parent pom.xml to use Java 22.
Keep the 'cdm-java' jar file produced to Java 8. (i.e. no impact to users)

## Compatibility ##

No impact to users of the CDM Jar files.
To build the CDM - Java 22 will be required.

tomhealey-icma:
Question was raised at the TAWG, is v22 an LTS and should CDM non-LTS version. From what I can determine v2 is not LTS.
@minesh-s-patel - please comment.

PR #3734, opened 2025-05-27
[master] Upgrade DSL to 9.54.0

SimonCockx:


SimonCockx:
Addresses https://github.com/finos/common-domain-model/issues/3004

PR #3806, opened 2025-06-16
Lucía Cortés Model update to support Asian Options in CDM 7 - Model update to support Asian Options in CDM 7

regnosys-prod-user:

PR #3827, opened 2025-06-24
Lucía Cortés Model update to support Asian Options in CDM 7 - Model update to support Asian Options in CDM 7 - Release

JayasriR:
Original PR: https://github.com/finos/common-domain-model/pull/3806


lolabeis (approved):
Re-approving following release notes fix

PR #3903, opened 2025-07-15
CDM Documentation - Update GOVERNANCE.md

igallizugaro:
-  ~Consider blueprint, resolve confusion (see docs disambiguation task)~
- [x] Link fixes
- [x] Removed sections broken out into separate pages and added links.

Aim is to retire governance.md and keep GOVERNANCE.md only or make clear distinction between policy and the CDM-specific organizational governance.

See also Issue https://github.com/finos/common-domain-model/issues/3933 for outstanding tasks around governance materials: e.g.
- Consider blueprint, resolve confusion (see docs disambiguation task)


dshoneisda (commented):
These changes look good when taken into context with the other PRs which provide the modularised pages on which all of this deleted information is being moved to, in order that there is one source for say approval guidelines or one source of maintainers list

eacunaISDA (approved):

PR #3970, opened 2025-08-11
FpML Confirmation - model-to-model Ingest

hugohills-regnosys:


hugohills-regnosys:
Replaced by https://github.com/finos/common-domain-model/pull/3996

PR #3971, opened 2025-08-13
JEAN BAPTISTE ZIADE AssetFlowBase - see GitHub Issue 3541

### LINKED ISSUE #3541: FRAGMOS -
## Related PR ##
ex-https://github.com/finos/common-domain-model/pull/3234

## Problem Statement ##

We want to enrich the representation of transfer i.e.

- represent not only the **quantity being transferred** that is what already exists in current model (**Quantity 1**), 
- but also, optionnaly, additional information :
   - the **notional quantity** at stake with it (**Quantity 2**)
   - as well as **price** related to the transfer (**Price**)
  
**BUSINESS CASE** : example : say transferred quantity is 100 Share (**Quantity 1**), and you want to have additional info such as the price, say 2 USD/Share (**Price**), with nominal quantity for it, thus being equal to 200 USD (**Quantity 2**)

**GAP ANALYSIS** : 
- as of today, only 1 quantity exists : current model ony permit to have **Quantity 1** 
- whereas business example above requires **Quantity 2 + Price attribute**

**ADDITIONAL BUSINESS CASES** : above example is the "easy" case : more kinds of quantity or price may be involved... for instance, let's keep same example, but say now there is **2nd Price involved** because of FX Features at stake e.g. settlement ccy <> fixing currency, so you may need to represent not only USD/Share Price, but possibly EUR/Share Price, and may also EUR/USD Price, etc.

**GAP ANALYSIS** : 
- as of today, only 1 quantity exists : current model ony permit to have **Quantity 1** 
- whereas business example above requires **Quantity 2 + Price attribute 1 + Price attribute 2**

## Proposal  = Additional Quantity + Price ##

Image

For clarity, because "transfer _extends_ TransferBase" that is where the changes are actually made i.e. in **TransferBase** :
Image

Besides, we propose modifications on the conditions : 
Image

## Compatibility ##

backward-incompatible = renaming prior "quantity" into "deliverableQuantity"

regnosys-prod-user:


regnosys-prod-user:
# *AssetFlowBase - see GitHub Issue 3541*

_Background_

Purpose of changes is to enrich current Transfer description by adding additional Quantity and Price components in AssetFlowBase. 
Main use case as an example among others : for representing the physical leg of a cash vs. physical transfer of a given Stock, you do not only need to represent number of shares being transferred (= deliverableQuantity), you also want to represent the corresponding nominal amount (= notionalQuantity), as well as the price at which transfer shall occurr (= price), and you would specify what Stock is at stake (= asset).

_What is being released?_

- prior quantity is renamed deliverableQuantity
- added : notionalQuantity
- added : price
- prior set of conditions is modified

_Review Directions_

Changes can be reviewed in PR: [#3971](https://github.com/finos/common-domain-model/pull/3971)

> [!NOTE]

JBZ-Fragmos:
hi @lolabeis

this PR relates to Issue https://github.com/finos/common-domain-model/issues/3541

lolabeis:
Hi @JBZ-Fragmos
Thanks for your PR, this has been pushed to our Sprint cycle and I'll be looking at it within the next week.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
```diff
@@ -19,17 +19,27 @@ type ComputedAmount: <"A class to specify the outcome of a computed amount, for
...
-        if asset -> Cash exists
-        then quantity -> unit -> currency exists
-        else if asset -> Commodity exists
-        then quantity -> unit -> capacityUnit exists
-        else if asset -> Instrument exists
-        then quantity -> unit -> financialUnit exists
+    
+    deliverableQuantity NonNegativeQuantity (1..1) <"The quantity that is transferred.">
+    notionalQuantity NonNegativeQuantity (0..*) <"The notional value or nominal amount of the asset to which the transfer is relative to.">
+    price Price (0..*) <"The price of the asset to which the transfer is relative to.">
```
lolabeis: notionalQuantity and price are both multi-cardinality. But as I understand it, there is a 1-to-1 mapping between each price (expressed in a certain currency) and notionalQuantity (in that same currency).

Consider creating a specific type insted that contains a single price and quantity, both with matching currency units, and include that attribute as optional with multiple cardinality.

Naming (TBD): e.g. `MonetaryQuantity`?

Such type would be useful in the existing `PriceQuantity` object, which can also have multiple quantities. In fact in this object, there is only one true quantity, the other ones are to accommodate alternative monetary expressions.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
```diff
@@ -19,17 +19,27 @@ type ComputedAmount: <"A class to specify the outcome of a computed amount, for
...
-    condition QuantityUnitExists: <"Aligns the type of unit and the type of asset.">
-        if asset -> Cash exists
-        then quantity -> unit -> currency exists
-        else if asset -> Commodity exists
-        then quantity -> unit -> capacityUnit exists
-        else if asset -> Instrument exists
-        then quantity -> unit -> financialUnit exists
+    
+    deliverableQuantity NonNegativeQuantity (1..1) <"The quantity that is transferred.">
+    notionalQuantity NonNegativeQuantity (0..*) <"The notional value or nominal amount of the asset to which the transfer is relative to.">
```
lolabeis: Naming-wise, consider:
* Keeping the existing "quantity" attribute name unchanged. Generally, renaming should clear a high hurdle. In this particular case, invoking a new "deliverable" terminology may sits oddly vs "Transfer".
* Using "monetary" instead of "notional". Usage of the latter has been scrubbed away from the model.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
```diff
@@ -19,17 +19,27 @@ type ComputedAmount: <"A class to specify the outcome of a computed amount, for
...
+
+  condition CashAsset: <"It is not permitted to deliver non-cash quantity if the asset subject to tranfer is Cash (however, it is possible to deliver cash quantity in relation to another asset type e.g. for instance when representing a cash transfer in relation to given share).">
+  if asset -> Cash exists
+    then deliverableQuantity -> unit -> currency exists
+
+  condition DeliverableQuantityUnit: <"The unit of the deliverable quantity must be consistent with the kind of asset the transfer is relative to (which does not apply to cash transfer because cash quantity may be delivered in relation to other assets then cash).">
+  if deliverableQuantity -> unit -> capacityUnit exists
+    then asset -> Commodity exists
+  else if deliverableQuantity -> unit -> financialUnit exists
+    then asset -> Instrument exists
```
lolabeis, line 44: The original condition was written the other way round:
```
if asset -> Commodity exists
then quantity -> unit -> capacityUnit exists
```
What's the reason for inverting the logic?

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
```diff
@@ -19,17 +19,27 @@ type ComputedAmount: <"A class to specify the outcome of a computed amount, for
...
+    
+    deliverableQuantity NonNegativeQuantity (1..1) <"The quantity that is transferred.">
+    notionalQuantity NonNegativeQuantity (0..*) <"The notional value or nominal amount of the asset to which the transfer is relative to.">
+    price Price (0..*) <"The price of the asset to which the transfer is relative to.">
+    asset Asset (1..1) <"The asset to which the transfer is relative to.">
+    settlementDate AdjustableOrAdjustedOrRelativeDate (1..1) <"The date on which the transfer to due.">
+
+  condition NotionalQuantityUnitIsMoney:
+  if notionalQuantity exists 
+    then notionalQuantity -> unit -> currency exists
```
lolabeis, line 34: When monetary quantities are provided, a condition should check that monetary quantity = quantity x price (same as `PriceQuantityTriangulation` on trade lots).

lolabeis (commented):
Thank you @JBZ-Fragmos for your contribution. Detailed review comments in line.

The build failure also needs to be addressed. Your workspace appears invalid as a result of the "quantity" renaming (because some static code relies on this attribute name), although there may be additional issues.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: understand why you could interprete in that way at first glance, but no, that is _not_ a correct assumption that **_"there is 1-to-1 mapping between each price (expressed in a certain currency) and notionalQuantity (in that same currency)"_**  - actually, there are many use cases cross-products / cross-assets, so that will depends which particular one is at stake

let me provide just 1 basic example where assumption is not correct : say product is a MtM CCS with EUR-fixed leg and USD-varying leg, and I want to represent the cash adjustment principal payment, that relates to the notional reset event, due to varying leg feature.... then you could have : 
- deliverableQuantity: 110 unit=USD), 
- notionalQuantity : 100 unit=EUR), 
- price (1.10 unit=USD perUnit=EUR, priceType=ExchangeRate)

about end-comment to consider _**"creation of specific type"**_ inside PriceQuantity - would not retain it :
- that looks to me clearly _beyond PR scope_ because that would touch a very core item in model (PriceQuantity) so I'm not at all confortable to push this as part of current "local" change... that is raising lot of questions... not the place here
- plus to that, have alredy added a condition _NotionalQuantityUnitIsMoney_ that meet the point, since it enforces "monetary" state when populating notionalQuantity, and I beleive that using local condition with an existing type whenever feasible instead of creating new type is aligned with sound practice, per CDM design principle

yet I appreciate your upfront comment _**"notionalQuantity and price are both multi-cardinality"**_ ; indeed, unbound cardinality might result in lack of clarity if having to link pairs of multi-price with multi-quantity, etc, (hence why you are proposing in that regards to have multiple of a new type where single price and single quantity are necessarily associated)
accordingly, i simply suggest to restrict cardinality to [1] instead of [*] - which have now changed in last PR version as below :

image
****

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: agree in principal about logic you describe when renaming or not, etc. but then do not agree about the current case, because here, there is precisely good reason for adding prefix such as 'delivarable" or "notional" to the native name "quantity" coming from original type Quantity - that is precisely to provide the modelling with a name that clarfies and ensures business readability about the local use case of the generic type behind it

let me elaborate :
- **deliverable**Quantity is clarifying and avoinding any doubt that, though another quantity might be populated, this one is the one being delivered (meaning that the other one, by comparison, is not being delivered, so if present, must be for informational reasons, etc.)
- **notional**Quantity is the far closer aligned with market practice and also with multiple ISDA Definitions cross-asset wise, etc. than monetaryQuantity that is really more "model" or if i may say "IT/tech" prefix...

so would not follow your suggestion here, with the hope explanations have fairly clarified why

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: because the deliverableQuantity unit is **not** necessarilyy driving the unit of the Asset

let's have 1 example : 
say there is PriceReturn swap or Forward or Call 0 etc. which underlier is Commodity, say GB Bushel, and such product is Cash-settled... then at payment date you may want to deliver the cash that is related to this underlier, meaning you would have : 
- deliverableQuantity: 1000000 unit=USD
- price : unit=USD perUnit->**capacityUnit**-> GBBSH, priceType=AssetPrice
- asset: Commodity -> etc.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: i understand that is "separate comment" for the record, I mean you are refering to a _**func**_ component as far as I can understand, and for the avoidance of doubt, there is currently no such "condition" in TradeLot type for now. so i record your comment, but i'm not changing PR in that regards, right ?

JBZ-Fragmos:
@lolabeis 

thanks a lot for reviewing my PR

after considering your feedback and providing back to you explanations etc. 

have made one change overall 
image

please review last PR update
image

thanks !

jb

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: > accordingly, i simply suggest to restrict cardinality to [1] instead of [*] - which have now changed in last PR version as below

In the original issue you raised, you referred to a business case for specifying multiple prices and monetary quantities:

> ADDITIONAL BUSINESS CASES : above example is the "easy" case : more kinds of quantity or price may be involved... for instance, let's keep same example, but say now there is 2nd Price involved because of FX Features at stake e.g. settlement ccy <> fixing currency, so you may need to represent not only USD/Share Price, but possibly EUR/Share Price, and may also EUR/USD Price, etc.

Why drop it now? If it's a genuine business case, reintroducing multiple cardinality later on could be painful.

lolabeis: > about end-comment to consider "creation of specific type" inside PriceQuantity

To be clear, I wasn't suggesting to change PriceQuantity. I was just noting that the proposed MonetaryQuantity type, if implemented, could be usefully redeployed in PriceQuantity. But it doesn't impact your PR.

lolabeis: > let me provide just 1 basic example where assumption is not correct : say product is a MtM CCS with EUR-fixed leg and USD-varying leg, and I want to represent the cash adjustment principal payment, that relates to the notional reset event, due to varying leg feature

I think your example proves my point? There is indeed a 1-1 correspondence between additional prices and monetary quantities being provided (and yes, in exchange rate cases, the currency unit in question could be over or under, depending on the quote basis)

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: > notionalQuantity is the far closer aligned with market practice and also with multiple ISDA Definitions cross-asset wise, etc. than monetaryQuantity that is really more "model"

It's good that you agree that monetaryQuantity is more aligned with "model", because that's what the CDM is 😆.

To tie logical model components to market-convention names, I suggest using either the attribute's description or a `docReference` annotation, if there are specific ISDA Defns you would like to link to.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: If the payout is cash-settled, the Transfer object will be populated with asset = cash, not commodity.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: The condition I'm referring to _is_ implemented, but at the TradableProduct level (arguably it could be implemented directly in TradeLot, or even in PriceQuantity - but it's still active). So I'm requesting to adjust the PR.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: that is key point for me : 

- as far as model can keep closer to authoritative source - typically association source e.g. ISDA, ICMA, ISLA - then we should do it, so given that "notional" is used mostly in every such source documentations, then that should be preferred terms by default, for obvious readability of the model by both implementors and end-user which eventually are refering in concrete business to such authoritative sources, that coming with a lot of operational, compliance and legal implications, another reason to keep in mind that proximity of the model with related association concepts is major driver for adodption (and is undermining adoption in the opposite way)

- for the avoidance of doubt, above principle is NOT equivalent to saying that "model items shall always equal to authoritative source items" meaning
    - above principel is simply a "best effort" , or say "design principel" 
    - meaning that, whenever not possible, whenever there is technical or functional reason not to do so, or conflict between multiple sources and none is better than another one, etc. then model indeed remains ultimatly free in regards of the "authoritative sources", that I fully agree with you.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: that is already taken into account with condition below : 

image

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: sorry, i do not have the skill for drafting this condition, so i'm kindly asking you to do it, if you beleive that is a blocker to the PR - thanks a lot in advance for that.

have prepared the description language that per se is saying wjat we expect to enforce via condition, but again I'm very sorry i'm not able to draft it, please help : 

image

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: i understand that : 

- ounbound cardinality is OK for you (so will be for me as well, as you say, that was original request) - hence reintroducing the ounbound cardinality in new last PR version
- AND that the "triangulation" condition should be drafted accordingly i.e. the condition must ensure that quantity x price = notionalQuantity for at least one of the pairs of notionalQuantity and price, meaning that eventually, it  is permitted to optionally populate multiple notionalQuantity and multiple price (via unbound card), provided at least for one of them the "triangulation" can be verified in regards of the unique quantity being transferred

regnosys-prod-user:
_What is being released?_

changes compared to prior PR version - all about AssetFlowBase attributes :
- prior bounded cardinality is relaxed i.e. unbounded cardinality for both: price AND notionalQuantity
- prior "deliverableQuantity" is renamed "quantity"
- additional condition is being added, named "PriceQuantityTriangulation" but kindly being asked that CDM Maintainer handles final drafting (because I do not have skills to properly manipulate "func" objects for that)

> [!NOTE]

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: Not exactly. My recommendation was:
- Either you have it as multi-cardinality. In this case, both a single price and single monetary quantity should be folded into a new type, which is then used with multi-cardinality in Transfer. The triangulation rule should apply for each element in the array.
- Or you have it as single cardinality. In which case it's fine to have both that single price / single monetary quantity in Transfer, and the triangulation rule applies only once. But I understood that your business case was for multi-cardinality.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: Then we agree that in the cash-settlement case (which was your example above), the Transfer's asset is cash not commodity.

I note that in that cash case, the condition is indeed written in the correct direction:
```
if asset -> Cash exists
then quantity -> unit -> currency exists
```

I'm suggesting to keep the logic written in the same direction for the other asset types, e.g.:
```
if asset -> Commodity exists
then quantity -> unit -> capacityUnit exists
```
This is how it's currently written.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: Yes, I'll help you with that

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: I can see that in your latest PR you have reverted deliverableQuantity back to quantity, which is helpful.

On notional vs monetary naming for the other quantity, to allow you to move forward one way or another I propose to run it past the CRWG's at the next meeting (your item is in the "follow-up" column so we can come back to it).

Note that as per my comment above, you may need to introduce a new type which has both price and quantity, so the naming should apply to that type as well.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: no... 

my proposal is designed in a way that will permit to represent cash transfer in relation to another Asset
the new conditions i'm proposing permit to have for instance : 
- cash payment e.g. quantity -----> currency Unit
- that relates to an Asset that might be Commodity, or more generally any kind of Asset involved in the Product with regards to which the cash payment is happening

let's consider a Basket TRS for instance, say it is Cash-Settled... Imagine there are 5 stocks in the Basket : stock A, stock B, stock C, stock D, stock E.... Now, on the Valuation Date / Payment Date, one of the Party has to pay for the related Performance. Say the Party wants to pay gross i.e. want to transfer cash amount for per each stock... Then the Party will need to represent multiple transfer, each associated to particular Asset

My proposal notably to represent it (provided current conditions are modified as proposed)

Each individual transfer will look like : 

- quantity = cash being transferred so Unit = Currency Unit
- asset = the particular stock [X], which Unit = Financial Unit

Current condition simply prevents to represent such multiple transfers

My condition proposal is well drafted for permitting that such cash transfer might relate to any kind of asset 
Also it is still enforcing consistency when the transfer is physically settled, because in that case, necessarily, the Unit of the quantity being transferred and the Asset to which the transfer is relating to must be the same

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: THANKS

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: i'm happy the topic is being discussed - please keep me posted, and please , if "escalating" the point to CRWG ensure the rationale behind it is fairly explained to participants I mean my core point is that it should be a rule in "best effort" by default, that 

- whenever feasible we shall prefer selecting a name that is equal or closer to one or several concepts named as such in some association authoritative document (typically ISDA Def + similar sources for other associations e.g. ICMA/ISLA)
- else if cannot be defined because there is deep model constraint or no common concept exist across multiple souces and/or exist with conflictual meaning, etc. then for sure we would select any name as appropriate given context and consensus

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: sorry but why are you not accepting my proposal on this point ? 

i'm considering your proposal, and for sure' as a maintainer you notably make proposals, suggestions, etc. but here looks like my proposal is Ok per se and still you are not validating it, and you re-state yours, meaning looks like you are "blocking" mine and I honestly do not understand why because my proposal looks to me both (a) fair from modelling design principle and (b) it is applicable together with no functional regression  

I mean : 
- (a) why it is fair / sound modelling method,  that I have already explained : my proposal re-use existing type (moreover fully consitent with other current type PriceQuantity that is good per se if we want to use further referencing / annotation, etc.) thus avoiding the creation of a new type when not strictly required, and honestly I do not understand why you are not supporting me on this : don't you agree with me that "blottering" the model with new types, locally used here and there is not scallable approach and is per se undermining readability / consistency and maintance of the model ?
- (b) it is applicable together with no functional regression because the "triangulation" can be drafted and applied as you recommend, i guess simply by adapting it to local situation by drafting it in such a way that we expect "at least one triangulation in case we have multiple price / multiple quantity "

I might have missed some blocking issue or inconsistency in my proposal, but if so, please elaborate, 

else please re-consider your position about it and hopefully based on all details above provided to explain it, let me know if you would eventually agree with it

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: Note that I raised a related bug #4109 re this quantity triangulation. It will need to be addressed in support of your PR.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: My recommendation is based on a design consideration for manipulating arrays.

When you have two arrays whose elements are associated pairwise, it's better to implement an array of pairs than a pair of arrays. Otherwise the pairwise consistency has to be retro-fitted as a clumsy condition (it also makes the triangulation rule much harder to implement). The array sizes will also need to be matched. And regardless of these conditions, because it not clear to a downstream consumer that this pair of array attributes are, in fact, associated pairwise, they could "pick and mix" without realising that they also need to enforce the same pairwise consistency.

Natively pairing elements in a single array solves these problems.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
lolabeis: Thank you for clarifying the change's intent, this is helpful.

This puts the proposal under quite a different light: it's not just "enriching" the Transfer object as the original title suggested, it's modifying its existing behaviour, since the Transfer's asset attribute may not be what's _actually_ transferred and the quantity may no longer match the asset's unit.

I will take this discussion back onto the issue, for the sake of transparency with the CRWG.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
JBZ-Fragmos: thanks for clarifying, I fully understand what you say, that is correct in my opinion... thing is that my proposal is logical as well... eventually would say there are pros/cons in either proposals... hummm.... still in that regards, would recommand to implement mine , not because would be intrinsically the best  but because that is not purpose of this ticket to open such considerations i mean my proposal again does not create new type, and if so, then the real topic behind this is PriceQuantity

so let me rephrase this : for now the type I want to use for modelling Price and Quantity under AssetBase is on purpose aligned on the type used in the PriceQuantity attributes (= QuantitySchedule and PriceSchedule) because I consider that these component are VERY KEY to the whole model, and so we MUST ensure consistency with it AS FAR AS WE CAN : i can tell by experience that whenever some concept of price or quantity are involved in Trade, Lifecycle events, etc. that do not re-use original types from PriceQuantity attributes, then this causes lot of limitations (main one is that referencing is not possible) and make implementation complex and difficult to maintain on scallable basis, whether simply "lookup based" functions or more synthetic including calculations : in all cases, life is less easy when too many different types co-exist for same concept of Price and Quantity

so if we want to open the topic of reviewing the structure of PriceQuantity and related attributes (= QuantitySchedule and PriceSchedule) why not but that should be separte ticket - and until this has not happened yet, definitly prefer using current existing types (= QuantitySchedule and PriceSchedule) 

what do you think ?

(of records : my opinion on current PriceQuantity type is that enrichment would be needed, but without undermining the current array structure :
- in my view : TradeLot->PQ is missing new attribute or new type associated with it for the purpose of clarifying which / where is the final net/resolved position among multiple PQ ; that would consists in **adding** something new...
- but at first glance, I'm not in favor of **changing/removing/deprecating** the rich array structure that currently exist to replace it for instance with "arrays of pairs" as per our discussion above, because in some cases, you need more than a pair, you need "n" that is to say you need unboubnd array...)

ReleaseManagement-BP:
Closing this PR as the linked issue has been closed

PR #4001, opened 2025-09-02

regnosys-prod-user:


regnosys-prod-user:
# *CDM Event Model - Termination for Schedules*

_Background_

When a termination applies to a quantity schedule, all the dated values from the effective period onwards should be set to zero.

_What is being released?_

The `UpdateQuantityAmountForEachMatchingQuantity` function has been enhanced to apply the quantity change to all dated values from the current period onward. A new function `UpdateDatedValues` has been created to perform the update of the `DatedValue`.

The period from which the change should take effect is determined using the `primitiveInstruction → quantityChange → change → effectiveDate`.

_Review Directions_

Changes can be reviewed in PR: [#4001](https://github.com/finos/common-domain-model/pull/4001)

> [!NOTE]

rosetta-source/src/main/rosetta/base-math-func.rosetta
```diff
@@ -133,6 +133,28 @@ func UpdateAmount: <"Updates an amount based on the given QuantityChangeDirectio
...
+
+    add newdatedValues:
+    datedValues
+        extract
+            DatedValue {
+                date:  item -> date,
+                value:
+                if item -> date >= effectiveDate then (
+                if direction = QuantityChangeDirectionEnum -> Increase then item -> value + changeAmount
+                else if direction = QuantityChangeDirectionEnum -> Decrease then item -> value + changeAmount
```
chrisisla: Is `changeAmount` going to include the sign e.g. -33? If not then shouldn't this be `item -> value - changeAmount`.

rosetta-source/src/main/rosetta/product-common-settlement-func.rosetta
```diff
@@ -27,7 +28,7 @@ func UpdateAmountForEachMatchingQuantity: <"Updates any price or quantity from t
                         extract
                             UpdateQuantityAmountForEachMatchingQuantity(
                                     item,
-                                    change -> quantity, // FilterChangePriceQuantity(priceQuantity, change) -> quantity,
+                                    change, // FilterChangePriceQuantity(priceQuantity, change),
```
chrisisla, line 31: May as well remove the `FilterChangePriceQuantity...` comment now.

rosetta-source/src/main/rosetta/product-common-settlement-func.rosetta
```diff
@@ -76,20 +77,31 @@ func UpdateQuantityAmountForEachMatchingQuantity: <"Updates any quantity from th
     inputs:
         quantity NonNegativeQuantitySchedule (0..1) <"List of NonNegativeQuantitySchedule to update.">
             [metadata location]
-        change NonNegativeQuantitySchedule (0..*) <"List of new NonNegativeQuantitySchedule to use where the units match.">
+        change PriceQuantity (0..*) <"List of new PriceQuantity to use where the units match.">
```
chrisisla: This function is accepting a `NnonNegativeQuantitySchedule` as the source data i.e. a quantity, but you have changed the search term to now be a `PriceQuantity` i.e. a quantity and a price (and an observable potentially too).

Are we looking to have separate functions to update quantities and prices or is this function going to now be able to update both? 
- If it's the former then we should leave the search term as being a `NonNegativeQuantitySchedule` i.e. keep it tied to quantities only.
- If it's the latter then we should consider updating the name of the function to show it operates on either a price or a quantity.

I think we just need a bit more clarity on what the overall goal is for these functions. Do we want an all in one PriceQuantity change function, or separate functions for prices and quantities.

regnosys-prod-user:
_What is being released?_

Comments added in the code.

> [!NOTE]

regnosys-prod-user:
_What is being released?_

fixed sign for decrease changed amount

> [!NOTE]

rosetta-source/src/main/rosetta/base-math-func.rosetta
gtarres: @chrisisla totally agree, thank you for pointing that out. The sign has been fixed in this commit: https://github.com/finos/common-domain-model/pull/4001/commits/8373fe87f450ed46c5687d574d6ddba9fddcd7ba

rosetta-source/src/main/rosetta/product-common-settlement-func.rosetta
gtarres: Thank you @chrisisla for the comments. I leave the answer we discussed from the CRWG 9th to keep track:

The function input has changed from `quantity` to `PriceQuantity` because `PriceQuantity` corresponds to the type of the higher level (`change`) that wraps both the `quantity` and its associated `effectiveDate`. This is needed to keep the `effectiveDate` linked with the `quantity` so that the logic contained in the function has the correct effective date for the applicable quantity, ensuring we don’t lose the linkage between the quantity and the corresponding effective date.


As we agreed, I've added a brief explanation in lines 32 and 81 from the code (see commit https://github.com/finos/common-domain-model/pull/4001/commits/fcb608fc63ca6f1360e72431b10313b6416161b6).

rosetta-source/src/main/rosetta/product-common-settlement-func.rosetta
gtarres: Hi @chrisisla, I think it’s better to keep it, as the comment is still relevant if the function `FilterChangePriceQuantity` becomes operative in the future. I’ve updated it: before it was `FilterChangePriceQuantity -> quantity`, and now it just stays at the `change` level.

gtarres:
The mapper in PR https://github.com/finos/common-domain-model/pull/4012, built by @hugohills-regnosys, should be contributed together with this PR since it includes the mapping of the effective date from FpML to the quantity change primitive instructions.

chrisisla:
Hi @llynhiavu 
Would you mind reviewing this please so we can get it merged.
Thanks!

CDM-ReleaseManagement-OT:
@JayasriR can you please investigate why the build is failing?

rosetta-source/src/main/rosetta/base-math-func.rosetta
```diff
@@ -133,6 +133,28 @@ func UpdateAmount: <"Updates an amount based on the given QuantityChangeDirectio
         Decrease then oldAmount - changeAmount,
         Replace then changeAmount
 
+func UpdateDatedValues: <"Updates the amounts of the dated values based on the given QuantityChangeDirectionEnum. For all periods after the effective date of the change, if the direction is Increase, the old amount and change amount are summed, if the direction is Decrease, then the change amount is subtracted from the old amount, and if the direction is Replace then the change amount replaces the old amount.">
+    inputs:
+        datedValues DatedValue (0..*)
+        changeAmount number (0..1)
```
llynhiavu, line 139: Any need to enforce the expected changeAmount must be positive? (perhaps a broader question)

llynhiavu (changes requested):

rosetta-source/src/main/rosetta/base-math-func.rosetta
gtarres: The `changeAmount` corresponds to the `value` of the quantity within the `quantityChangeInstruction`. It is defined by the `NonNegativeQuantitySchedule` type, which contains a constraint to ensure it is a positive value.

image


image

Issue #3626, opened 2025-04-07
Missing Support for Schedule Quantities in `Create_QuantityChange` Function


gtarres:
### Background

The current implementation of the `Create_QuantityChange` function does not account for schedules. When a termination is applied to a schedule, the quantity change is only applied to the tradelot's quantity (`tradeLot → priceQuantity → quantity → value`). However, the quantities in the dated values (`tradeLot → priceQuantity → quantity → datedValue → value`) are not updated. 

Specifically, for the case of terminations, when a termination applies to a quantity schedule, all the dated values from the effective period onwards should be set to zero.

### Proposal

To address this issue, the `UpdateAmountForEachMatchingQuantity` function (implemented in Java, and used in the `Create_QuantityChange`) should be enhanced to apply the quantity change to all dated values from the effective period onward. 

The period from which the change should take effect can be determined using the `primitiveInstruction → quantityChange → change → effectiveDate`. This enhancement will ensure that quantity changes are correctly applied to all relevant dated values in a schedule.

### Compatibility

This change would be backward-compatible as it extends the coverage to schedules without affecting the existing functionality. The target CDM versions are CDM 5, 6, and 7dev.

CDM-ReleaseManagement-OT:
Hi @gtarres, is this issue ready for discussion on the CRWG? I moved to current

gtarres:
> Hi [@gtarres](https://github.com/gtarres), is this issue ready for discussion on the CRWG? I moved to current

Yes, it is. Thank you,

CDM-ReleaseManagement-OT:
Discussed on [CDM Contribution Review Working Group - May 20th, 2025](https://github.com/finos/common-domain-model/discussions/3675)

@gtarres to proceed with proposal, with further analysis required for edge cases where datedValues fall between 2 dates in the schedules.

chrisisla:
CRWG - 03/06 - Manuel is investigating, ties into the whole QuantityChange discussion.

gtarres:
The mapper in PR https://github.com/finos/common-domain-model/pull/4012, built by @hugohills-regnosys, has been contributed to map the effective date from FpML to the quantity change primitive instructions.

llynhiavu:
approved PR with a minor question.

PR #4012, opened 2025-09-02
Add mapper for schedules decrease effective date

hugohills-regnosys:


gtarres:
@CDM-ReleaseManagement-OT This PR is no longer needed and can be closed. The fix for ISSUE #3626 will be released after the legacy mappings are deprecated, so no mapper is required. Thank you,

PR #4028, opened 2025-09-11

regnosys-prod-user:


gtarres:
The mapper in PR https://github.com/finos/common-domain-model/pull/4012, built by @hugohills-regnosys, should be contributed together with this PR since it includes the mapping of the effective date from FpML to the quantity change primitive instructions.

ReleaseManagement-BP:
Closing this PR as the comment above mentions

PR #4029, opened 2025-09-11

regnosys-prod-user:


gtarres:
This PR (https://github.com/finos/common-domain-model/pull/4029) should be closed without merging because it introduces backward-incompatible changes. In essence, it includes an enhancement to support the representation of terminations for schedules and does not directly affect the DRR reports.

ReleaseManagement-BP:
Closing this PR as the above comment mentions

PR #4042, opened 2025-09-20
Georgina Tarres CDM Event Model - Qualification for Termination and Partial Termination

### LINKED ISSUE #3798: The qualification logic for Terminations doesn't align with the changes expected for this event
### Issue Summary

The logic to qualify terminations includes a check to ensure that the quantities in the `tradeLot` have been decreased to zero. This is done by the function `CompareTradeLotToAmount` (called within the `QuantityDecreasedToZero` function).

This function only returns true if *all* quantities in the `tradeLot` have been reduced to zero:

Image

This is missaligned with the Termination event, since for certain products (e.g. bond options), there are values in the TradeLot that are not reduced to zero, and we do not expect them to be reduced.

To solve the issue, we should look into the logic applied by the function `UpdateAmountForEachMatchingQuantity` to select the `tradeLots` where the quantity is to be reduced to zero and apply the same criteria at the qualification stage.

### Steps to Reproduce

In a IR Debt Option, there are the following quantities in the `tradeLot`:
- The notional amount: total traded notional amount of the underlying debt option trade
- The option entitlement (this is the name in FpML): The number of units of underlyer per option comprised in the option transaction. 

The option entitlement it's a feature of the product, therefore, we do not expect this quantity to be reduced to zero. Being the notional amount the only quantity we expect to be reduced to zero. However, then the event is not qualified as a termination.

This behaviour can be proven in the following sample (mock up):
[ir-option-debt-option-ex01-output-term-cdm-proposed.json](https://github.com/user-attachments/files/20706655/ir-option-debt-option-ex01-output-term-cdm-proposed.json)

Where the quantities are:
 - The notional amount: 600000000 USD 
- The option entitlement: 100000000 USD

### Expected Behaviour

Should qualify as Termination, even if one of the quantities on the `tradeLot` is not zero.

### Actual Behaviour

Doesn't qualify as Termination.

### Compatibility

This change is backward compatible.

### Release

CDM 7dev.

### LINKED ISSUE #3987: Partial Termination Not Qualified properly
### Issue Summary

When processing a Partial Termination event, the `QuantityDecreased` function enforces two strict conditions:
- *All* quantities must decrease.
- *All* quantities in the after state must be *greater than zero*.

Certain scenarios of partial terminations could not meet both conditions above. Specifically, the system does not qualify it correctly in the following two cases:
1. If only one quantity is reduced and the others remain with the same value.
2. If a quantity in the trade is zero.

In cases where a quantity is decreased, and all the other quantities remain unchanged, the model should consider it a decrease (partial termination).

### Steps to Reproduce

Run the function _QuantityDecreased_ on the sample [msg-partial-termination-xccy-mockup-one-qty-CDM.json](https://github.com/user-attachments/files/21972452/msg-partial-termination-xccy-mockup-one-qty-CDM.json) and observe how the qualification is False:

Image

Run the function _QuantityDecreased_ on the sample [msg-partial-termination-xccy-mockup-qty-zero-CDM.json](https://github.com/user-attachments/files/21972470/msg-partial-termination-xccy-mockup-qty-zero-CDM.json) and observe how the qualification is False:

Image

### Expected Behaviour

Since a quantity is being decreased, it should be captured as such and return *True*.

### Actual Behaviour

Currently, it returns *False* if all quantities are not decreased or if any is equal to zero.

### Compatibility

Changes are backward-compatible.

### Release

CDM 5 prod, CDM 6 prod, CDM 7 dev.

### Additional Context

_No response_

regnosys-prod-user:


regnosys-prod-user:
# *CDM Event Model - Qualification for Termination and Partial Termination*

_Background_

The current qualification logic for Termination and Partial Termination events imposes conditions on all `tradeLot` quantities. This is incorrect because, for certain products (e.g. bond options), some `tradeLot` values represent trade features that are not expected to be impacted by the event. Therefore, they should not be considered in the qualification logic, as they can lead to misqualifying the event. Event qualification should instead be based only on the values impacted by the event, regardless of other unaffected quantities.

This release updates the logic for qualifying partial terminations and full terminations by applying constraints only to relevant quantities.

_What is being released?_

This release updates the following functions:

### Partial Termination qualification

* **`QuantityDecreased`**: updated to check that changed quantities have not decreased to zero, instead of checking that all quantities in the resulting `tradeLot` are greater than zero. This check is done through the function `QuantityDecreasedToZero`.

* **`CompareTradeLot`**: updated to succeed if at least one quantity meets the condition.


### Termination qualification

* **`QuantityDecreasedToZero`**: updated to take the `instruction` as input and extract the `unitType`s from the `quantityChange` instructions. The constaint now is applied only to quantities with the same unit type — assuming those are the ones affected by the event.

* **`CompareTradeStatesToAmount`**: updated to allow `unitType` to be specified as input. If unit types are specified, only quantities with those units are compared. Otherwise, the comparison applies to all quantities.


* **`CompareTradeLotToAmount`**: updated to allow `unitType` to be specified as input. If unit types are specified, only quantities with those units are compared. Otherwise, the comparison applies to all quantities.



### Termination construction

* **`Create_QuantityChange`**: updated to set `positionState` to `Closed` if all quantities with the instruction’s unit have decreased to zero.


### Additional updates

* Minor technical adjustments were made to the following functions to accommodate the above changes, without any impact on the business logic:
  `Qualify_FullReturn`, `Qualify_PartialNovation`, `Qualify_PartialTermination`, `Qualify_Reallocation`, `Qualify_Repurchase`, `Qualify_Renegotiation`, `Qualify_Termination`.

_Review Directions_

Changes can be reviewed in PR: [#4042](https://github.com/finos/common-domain-model/pull/4042)

> [!NOTE]

gtarres:
This PR solves issues https://github.com/finos/common-domain-model/issues/3987 and https://github.com/finos/common-domain-model/issues/3798.

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -38,10 +38,13 @@ func QuantityIncreased:
 
 func QuantityDecreased:
     inputs:
-        before TradeState (1..1)
+        instruction Instruction (0..*)
```
chrisisla: Would we not always expect at least one `instruction`? I guess as long as the function returns `False` in this scenario then it would be okay to have (0..*) rather than (1..1) cardinality.

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -50,45 +53,57 @@ func QuantityDecreased:
...
+                            instruction -> before -> trade -> tradeLot only-element
                         ) = True
+                    and // Check that changed quantities have not decreased to zero
+                    QuantityDecreasedToZero(instruction, after) = False
             ] all = True
 
 func QuantityDecreasedToZero:
     inputs:
-        before TradeState (0..*)
+        instruction Instruction (0..*)
```
chrisisla: Same question of cardinality as `QuantityDecreased` function. Interesting that originally `tradeState` was (0..*) here but (1..1) for `QuantityDecreased`.

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -50,45 +53,57 @@ func QuantityDecreased:
...
+
     output:
         result boolean (1..1)
 
-    set result:
+    alias resultAll: <"Result of applying the comparison to all quantities, regardless of unit.">
         tradeStates
             extract [
-                CompareTradeLotToAmount(item -> trade -> tradeLot only-element, op, amount)
+                CompareTradeLotToAmount(item -> trade -> tradeLot only-element, op, amount, unitType)
```
chrisisla, line 94: Should this not exclude `unitType`? If this is included then the `resultAll` and `resultUnit` aliases will hold the same results?

rosetta-source/src/main/rosetta/event-common-func.rosetta
```diff
@@ -988,9 +1003,11 @@ func Create_QuantityChange: <"A specification of the inputs, outputs and constra
     set quantityChange -> trade -> ancillaryParty: trade -> ancillaryParty
     set quantityChange -> trade -> adjustment: trade -> adjustment
 
-    set quantityChange -> state -> positionState:
-        if newTradeLots -> priceQuantity -> quantity -> value all = 0
-        then PositionStatusEnum -> Closed
+    set quantityChange -> state -> positionState: <"Set positionState to Closed if all quantities with the instruction’s unit have decreased to zero.">
```
chrisisla, line 996: Not convinced this is correct. For a stock loan against cash you would have 2 quantities, one holding the securities (unit = financialUnit -> Share) and one holding the cash (unit = currency). Both of these quantities would need to be set to zero in order to set the `positionState` to "Closed". 
If the QuantityChange instruction only decreases one of these quantities to zero then this condition I think would still set the trade to Closed.

chrisisla (changes requested):
I've still got a couple of changes to review so will add some further comments when I am able to pick this up again (which should be this week) 👍 

Main thing I need to think about is whether there is anything that might prevent the `QuantityDecreased` functions from taking an `instruction` as input rather than a `tradeState`.

rosetta-source/src/main/rosetta/event-common-func.rosetta
gtarres: If the instructions were absent it would simply return `False`, without causing any issue. However, I agree that setting the minimum cardinality to 1 makes more sense, since the function expects an instruction to work correctly, so I agree it would be more reasonable to set it as (1..1). 

I can apply the changes to both functions: `QuantityDecreased` and `QuantityDecreasedToZero`.
Thank you @chrisisla !

rosetta-source/src/main/rosetta/event-common-func.rosetta
gtarres: I totally agree. Moreover, the comparison between `resultAll` and `resultUnit` is redundant in this function because `CompareTradeLotToAmount` already performs this check.

rosetta-source/src/main/rosetta/event-common-func.rosetta
gtarres: I agree. If we can have quantities with mixed types of units related to the actual quantities exchanged, then we'll need to rethink this logic.

This has been modelled from a Debt Option perspective, where we expect the trade to be closed even if the option entitlement quantity, which has a unit type of `funancialUnit`, is not reduced to zero, as this quantity is just a feature of the contract.

image

chrisisla: Ah okay, what we may need to do then is check the product to determine what constitutes a Termination, or have separate functions to qualify termination for different product types. Either way it looks like it is not as straightforward as we'd hoped 🙁

ReleaseManagement-BP:
Closing this PR as this has been open for a considerable time without any movement. If you would like the issue re-opened please reach out

Issue #4039, opened 2025-09-17
FpML synonym mapping of payer and receiver incorrect for FX (for CDM 6.x.x and 7-dev)


hugohills-regnosys:
### Issue Summary

TODO

### Steps to Reproduce

TODO

### Expected Behaviour

TODO

### Actual Behaviour

TODO

### Compatibility

TODO

### Release

The bug affects the following versions:
- CDM 6.x.x (synonyms)
- CDM 7-dev (synonyms)

PR #4063, opened 2025-09-30
[6.x.x] STORY-4102: Fix FX payer receiver mappings from FpML (synonym)

hugohills-regnosys:


rosetta-source/src/main/resources/result-json-files/fpml-5-10/products/fx/fx-ex08-fx-swap.json
```diff
@@ -274,14 +274,14 @@
     "counterparty" : [ {
       "role" : "Party1",
       "partyReference" : {
-        "globalReference" : "4a5d2d9f",
-        "externalReference" : "party1"
+        "globalReference" : "a41bc6e9",
+        "externalReference" : "party2"
```
lolabeis, line 278: Party1 (enum) becomes party2 (external ref) and conversely, which could be confusing.

I would have expected Counterparty to stay unch'ed, and PayerReceiver to flip instead.

lolabeis (commented):
The mapping change now produces the correct output for fx spot, forward and swap, but in a lopsided way. It's the Counterparty object that is flipped instead of the PayerReceiver.

Is there a way to flip PayerReceiver instead?

rosetta-source/src/main/resources/result-json-files/fpml-5-10/products/fx/fx-ex08-fx-swap.json
lolabeis: Discussed offline - Not possible with the way synonyms currently work.

RELEASE.md
```diff
@@ -1,3 +1,29 @@
...
+
+This release fixes the FpML ingestion (based on synonyms) mapping of payer and receiver for FX products, as per GitHub issue [#4039](https://github.com/finos/common-domain-model/issues/4039).
+
+_What is being released?_
+
+Update to the FpML synonym mappings for `SettlementPayout` attributes `payer` and `receiver` to correctly correspond the exchange rate quote basis.
+
+_Review Directions_
+
+In Rosetta, select the Ingest tab, select `FpML_5_Confirmation_To_TradeState` and review the following FpML samples:
```
llynhiavu, line 13: I was looking at the suggested examples in Rosetta CDM 6.9.0 Prod > Ingest > ...

I assume these are corrected already?? can you point me to a version where the examples are incorrect? should I look at 5.27.0?

hugohills-regnosys: No, this is still an issue in CDM version 6.9.0.  Have a look at the Actual Behaviour section in the Issue https://github.com/finos/common-domain-model/issues/4039, then review sample `fx-ex01-fx-spot.xml`.

In CDM 6.9.0, the payer on the SettlementPayout is incorrectly set as LEI `5493000SCC07UI6DB380`.

In this PR, the payer on the SettlementPayout is correctly set as LEI `529900DTJ5A7S5UCBB52`.

See https://github.com/finos/common-domain-model/pull/4063/files#diff-0c04fbfc1486b21977108123553051f128aebec298e8b2d3753c67a478250b91

Screenshots below
image
image
image

llynhiavu: understood. Thanks Hugo.

PR #4119, opened 2025-10-19
[6.x.x] Contribute FpML Confirmation Ingest functions for CDM 6

hugohills-regnosys:


hugohills-regnosys:
Should be released on a special tag, e.g., `6.10.0-ing.1`

Issue #3925, opened 2025-07-24
Regularly Broken Builds: "Pipeline could not be executed because codefresh.yaml file is missing" during docs updates


eacunaISDA:
### Issue Summary

ISSUE
We're still experiencing broken builds, during documentation updates that do not change the model. 

TYPE/LABEL:
Documentation PRs (no code changes)

QUESTION
Would it be possible for the pipeline / actions be updated somehow to cater for this type of PR? It appears the build is failing erroneously.

IMPACT
PR's with a broken build will not be approved, and they accumulate over time becoming a bottleneck. 

### Steps to Reproduce

Behavior observed: There seems to be a pattern: it seems to happen more frequently to documentation PRs where some of the changes have been made in GitHub themselves (not as pushed PRs).

To view error, click on details for the unsuccessful check.

### Expected Behaviour

Build terminates successfully, since all other checks are successful and the deploy preview testing has a successful outcome.

### Actual Behaviour

Build terminated with message "Failed to retrieve pipeline" and supposedly 'missing' yaml file. 

<img width="749" height="280" alt="Image" src="https://github.com/user-attachments/assets/8518a0ce-dcfe-47c5-bcbd-5d235c166121" />

### Compatibility

N/A

### Release

Should work for all

CDM-ReleaseManagement-LSG:
Raised to REGnosys team for validation

PR #4127, opened 2025-10-24
Add documentation check to skip build if only docs change

gliptak:
needs to be tested on main repo
alternative solution could be to convert into Github Actions

#3925


ReleaseManagement-BP:
Closing this PR as this has been open for a considerable time without any movement. If you would like the issue re-opened please reach out

PR #4190, opened 2025-11-14
Adrian Jenks Legaldocumentation - Enhance the Umbrella functionality within the Legaldocumentation section of the CDM.

regnosys-prod-user:


regnosys-prod-user:
# *Legaldocumentation - Enhance the Umbrella functionality within the Legaldocumentation section of the CDM.*

_Background_

The Umbrella functionality within the CDM required to be built out to meet the requirements of the members to support the capture of Legal Agreement terms within an Umbrella Structure.

The original solution only allowed for the identification that the agreement had an Umbrella structure and allowed the capture of term election variations through the use of a single string variable.

The proposed solution creates an Umbrella structure that is mapped to the existing Legal documentation agreement election structures while allowing multiple agreement sets to be created that identify different election combinations and allows the agreement set to be defined against 1 or more parties to the agreement.

Parties to the agreement are also assigned roles identifying the legal and business purpose of those entities within the agreement.

_What is being released?_

Updated types and enumerated lists have been added to the legal documentation component of the CDM as listed below:

Enum
- UmbrellaPartyRoleEnum - Represents the legal role a party is assigned for the agreement

Type
- UmbrellaAgreementSet - Allows multiple sets of elections to be captured and assigned to specific parties to the agreement
- UmbrellaAgreementEntity - Defines the parties to the agreement
- Parentparty - Allows parties to be associated with a parent party on the agreement. Within umbrella agreements funds, portfolio or managed accounts which are not legal entities can be defined that are required to be rolled up to a parent party for things such as margin management. Additionally multiple investment managers may exist within a single agreement and parties need to be associated with the applicable Investment Manager.

_Review Directions_

Changes can be reviewed in PR: [#4190](https://github.com/finos/common-domain-model/pull/4190)

> [!NOTE]

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
 
 type UmbrellaAgreement: <"A class to specify a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement. This data representation reflects the ISDA Create representation.">
```
chrisisla: Not added as part of this PR but the final sentence referencing ISDA Create should be removed as this will be used more broadly now

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
 
-    isApplicable boolean (1..1) <"The determination of whether Umbrella Agreement terms are Applicable (True), or Not Applicable (False)">
-    language string (0..1) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
-    parties UmbrellaAgreementEntity (0..*) <"Underlying principals to the umbrella agreement.">
+    agreementSet UmbrellaAgreementSet (1..*) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
+    parties UmbrellaAgreementEntity (2..*) <"Underlying principals to the umbrella agreement.">
```
chrisisla: Can we clarify whether we're talking about parties, entities or principals please? The attribute is named `parties` but the type is `UmbrellaAgreementEntity`. There is some ongoing discussion around a legal entity vs a party which might be useful here 🙂

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    language string (0..1) <"Represents any bespoke language that may need to be captured regarding the structure or application of terms that may exist for the agreement.">
```
chrisisla: better perhaps to call the attribute `bespokeLanguage`?

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
 
-    condition UmbrellaAgreementExists: <"Umbrella Agreement language and parties should not exist when Umbrella Agreement terms are Not Applicable.">
-        if isApplicable = True then language exists and parties exists
+type UmbrellaAgreementEntity extends Counterparty: <"A class to specify the legal entities that are part of the umbrella agreement.">
+    umbrellapartyrole UmbrellaPartyRoleEnum (0..1) <"Represents the legal role that each defined entity to the agreement performs within the structure of that agreement.">
```
chrisisla: `umbrellaPartyRole` rather than `umbrellapartyrole`

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    parentparty Parentparty (0..1)
```
chrisisla: `parentParty` and `ParentParty` perhaps?
And perhaps add a comment next to it.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
 
-type UmbrellaAgreementEntity extends LegalEntity: <"A class to specify the legal entities that are part of the umbrella agreement.">
+    condition Parentparty: <"Identifies where a parent child relationship exists between two parties to the Umbrella agreement, allowing none legal entities to be linked to the appropriate legal entity and allowing funds, protfolio or managed accounts to be linked to their applicable Investment Manager or Agent within the agreement.">
+   if umbrellapartyrole = UmbrellaPartyRoleEnum -> Sleeve then parentparty exists
+   or
+   if umbrellapartyrole = UmbrellaPartyRoleEnum -> Agent then parentparty is absent
 
-    terms string (0..1) <"The terms that might be associated with each party to the umbrella agreement.">
+type Parentparty: <"A class to specify the heirarchical structure of the entity relationships within the Umbrella agreement.">
```
chrisisla: ParentParty
heirarchical = hierarchical 
Can you elaborate a bit on what the role of the parent party is please? From previous comments it seems to be an important way of linking parties/entities to determine a hierarchy. Just wondering if strings are the best way to link them?

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+   parentIdentifier string (0..1) <"Represents the Entity Identifier of the parent entity on this agreement for the party defined.">
```
chrisisla: What is the Entity Identifier going to be? Would this be a Legal Entity and thus could use the `LegalEntity` type?

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
```diff
@@ -280,3 +280,13 @@ enum TerminationCurrencyConditionEnum:
     FreelyAvailable <"A currency that is freely available.">
     PaymentsDueAndFreelyAvailable <"A currency in which payments would be due under one or more Transactions and that is freely available.">
     Specified <"Termination Currency Conditions are specified.">
+
+enum UmbrellaPartyRoleEnum:
+    Agent <"Represents a Legal Entity who has authority to negotiate, execute and deliver confirmations on behalf of each affiliate that is linked to the Agent.">
+    InvestmentManager <"Represents a Legal Entity who has been authorised to act as a centralised authority empowered to negotiate, execute and manage derivative transactions on behalf of multiple affiliated funds or accounts.">
```
chrisisla: can drop the word "derivative" in the comments

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
```diff
@@ -280,3 +280,13 @@ enum TerminationCurrencyConditionEnum:
...
+    Principal <"Represents a Legal Entity that enters into and assumes direct responsibility for derivative transactions.">
```
chrisisla: as above

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
```diff
@@ -280,3 +280,13 @@ enum TerminationCurrencyConditionEnum:
...
+    Sub Account <"Represents an individual trading entity, fund, portfolio or managed account associated to a principal, Investment Manager or Agent LEgal Entity.">
```
chrisisla: "SubAccount" rather than "Sub Account" i.e. no space please. 
LEgal = Legal

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
```diff
@@ -280,3 +280,13 @@ enum TerminationCurrencyConditionEnum:
...
+    Sleeve <"Represents a distinct trading strategy, portfolio, or sub account amanged within a broader legal entity or fund. It is not a legal party to the agreement.">
```
chrisisla: amanged = managed

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
 
-    condition UmbrellaAgreementExists: <"Umbrella Agreement language and parties should not exist when Umbrella Agreement terms are Not Applicable.">
-        if isApplicable = True then language exists and parties exists
+type UmbrellaAgreementEntity extends Counterparty: <"A class to specify the legal entities that are part of the umbrella agreement.">
+    umbrellapartyrole UmbrellaPartyRoleEnum (0..1) <"Represents the legal role that each defined entity to the agreement performs within the structure of that agreement.">
+    parentparty Parentparty (0..1)
 
-type UmbrellaAgreementEntity extends LegalEntity: <"A class to specify the legal entities that are part of the umbrella agreement.">
+    condition Parentparty: <"Identifies where a parent child relationship exists between two parties to the Umbrella agreement, allowing none legal entities to be linked to the appropriate legal entity and allowing funds, protfolio or managed accounts to be linked to their applicable Investment Manager or Agent within the agreement.">
```
chrisisla: Can we remove the word "none" from the comment?
protfolio = portfolio

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -155,16 +155,25 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+   if umbrellapartyrole = UmbrellaPartyRoleEnum -> Sleeve then parentparty exists
```
chrisisla: indentation, just needs to go out one more tab to be under the condition 🙂

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: This is an old description which I have updated in the revised contribution.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Agree the naming can be confusing here. I believe Entity is the correct name in this case as this type is capturing all entities within the agreement. These could be Parties to the agreement or they could be entities that are solely managed by a party to the agreement and are not legal entities themselves. For example Sub Account, Alert Codes or Sleeves. In these situations the Entity still needs to be recorded but the party that the Entity rolls up to must also be recorded. This is important when considering Margin calculations.

I have updated the description for UmbrellaAgreementEntity

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: I have amended the description to remove the word bespoke. Amending the name from to bespokeLanguage will require Synonym changes for creatiq

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Change made as recommended

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Change to name made as suggested and comment added to provide a description of the purpose of the type

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: There are a number of entity types that exist within an Umbrella agreement structure whereby not all entities to the agreement exist as Legal Entities and therefore the parent of the Entity needs to be defined. This is especially critical where we have sub accounts, alert codes or sleeves which require managing entities to be defined. This is critical where Margin definitions need to be defined across all managed entities and not assume that each entity gets the full margin definition applied.

Additionally an Umbrella Agreement may have one or more Agents or Investment Managers defined. It is critical that the parties to the agreement are defined against the correct Agent or Investment Manager.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: As not all entities will be Legal Entities as described earlier, I believe this needs to be as open as possible allowing the user to define how they are using the Entity Identifier to accurately record relationships between entities.

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
AdrianJenksD2LT: Derivative has been removed from the description

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
AdrianJenksD2LT: Derivative removed from the description

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
AdrianJenksD2LT: Changes made as described

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
AdrianJenksD2LT: Changes made as shown

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: I think it is important to retain something in the description to state that not all Entities are Legal Entities, I have therefore removed the word none and replaced with non-legal, if this still feels incorrect I can just leave the description as entities but I want to ensure that members recognise that all form of entities to the Umbrella agreement should be captured under this type.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Additional indentation made as suggested

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
```diff
@@ -280,3 +280,19 @@ enum TerminationCurrencyConditionEnum:
...
     PaymentsDueAndFreelyAvailable <"A currency in which payments would be due under one or more Transactions and that is freely available.">
     Specified <"Termination Currency Conditions are specified.">
+
+enum UmbrellaPartyRoleEnum:
+    Agent <"Represents a Legal Entity who has authority to negotiate, execute and deliver confirmations on behalf of each affiliate that is linked to the Agent.">
+    InvestmentManager <"Represents a Legal Entity who has been authorised to act as a centralised authority empowered to negotiate, execute and manage transactions on behalf of multiple affiliated funds or accounts.">
+    Principal <"Represents a Legal Entity that enters into and assumes direct responsibility for transactions.">
+    SubAccount <"Represents an individual trading entity, fund, portfolio or managed account associated to a principal, Investment Manager or Agent Legal Entity.">
+    Sleeve <"Represents a distinct trading strategy, portfolio, or sub account managed within a broader legal entity or fund. It is not a legal party to the agreement.">
+enum AmendmentPurposeEnum:
```
paulhands: Minor, looking at it, all enums prior to this had a line of whitespace between them for readability. Would recommend following suit

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
```diff
@@ -280,3 +280,19 @@ enum TerminationCurrencyConditionEnum:
...
+    AmendTerms <"Amendment of Legal Agreement Terms">
+    AmendEntities <"Add, Remove or Amend Legal Entities to the agreement">
+    AmendandRestate <"Amend terms and restate the legal agreement elections">
+    Novation <"Novate trades from existing agreement to alternative agreement">
+    Superseding <"Terminate existing agreement and define new agreement as superseding original agreement">
+    Termination <"Termination of existing agreement">
```
paulhands: Super minor, these descriptions are missing full stops and they appear to be in the rest of the descriptions in this file

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -63,14 +63,20 @@ type AgreementTerms: <"Specification of the content of a legal agreement.">
     clauseLibrary boolean (0..1) <"Specification of whether the agreement terms have been negotiated using the clause library methodology.">
     counterparty Counterparty (2..2) <"Specification of the roles of the counterparties to the agreement.">
 
+type AmendmentTerms: <"Represents the purpose and terms associated with an Amendment document">
+    purpose AmendmentPurposeEnum (1..*) <"Purpose of the amendment">
+    impactedAgreements LegalAgreementBase (0..*) <"Test">
+    agreementTerms AgreementTerms (0..*) <"Test">
```
paulhands: Should not have "Test" as a description

AdrianJenksD2LT: Thanks Paul, those types should not have been in this contribution I will revert

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
AdrianJenksD2LT: Thanks Paul, those types should not have been in this contribution I will revert

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
paulhands: I'm presuming the `AmendmentTerms` here are capturing which agreements (which presumably will most likely just be 1) are impacted, and what the updated agreementTerms are.

Is there a way to add some validation here that will allow for a quick sanity check that the impacted agreements and the terms line up correctly? even checking just that e.g. length of agreement terms is not greater than impacted agreements would be sensible.

And final point, should both of these not be `(1..*)`. I wouldn't have an amendment term without impacting at least one agreement with related terms?

paulhands: > Thanks Paul, those types should not have been in this contribution I will revert

Ah, typed out more info before I saw reply, you can ignore my other comment

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -63,14 +63,20 @@ type AgreementTerms: <"Specification of the content of a legal agreement.">
...
 type LegalAgreement extends LegalAgreementBase: <"The specification of a legal agreement between two parties, being negotiated or having been executed. This includes the baseline information and the optional specialised elections">
     [metadata key]
     [rootType]
 
     agreementTerms AgreementTerms (0..1) <"Specification of the content of the legal agreement.">
     relatedAgreements LegalAgreement (0..*) <"Specifies the agreement(s) that govern the agreement, either as a reference to such agreements when specified as part of the CDM, or through identification of some of the key terms of those agreements, such as the type of agreement, the publisher, the vintage, the agreement identifier and the agreement date.">
+    amendment AmendmentTerms (0..*) <"Represents the legal purpose of an amendment document and the Legal agreements which are impacted by the amendment">
     umbrellaAgreement UmbrellaAgreement (0..1) <"The determination of whether Umbrella Agreement terms are applicable (True) or Not Applicable (False).">
-
+
```
paulhands, line 73: super minor, unnecessary indent

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +159,27 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+type UmbrellaAgreement: <"Specifies a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement.">
 
-    isApplicable boolean (1..1) <"The determination of whether Umbrella Agreement terms are Applicable (True), or Not Applicable (False)">
-    language string (0..1) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
-    parties UmbrellaAgreementEntity (0..*) <"Underlying principals to the umbrella agreement.">
+    agreementSet UmbrellaAgreementSet (1..*) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
+    parties UmbrellaAgreementEntity (2..*) <"Specifies the underlying entities to the agreement that either act as legal parties to the agreement or represent Legal constructs that are managed by a party to the agreement but do not represent a Legal Entity as of themselves.">
```
paulhands: Aware the `Legal Entity` capitalisation is standard, but gave me pause at `Legal constructs`. Very minor

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,27 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+
+    agreementSet UmbrellaAgreementSet (1..*) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
+    parties UmbrellaAgreementEntity (2..*) <"Specifies the underlying entities to the agreement that either act as legal parties to the agreement or represent Legal constructs that are managed by a party to the agreement but do not represent a Legal Entity as of themselves.">
+    language string (0..1) <"Represents any additional language that may need to be captured regarding the structure or application of terms that may exist for the agreement.">
 
-    isApplicable boolean (1..1) <"The determination of whether Umbrella Agreement terms are Applicable (True), or Not Applicable (False)">
-    language string (0..1) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
-    parties UmbrellaAgreementEntity (0..*) <"Underlying principals to the umbrella agreement.">
+type UmbrellaAgreementEntity extends Counterparty: <"A class to specify the legal entities that are part of the umbrella agreement.">
```
paulhands: Is this description technically true now? We are defining both legal and non legal entities in this type? I believe the parentParty does need to be legal, but the actual entity possibly not?

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,27 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    umbrellaPartyrole UmbrellaPartyRoleEnum (0..1) <"Represents the legal role that each defined entity to the agreement performs within the structure of that agreement.">
+    parentParty Parentparty (0..1) <"Represents the identification of a parent child relationship between two entities in the Umbrella agreement, allowing non-legal entities to be linked to the appropriate legal entity and allowing funds, portfolio or managed accounts to be linked to their applicable Investment Manager or Agent within the agreement.">
```
paulhands: Would rename `Parentparty` to `ParentParty` for syntax

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
paulhands: But is this not always going to be `ParentParty` and hence will be a legal entity? in which case Chris is right, and it would provide a good sanity check that nobody sets up a non legal entity as a parent?

If non legal entities _can_ be parents, please ignore me, but as far as I can see this is sensible

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,27 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
 
-    condition UmbrellaAgreementExists: <"Umbrella Agreement language and parties should not exist when Umbrella Agreement terms are Not Applicable.">
-        if isApplicable = True then language exists and parties exists
+    condition Parentparty: <"Identifies where a parent child relationship exists between two parties to the Umbrella agreement, allowing none legal entities to be linked to the appropriate legal entity and allowing funds, protfolio or managed accounts to be linked to their applicable Investment Manager or Agent within the agreement.">
+        if umbrellaPartyrole = UmbrellaPartyRoleEnum -> Sleeve then parentParty exists
```
paulhands: So a sleeve is the only non legal entity here, sub accounts can stand on their own right?

And in the same breath, a principal or an investment manager can have an agent as a parent?

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,27 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
         if creditSupportAgreementMarginType exists
         then creditSupportAgreementType exists
 
-type UmbrellaAgreement: <"A class to specify a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement. This data representation reflects the ISDA Create representation.">
+type UmbrellaAgreement: <"Specifies a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement.">
+
+    agreementSet UmbrellaAgreementSet (1..*) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
+    parties UmbrellaAgreementEntity (2..*) <"Specifies the underlying entities to the agreement that either act as legal parties to the agreement or represent Legal constructs that are managed by a party to the agreement but do not represent a Legal Entity as of themselves.">
```
paulhands: On line 156 we state this captures the legal entities beyond the two contracting parties. But then we specify here parties must be a minimum of length 2. If I have an umbrella agreement between bank A and bank B, but bank B is acting as an agent for a single principal, would this not fall down? Aware that scenario is unlikely, but it would be technically possible?

Party A Bank A
Party B Bank B (acting as an agent)
Principal Fund 1

Or would the top layer parties also be captured here, in which case the description is slightly off?

Again, legalese not my forte, so point out the flaws in my logic

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: At the moment that is true, although if there are believed to be other entity types these can be added to the enum and the appropriate condition added to ensure compliance to ensure non legal entities have a legal entity parent.

As you state we also need to recognise that any Legal entity can also have a parent but we cannot enforce that a Legal Entity must have a parent if they are not an Agent or Investment Manager.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: I think the description is off and needs to be updated to reflect the scenario you state. A party can be both an Agent on the agreement and a Principal as well.

rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta
AdrianJenksD2LT: This was resolved by removing some code that was part of a later planned submission

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Amended

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Amended capitalisation

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Description amended

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Changed as recommended

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -69,8 +69,8 @@ type LegalAgreement extends LegalAgreementBase: <"The specification of a legal a
 
     agreementTerms AgreementTerms (0..1) <"Specification of the content of the legal agreement.">
     relatedAgreements LegalAgreement (0..*) <"Specifies the agreement(s) that govern the agreement, either as a reference to such agreements when specified as part of the CDM, or through identification of some of the key terms of those agreements, such as the type of agreement, the publisher, the vintage, the agreement identifier and the agreement date.">
-    umbrellaAgreement UmbrellaAgreement (0..1) <"The determination of whether Umbrella Agreement terms are applicable (True) or Not Applicable (False).">
-
+    umbrellaAgreement UmbrellaAgreement (0..1) <"TSpecifies a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement.">
```
Oscar-Twomey: Should this be `(0..*)`, and the number of agreements within the `UmbrellaAgreement` type should `(0..1)`? 

Therefore, you can have multiple `UmbrellaAgreements` within a `LegalAgreement`, but in each of them you only have **one** agreement with multiple parties to that specific agreement. Doing it this way would remove the need for a `UmbrellaAgreementSet` type.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,34 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
         if creditSupportAgreementMarginType exists
         then creditSupportAgreementType exists
 
-type UmbrellaAgreement: <"A class to specify a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement. This data representation reflects the ISDA Create representation.">
+type UmbrellaAgreement: <"Specifies a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement.">
+
+    agreementSet UmbrellaAgreementSet (1..*) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
```
Oscar-Twomey, line 160: Following on from the comment above, I believe this type should be `Agreement` with a cardinality of `(0..1)`. Then you can specify multiple parties, but they are only party to one agreement. 

If you have multiple agreements they can all be added in `LegalAgreement -> UmbrellaAgreement` if it has a cardinality of (0..*)

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,34 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    parties UmbrellaAgreementEntity (2..*) <"Specifies the underlying entities to the agreement that either act as legal parties to the agreement or represent legal constructs that are managed by a party to the agreement but do not represent a Legal Entity as of themselves.">
+    language string (0..1) <"Represents any additional language that may need to be captured regarding the structure or application of terms that may exist for the agreement.">
 
-    isApplicable boolean (1..1) <"The determination of whether Umbrella Agreement terms are Applicable (True), or Not Applicable (False)">
-    language string (0..1) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
-    parties UmbrellaAgreementEntity (0..*) <"Underlying principals to the umbrella agreement.">
+type UmbrellaAgreementEntity extends Counterparty: <"Specifies the entities that are part of the umbrella agreement.">
```
Oscar-Twomey: Instead of extending counterparty, you could add a party attribute, and an attribute to specify the role it is acting _onBehalfOf_. The language used in the `Counterparty` type.

1. The reason this is suggested is to make the language in this type a bit clearer, i.e., you have a party to the agreement and the counterparty they act on behalf of, rather than multiple parties acting under the same party role. 
2. It makes specifying either a `nonLegalEntity` or a `party` more obvious
3. When the new changes to Party/LegalEntity are added to CDM, updating this type will be made easier. We are proposing adding a `NonLegalEntity` type to the new Party structure so eventually, the nonLegalEntity attribute (see comment below) can be removed

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,34 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    umbrellaPartyrole UmbrellaPartyRoleEnum (0..1) <"Represents the legal role that each defined entity to the agreement performs within the structure of that agreement.">
+    nonLegalEntityIdentifier string (0..1) <"Specifies the unique identifier value for a non legal entity such as a sleeve, or sub account, that is a party to the agreement and has a relationship with a legal entity defined on the agreement.">
+    nonLegalEntityIdentifierType string (0..1) <"Specifies the Identifier type for a non Legal entity to the agreement, for example the use of an alert code to uniquely identify a non Legal entity.">
+    parentParty ParentParty (0..1) <"Represents the identification of a parent child relationship between two entities in the Umbrella agreement, allowing non-legal entities to be linked to the appropriate legal entity and allowing funds, portfolio or managed accounts to be linked to their applicable Investment Manager or Agent within the agreement.">
```
Oscar-Twomey: `nonLegalEntityIdentifier`, `nonLegalEntityIdentifierType`, and `parentParty` should be added to a new type called `NonLegalEntity`

```
type NonLegalEntity:
    identifier string (1..1)
    identifierType string (0..1) 
    parentParty Party (1..1)
        [metadata reference]
```

This type will eventually be removed when `NonLegalEntity` is added to Party under the new Party structure. The existing party attribute can then define either the party or nonLegalEntity to the agreement.

The `ParentParty` type can be removed, and all the conditions updated to point to the `NonLegalEntity -> parentParty` attribute.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,34 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+        if umbrellaPartyrole = UmbrellaPartyRoleEnum -> Sleeve then parentParty exists
+        or
+        if umbrellaPartyrole = UmbrellaPartyRoleEnum -> Agent then parentParty is absent
 
-type UmbrellaAgreementEntity extends LegalEntity: <"A class to specify the legal entities that are part of the umbrella agreement.">
+type ParentParty: <"Specifies the hierarchical structure of the entity relationships within the Umbrella agreement.">
+   parentIdentifier LegalEntity (0..1) <"Represents the Entity Identifier of the parent entity on this agreement for the party defined.">
 
-    terms string (0..1) <"The terms that might be associated with each party to the umbrella agreement.">
+type UmbrellaAgreementSet: <"Represents the groups of agreement term elections that exist within the agreement and to which entities that particular group of elections apply. Where all entities under the umbrella have the same set of elections only one agreement set will exist. At least one agreement set must be defined for an Umbrella agreement">
```
Oscar-Twomey: This type can be removed.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,34 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    agreement Agreement (1..1) <"Specification of the standard set of terms that define a legal agreement.">
```
chrisisla, line 183: When specifying the `agreement` (whether it is here in `UmbrellaAgreementSet` or in `UmbrellaAgreement`, depending on the final contribution), we need to ensure that the ISLA GMSLA is excluded as it is not applicable for use within an umbrella agreement. This can be done using a condition like this for example:
```
condition ExcludeGMSLA:
        agreement -> masterAgreementElections -> islaGmsla is absent
```

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
Oscar-Twomey: Agreed to update this to (0..1) and re-include an `UmbrellaAgreementSet`

```
type UmbrellaAgreement: 
    agreementSet UmbrellaAgreementSet (1..*)

File: rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta (hunk 24 above)
Comment (Oscar-Twomey) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta:160]: Agreed to update the `UmbrellaAgreementSet` type to use the following structure


```
type UmbrellaAgreementSet: 
    agreement Agreement (0..1) 
    umbrellaAgreementParty UmbrellaAgreementParty (2..*) 
    language string (0..1)

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
Oscar-Twomey: Agreed to update the `UmbrellaAgreementSet` type to use the following structure


```
type UmbrellaAgreementSet: 
    agreement Agreement (0..1) 
    umbrellaAgreementParty UmbrellaAgreementParty (2..*) 
    language string (0..1)

File: rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta (hunk 23 above)
Comment (AdrianJenksD2LT) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta]: Updated as identified

File: rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta (hunk 24 above)
Comment (AdrianJenksD2LT) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta:160]: Updated as recommended

File: rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta (hunk 28 above)
Comment (AdrianJenksD2LT) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta:183]: Added as requested

File: rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta (hunk 27 above)
Comment (AdrianJenksD2LT) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta]: Updated as requested

File: rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta (hunk 26 above)
Comment (AdrianJenksD2LT) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta]: parentParty has been moved under UmbrellaAgreementParty as it is applicable to any party that is defined under the umbrella agreement and not just to a Non Legal Entity

File: rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta (hunk 25 above)
Comment (AdrianJenksD2LT) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta]: Amended as recommended

**Comment by regnosys-prod-user** [comment]
_What is being released?_

Contribution updated to reflect comments received on the original contribution.

> [!NOTE]

File: rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta (hunk 23 above)
Comment (Oscar-Twomey) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta]: @AdrianJenksD2LT there is a typo above in 'specified'

Comment (AdrianJenksD2LT) [reply rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta]: I have resolved the typo and re contributed

File: rosetta-source/src/main/rosetta/legaldocumentation-common-enum.rosetta [hunk 29]
```diff
@@ -280,3 +280,11 @@ enum TerminationCurrencyConditionEnum:
     FreelyAvailable <"A currency that is freely available.">
     PaymentsDueAndFreelyAvailable <"A currency in which payments would be due under one or more Transactions and that is freely available.">
     Specified <"Termination Currency Conditions are specified.">
+
+enum UmbrellaPartyRoleEnum:
+    Agent <"Represents a Legal Entity who has authority to negotiate, execute and deliver confirmations on behalf of each affiliate that is linked to the Agent.">
```
Oscar-Twomey: Hi @AdrianJenksD2LT are the descriptions of the values here all correct in regard to the LegalEntity / Party distinction?

AdrianJenksD2LT: I have added a clarification to the description

Oscar-Twomey: Can you do the same for all the other enum values if required.

AdrianJenksD2LT: All Enum descriptions updated and contributed

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,36 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
         then creditSupportAgreementType exists
 
-type UmbrellaAgreement: <"A class to specify a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement. This data representation reflects the ISDA Create representation.">
+type UmbrellaAgreement: <"Specifies a set of parties which are part of a legal agreement beyond the two contracting parties to that agreement.">
+    agreementSet UmbrellaAgreementSet (1..*) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
+    
+type UmbrellaAgreementParty: <"Specifies the entities that are part of the umbrella agreement.">
+    party Party (0..1) <"Specifies the party to the umbrella agreement">
+    actingAs CounterpartyRoleEnum (0..1) <"Specifies the CounterpartyRoleEnum, e.g. either Party1 or Party2, that is associated to the party.">
+    umbrellaPartyrole UmbrellaPartyRoleEnum (0..1) <"Represents the legal role that each defined entity to the agreement performs within the structure of that agreement.">
```
chrisisla: umbrellaParty**R**ole as the type name

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,36 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+
+type UmbrellaAgreementSet: <"Represents the groups of agreement term elections that exist within the agreement and to which entities that particular group of elections apply. Where all entities under the umbrella have the same set of elections only one agreement set will exist. At least one agreement set must be defined for an Umbrella agreement">
+    agreement Agreement (1..1) <"Specification of the standard set of terms that define a legal agreement.">
+    umbrellaAgreementParty UmbrellaAgreementParty (2..*) <"Specifies the underlying parties to the agreement that either act as legal parties to the agreement or represent legal constructs that are managed by a party to the agreement but do not represent a Legal Entity as of themselves.">
+    additionalLanguage string (0..1) <"Represents any additional language that may need to be captured regarding the structure or application of terms that may exist for the agreement.">
 
-    terms string (0..1) <"The terms that might be associated with each party to the umbrella agreement.">
+    condition ExcludeGMSLA: <"Umbrella agreement fn=unctionality is not applicable to the GMSLA legal agreement type">
```
chrisisla: typo in functionality

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,36 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    identifier string (1..1) <"Specifies the unique identifier value for a non legal entity such as a sleeve, or sub account, that is a party to the agreement and has a relationship with a legal entity defined on the agreement.">
+    identifierType string (1..1) <"Specifies the Identifier type for a non Legal entity to the agreement, for example the use of an alert code to uniquely identify a non Legal entity.">
+
+type UmbrellaAgreementSet: <"Represents the groups of agreement term elections that exist within the agreement and to which entities that particular group of elections apply. Where all entities under the umbrella have the same set of elections only one agreement set will exist. At least one agreement set must be defined for an Umbrella agreement">
+    agreement Agreement (1..1) <"Specification of the standard set of terms that define a legal agreement.">
+    umbrellaAgreementParty UmbrellaAgreementParty (2..*) <"Specifies the underlying parties to the agreement that either act as legal parties to the agreement or represent legal constructs that are managed by a party to the agreement but do not represent a Legal Entity as of themselves.">
```
Oscar-Twomey: `umbrellaAgreementParty` should be renamed `party`

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,36 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    parentParty Party (0..1) <"Represents the identification of a parent child relationship between two entities in the Umbrella agreement, allowing non-legal entities to be linked to the appropriate legal entity and allowing funds, portfolio or managed accounts to be linked to their applicable Investment Manager or Agent within the agreement.">
+
+    condition ParentParty: <"Specifies that certain defined roles require that a parent party is identified.">
+        if umbrellaPartyrole = UmbrellaPartyRoleEnum -> Sleeve then parentParty exists
 
-    isApplicable boolean (1..1) <"The determination of whether Umbrella Agreement terms are Applicable (True), or Not Applicable (False)">
-    language string (0..1) <"The language associated with the umbrella agreement, and which applies to all the parties to the umbrella agreement.">
-    parties UmbrellaAgreementEntity (0..*) <"Underlying principals to the umbrella agreement.">
+    condition UmbrellaParty: <"Only a Party or a Non Legal entity can be set.">
+        optional choice party,nonLegalEntity
```
Oscar-Twomey: Please change this to `required choice`. A party or nonLegalEntity should always be specified.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
Oscar-Twomey: This should just be partyRole

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,36 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
...
+    party Party (0..1) <"Specifies the party to the umbrella agreement">
+    actingAs CounterpartyRoleEnum (0..1) <"Specifies the CounterpartyRoleEnum, e.g. either Party1 or Party2, that is associated to the party.">
+    umbrellaPartyrole UmbrellaPartyRoleEnum (0..1) <"Represents the legal role that each defined entity to the agreement performs within the structure of that agreement.">
+    nonLegalEntity NonLegalEntity (0..1) <"Specifies a non legal entity that exists in the agreement with a defined relationship to a Legal entity to the agreement.">
+    parentParty Party (0..1) <"Represents the identification of a parent child relationship between two entities in the Umbrella agreement, allowing non-legal entities to be linked to the appropriate legal entity and allowing funds, portfolio or managed accounts to be linked to their applicable Investment Manager or Agent within the agreement.">
+
+    condition ParentParty: <"Specifies that certain defined roles require that a parent party is identified.">
+        if umbrellaPartyrole = UmbrellaPartyRoleEnum -> Sleeve then parentParty exists
```
Oscar-Twomey: @AdrianJenksD2LT  was there logic originally around UmbrellaPartyRoleEnum -> Agent where parentParty didn't exists?

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Amended as requested

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: typo resolved

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Amended as requested

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Amended as requested

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: There was but it was not needed

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -153,18 +153,36 @@ type AgreementName: <"Specifies the agreement name through an agreement type and
         if creditSupportAgreementMarginType exists
         then creditSupportAgreementType exists
 
-type UmbrellaAgreement: <"A class to specify a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement. This data representation reflects the ISDA Create representation.">
+type UmbrellaAgreement: <"Specifies a set of parties which are part of a legal agreement beyond the two contracting parties to that agreement.">
```
chrisisla: This comment isn't quite right, the `UmbrellaAgreement` type defines the set of agreements, the parties are set within `UmbrellaAgreementSet`. Can we just change the comment to reflect this please?

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
```diff
@@ -69,8 +69,8 @@ type LegalAgreement extends LegalAgreementBase: <"The specification of a legal a
 
     agreementTerms AgreementTerms (0..1) <"Specification of the content of the legal agreement.">
     relatedAgreements LegalAgreement (0..*) <"Specifies the agreement(s) that govern the agreement, either as a reference to such agreements when specified as part of the CDM, or through identification of some of the key terms of those agreements, such as the type of agreement, the publisher, the vintage, the agreement identifier and the agreement date.">
-    umbrellaAgreement UmbrellaAgreement (0..1) <"The determination of whether Umbrella Agreement terms are applicable (True) or Not Applicable (False).">
-
+    umbrellaAgreement UmbrellaAgreement (0..1) <"Specifies a set of legal entities which are part of a legal agreement beyond the two contracting parties to that agreement.">
+
```
chrisisla, line 73: What is the relationship between `agreementTerms` and `umbrellaAgreement`. Should it be possible to have both included under one `LegalAgreement`?

chrisisla (changes requested):
Changes made last iteration look good 🙂 I am still not that happy with the fact we have `UmbrellaPartyRoleEnum` referencing something that is not a "party" (the Sleeve) but I think this is about as good as we're going to get!

I have added a couple more comments but also want to ask, should it be "non legal entity" or "non-legal entity"? We actually have both in the comments. My feeling is "non-legal entity" is better but happy either way, if we can just decide on one and make sure we're consistent in its usage please.

Final point is that the Rosetta Workspace has some additional types in it that we need to ensure are not contributed as part of this PR (`ContactInformationElection`, `NoticeInformationElection` etc)

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: Amended as suggested

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: agreementTerms is used to capture the terms of a one way or Bilateral CSA. The umbrellaAgreement is used as the entry point to define that the structure of this agreement being captured is an Umbrella.

AdrianJenksD2LT: When capturing an agreement you should either capture agreementTerms or umbrallaAgreement not both.

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
AdrianJenksD2LT: As stated in an earlier comment I have updated all references in comments to non-legal entity to maintain consistency.

regnosys-prod-user:
_What is being released?_

Updated comments to maintain consistency of use of term non-legal entity and amended the umbrellaAgreement definition.

> [!NOTE]

rosetta-source/src/main/rosetta/legaldocumentation-common-type.rosetta
chrisisla: If we should only have `agreementTerms` OR `umbrellaAgreement` and not both then we’ll need to add a condition under `LegalAgreement` to enforce this. Something like this should work – please feel free to adapt the comment to better describe an umbrella agreement:

```
condition AgreementType: <"Either the Agreement Terms (i.e. a bilateral agreement) or an Umbrella Agreement (i.e. an agreement with multiple parties) should be included, not both.">
        required choice agreementTerms, umbrellaAgreement
```

AdrianJenksD2LT: Condition added as recommended and contribution submitted

regnosys-prod-user:
_What is being released?_

Updated descriptions with full stop to allow contribution and resolve EasyCLA security authorisation issue

> [!NOTE]

PR #4207, opened 2025-11-24
Chris Rayner Observable - Addition of price sub types

regnosys-prod-user:


rosetta-source/src/main/rosetta/observable-asset-enum.rosetta
```diff
@@ -294,6 +294,12 @@ enum CashPriceTypeEnum: <"Provides a list of possible types of cash prices, appl
     Fee <"A generic term for describing a non-scheduled cashflow that can be associated either with the initial contract, with some later corrections to it (e.g. a correction to the day count fraction that has a cashflow impact) or with some lifecycle events. Fees that are specifically associated with termination and partial termination, increase, amendment, and exercise events are qualified accordingly.">
     Discount <"Denotes a discount factor expressed as a decimal, e.g. 0.95.">
 
+enum PriceSubTypeEnum: <"Provides a list of possible sub types of the chosen price type, used in conjunction with the PriceTypeEnum.">
```
lolabeis, line 292: Rather than building a new enum, this should replace the existing CashPriceTypeEnum (otherwise we're duplicating these enum values).

The latter is only used in CashPrice, consider renaming attribute there as well to priceSubType.

chrisisla: Yes that makes sense. I left `CashPriceTypeEnum` in place really just because we have another issue specifically to review `CashPrice`, but we can update this now no problem. I will make the change and commit.

regnosys-prod-user:
_What is being released?_

Updated to remove `CashPriceTypeEnum` as this is only used by the `CashPrice`.
Within `CashPrice` the `cashPriceType` was updated to now be `priceSubType` of type `PriceSubTypeEnum`. The condition that used `cashPriceType` was also updated to reflect this name change.

> [!NOTE]

PR #4222, opened 2025-11-27
[5xx] Georgina Tarres CDM Party Model - Support for Legal Entity Identifier Type 

PayalKhanna:
updated expectations for #4221

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,25 +1,27 @@
-# CDM Party Model - Support for Legal Entity Identifier Type
+# Reference Data Model – Legal Entity Identifier Type Support
 
 _Background_
 
-The LegalEntity structure did not allow specify the identifier type, unlike PartyIdentifier, making it impossible to distinguish the type of ID being expressed.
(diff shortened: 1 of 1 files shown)


rosetta-source/src/main/resources/cdm-sample-files/functions/business-event/execution/execution-credit-default-swap-func-output.json
```diff
@@ -209,8 +209,20 @@
...
                           "value" : "ACOM CO., LTD."
                         },
+                        "entityIdentifier" : [ {
+                          "identifier" : {
+                            "value" : "004CC9",
+                            "meta" : {
+                              "scheme" : "http://www.fpml.org/spec/2003/entity-id-RED-1-0"
+                            }
+                          },
+                          "identifierType" : "REDID",
```
gtarres, line 219: The changes look as expected. The new `entityIdentifier` attribute is mapped while keeping the previous representation mapped too.

Although only `REDID` appears in the samples, we’ve added support for all id types (`LEI`, `MIC`, `BIC`, `CountryCode`). This happens because the test packs include only legal entities with RED IDs.

Oscar-Twomey (approved):
Reviewed, and I approve the model changes

hugohills-regnosys (approved):

PR #3819, opened 2025-06-19
CDM Reference Data - external list code validation with condition typeAliases

manel-martos:
This release improves externally defined code list validation and enabling user-defined customizations.  
Replaces PR #3484 

Initial CDM Issue: #3512 

## New Features & Enhancements

- Updated `BusinessCenterEnum` attributes – Now uses the `FpMLCodingScheme` type alias for better alignment with standardized code lists.
- Default validation implementation – `ValidateFpMLCodingSchemeImpl` is now bound in `CdmRuntimeModule`.
- Introduced custom validation support – Added `CustomCodelistRuntimeModule`, `ValidateCodeListCustomImpl`, and `CodeValidationTest` to demonstrate user-defined validations
- Automated CodeList XML-to-JSON transformation – Configured a new Maven execution in `pom.xml` to run `org.isda.cdm.codelist.CodeListTransformer` automatically during the build process.
- Relocated generated JSON code lists – Moved them into project resources at `src/main/resources/org/isda/codelist` for consistent access.
- Updated test scenarios to remove orphan type `BusinessCenterEnum` references.

PR #4227, opened 2025-12-01
CDM Reference Data - external list code validation with condition typeAliases

SimonCockx:
Copied from https://github.com/finos/common-domain-model/pull/3819/


rosetta-source/src/main/resources/codelist/codelist2cdmjson.xsl
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hugohills-regnosys, line 1: This introduces a top level folder named `org` in `src/main/resources`.  If this folder is for reference data or schemes (or something) it would better to group it under a appropriately named folder

hugohills-regnosys (dismissed):

rosetta-source/src/main/resources/org/isda/codelist/json/account-type-1-1.json
SimonCockx, line 1: Hm, this doesn't look right. These files are duplicated.

PR #4249, opened 2025-12-08
Rosetta Visualisations - Update securities lending Allocation example

chrisisla:
**Background**
Rosetta provides examples of how to perform lifecycle events. The example showing the use of the Allocation process for securities lending trades has been updated.

**Proposal**
The existing Allocation visualisation has been updated as follows:
- The `after` trades are now assigned to the agent lender and the fund, rather than the borrower and the fund
- The `after` trades have been given Unique Transaction Identifiers
- The original Trade Execution that the allocation is being performed upon is no longer closed. This has to remain open as this is the agreement between the borrower and the lender.

**Compatibility**
There are no compatibility issues with the CDM as these files are separate to the core model.

**Release**
CDM V7

**Additional Context**
N/A


chrisisla:
Retracting this PR.

PR #4300, opened 2025-12-17
Chris Rayner Product Model - Support for Uncollateralised trades

regnosys-prod-user:


regnosys-prod-user:
# *Product Model - Support for Uncollateralised trades*

_Background_

In certain scenarios it is possible for securities lending trades to be set up with no collateral defined i.e. where there is an agreement between the borrower and the lender that there will be no cash or non-cash collateral posted as collateral against the trade.

This is commonly seen between internal counterparties, i.e. intra-group trades.

_What is being released?_

A new option is being added to the `CollateralTypeEnum` of "Uncollateralised". 

This allows trades to specifically state that there is no collateral expected to be posted against them.

_Review Directions_

Changes can be reviewed in PR: [#4300](https://github.com/finos/common-domain-model/pull/4300)

> [!NOTE]

Oscar-Twomey (approved):

rosetta-source/src/main/rosetta/product-collateral-enum.rosetta
```diff
@@ -3,10 +3,11 @@ version "${project.version}"
...
 
-enum CollateralTypeEnum: <"Specifies the types of collateral that are accepted by the Lender">
-    Cash <"Security Lending Trades against Cash collateral">
-    NonCash <"Security Lending Trades against NonCash collateral">
-    CashPool <"Security Lending Trades against CashPool collateral">
+enum CollateralTypeEnum: <"Specifies the types of collateral that are accepted by the Lender.">
+    Cash <"Security Lending Trades against Cash collateral.">
+    NonCash <"Security Lending Trades against NonCash collateral.">
+    CashPool <"Security Lending Trades against CashPool collateral.">
+    Uncollateralised <"A trade where the parties have agreed that no collateral is required. This would typically be an internal i.e. intra-group trade.">
```
Oscar-Twomey, line 10: @chrisisla, I just realised - would you need to add conditions to `CollateralProvisions` for when the CollateralTypeEnum = Uncollateralised?

chrisisla: Hi @Oscar-Twomey 

The only attribute in `collateralProvisions` that we might want to put a restrictive condition around would be `substitutionProvision`. As the trade has no collateral on it, as it is uncollateralised, then it would be very unlikely that at a later date you would then substitute collateral onto it - and in fact, if you did that, then the `collateralType` would need to change to reflect the collateral that you have substituted onto it.

Would you like me to add a condition to `collateralProvisions` to make sure `substitutionProvisions` is not set if the `collateralType` is "Uncollateralised"?

Thanks!

Oscar-Twomey: Hi @chrisisla, yes, I think we should add it in. We can remove later if needs be and it would be backward compatible.

PR #4334, opened 2026-01-07
Chris Rayner Product Model - New PrincipalReturn transfer type

regnosys-prod-user:


regnosys-prod-user:
# *Product Model - New _PrincipalReturn_ transfer type*

_Background_

When a securities loan ends, the borrower will need to return the securities to the lender, and the lender will need to return the collateral that they received against that loan to the borrower.

Similarly, during the term of the loan, the borrower may decide that they wish to return a proportion of the shares that they have been lent.

The transfer of these assets between parties needs to be performed using the `TransferInstruction`.

_What is being released?_

A new option _PrincipalReturn_ is being added to the `ScheduledTransferEnum`. This will allow the `TransferInstruction` to differentiate 
between:
 - Settlements, where the securities and collateral are initially transferred between the parties
 - Recalls, where the borrower is mandated by the lender to return some shares
 - Returns, where the borrower is deciding that they want to return some or all of the shares

_Review Directions_

Changes can be reviewed in PR: [#4334](https://github.com/finos/common-domain-model/pull/4334)

> [!NOTE]

rosetta-source/src/main/rosetta/product-common-settlement-enum.rosetta
```diff
@@ -20,6 +20,7 @@ enum ScheduledTransferEnum: <"The qualification of the type of transfers and cas
     NetInterest <"Net interest across payout components. Applicable to products such as interest rate swaps.">
     Performance <"A cash flow corresponding to a performance return.  The settlementOrigin attribute on the Transfer should point to the relevant Payout defining the performance calculation.">
     Principal <"An amount which corresponds to the notional of the contract for various business reasons. This could be associated to a cashflow (e.g. a principal payment) or a transfer (e.g. delivery of an asset).">
+    PrincipalReturn <"An amount that corresponds to a proportion (or all) of the notional on the contract being returned to the origin. This could be associated to a cashflow (e.g. the return of cash to the payer) or a transfer (e.g. return of securities to the receiver).">
```
Oscar-Twomey, line 23: @chrisisla, after discussing with Leo, he suggested PrincipalReturn should be added to the FeeTypeEnum as this enum is for scheduled transfers rather than events. 

The `priceTransfer` and `FeeTypeEnum` could then be redesigned to represent unscheduled transfers. 

What are your thoughts on this?

chrisisla: Hi @Oscar-Twomey 
A return is a transfer of shares/collateral back to the originator. So this is not a `priceTransfer` and I can't see how it would relate to `FeeTypeEnum` either? 
We've got `Principal` as an option which is for the initial transfer of shares/collateral for settlement, and `PrincipalReturn` is the same principle but the opposite direction.
Thanks!

PayalKhanna:
Superseded by https://github.com/finos/common-domain-model/pull/4349

rosetta-source/src/main/rosetta/product-common-settlement-enum.rosetta
Oscar-Twomey: priceTransfer & FeeTypeEnum were originally designed for unschedule transfer but only anticipated the cash/fee case. To capture PrincipalReturn and Recall it would be better placed within: 

```

type TransferExpression: 
    unscheduledTransfer UnscheduledTransfer (0..1)
    scheduledTransfer ScheduledTransfer (0..1)

    condition: 
         one-of

type UnscheduledTransfer: 
    transferType UnscheduledTransferEnum (1..1)
    priceTransfer FeeTypeEnum (0..1) 
   
    condition: 
         one-of
      
Enum UnscheduledTransferEnum:
   Return
   Recall

chrisisla: Ahhhh, okay yes, that makes more sense now!

PR #4342, opened 2026-01-08
Luke Banaszkiewicz Product Model - Inclusion of time in the event instruction and economic terms

regnosys-prod-user:


rosetta-source/src/main/rosetta/event-workflow-type.rosetta
```diff
@@ -76,6 +76,7 @@ type EventInstruction: <"Specifies instructions to create a BusinessEvent.">
     intent EventIntentEnum (0..1) <"The intent attribute is meant to be specified when the event qualification cannot be programmatically inferred from the event features. As a result it is only associated with those primitives that can give way to such ambiguity, the quantityChange being one of those. An example of such is a reduction in the trade notional, which could be interpreted as either a trade correction (unless a maximum period of time post-event is specified as part of the qualification), a partial termination or a portfolio rebalancing in the case of an equity swap. On the other hand, an event such as the exercise is not expected to have an associated intent as there should not be ambiguity.">
     corporateActionIntent CorporateActionTypeEnum (0..1)
     eventDate date (0..1) <"Specifies the date on which the event is taking place. This date is equal to the trade date in the case of a simple execution.  However it can be different from the trade date, for example in the case of a partial termination.">
+    eventTime TimeZone (0..1) <"The time and zone of the date in which the event is taking place.">
```
chrisisla, line 79: Do we also need an `effectiveTime` associated to the `effectiveDate` in the same way that you have added `eventTime` for `eventDate`?

rosetta-source/src/main/rosetta/base-datetime-type.rosetta
```diff
@@ -172,10 +173,28 @@ type RelativeDateOffset extends Offset: <"A class defining a date (referred to a
     adjustedDate date (0..1) <"The date once the adjustment has been performed. (Note that this date may change if the business center holidays change).">
 
 type TimeZone: <"The time alongside with the timezone location information. This class makes use of the FpML TimezoneLocation construct.">
+     [metadata key]
     time time (1..1) <"The observation time.">
     location string (0..1) <"FpML specifies the timezoneLocationScheme by reference to the Time Zone Database (a.k.a. tz database) maintained by IANA, the Internet Assigned Numbers Authority.">
         [metadata scheme]
 
+type DirectOrRelativeTime: <"Represents a time definition that can be either an absolute (direct) time or a time calculated relative to another reference.">
+    directTime TimeZone (0..1)
```
chrisisla: Please add a brief comment next to the attributes you have added in these new types 🙂 In general the attributes are quite self explanatory but it is good practice to add something brief to explain their intended usage.

Also, just on `directTime`, I notice in the comment for the type you have put "absolute (direct) time". Which is a better way to describe this time, absolute or direct? I think direct makes more sense as absolute time has it's own definition which I don't think applies here.

rosetta-source/src/main/rosetta/base-datetime-type.rosetta
```diff
@@ -172,10 +173,28 @@ type RelativeDateOffset extends Offset: <"A class defining a date (referred to a
...
+    timeRelativeTo TimeZone (0..1)
+        [metadata reference]
+
+type RelativeTimeOffset: <"Describes how a time is offset from a reference point, including the unit, magnitude, and rounding behavior.">
+    offsetTimeUnit TimeUnitEnum (0..1)
+    offsetMultiplier int (0..1) <"A time offset multiplier, e.g. 1, 2 or 3 etc. A negative value can be used when specifying a negative offset, e.g. -2 hours.">
+    rounding TimeRounding (0..1)
+
+type TimeRounding: <"Specifies how a calculated time should be rounded, including the direction and the unit to round to.">
+    roundingDirection RoundingDirectionEnum (1..1)
```
chrisisla: If you have to specify a direction then would it not make sense to also specify the unit to round to?

rosetta-source/src/main/rosetta/event-workflow-type.rosetta
lukebanas: we have added effective time below ?

rosetta-source/src/main/rosetta/product-template-type.rosetta
```diff
@@ -30,11 +30,13 @@ type EconomicTerms: <" This class represents the full set of price-forming featu
             provision "As defined in GMRA paragraph 2(mm) The date on which Purchased Securities are sold or are to be sold by Seller to Buyer."]
         [docReference ICMA ERCCBestPractice namingConvention "Purchase Date"
             provision "ERCC Guide: Annex II  Glossary of repo terminology. The term for the value date of a repo."]
+    effectiveTime DirectOrRelativeTime (0..1) <"The time and zone of the first day of the terms of the trade. It can also be specified in relation to another.">
```
lukebanas: @chrisisla here

rosetta-source/src/main/rosetta/base-datetime-type.rosetta
lukebanas: line 196 has the roundToUnit? I will add a comment on this type as per your reply further up

rosetta-source/src/main/rosetta/event-workflow-type.rosetta
chrisisla: The `EventInstruction` also has an `effectiveDate` which is what I was referring to with this comment, rather than the `effectiveDate` under `EconomicTerms` that I can see you have added an `effectiveTime` for already. So do we need an `EventInstruction -> effectiveTime` as well?

rosetta-source/src/main/rosetta/product-template-type.rosetta
chrisisla: Could you add the word "time" to the end of the comment for `effectiveTime` please?

rosetta-source/src/main/rosetta/event-workflow-type.rosetta
lagarciath: I understand that the `EventInstruction` should be the type that describes the event that is going to take place. In this case, I believe the relevant extension is `eventTime`, since it relates directly to the event itself. As for the `effectiveDate` in the `EventInstruction`, I’m not entirely sure in which contexts it is used. However, if the effective time has already been defined in the economic terms, that may be sufficient to clearly specify the time of day when the transaction is effective, whereas the time of day when the event itself is executed should be established here (`EventInstruction`). Does that make sense?

chrisisla: I believe `eventDate` is to specify the actual date that the event is taking place on, whereas the `effectiveDate` is the date that the event is effective from i.e. you can place an event today but it is not effective until the `effectiveDate`. This allows you to enter an event today but it is not effective today, thus allowing you to enter forward or back dated activities. 

The `effectiveDate` in the `economicTerms` is associated to the product/contract itself, whereas the `effectiveDate` in the `EventInstruction` is associated to an event and I believe serves two separate purposes. 

I'm not sure if anyone is using the `EventInstruction -> effectiveDate` (which is a different conversation!) but I would suggest adding an `effectiveTime` here to be consistent and to future proof this change.

lagarciath: It makes sense!

lukebanas: @chrisisla apologies for the delay - should now be resolved... please check

chrisisla (approved):

PR #4335, opened 2026-01-07
Chris Rayner Product Model - New Recall Provisions and Recall transfer type

regnosys-prod-user:

PR #4348, opened 2026-01-14
[7dev] Chris Rayner Product Model - New Recall Provisions and Recall transfer type 

PayalKhanna:
Release PR for #4335


RELEASE.md
eacunaISDA, line 1: Could we perhaps reformulate this phrase: "_or a share goes special_" using a different verb than 'go'? @chrisisla

PayalKhanna: @Oscar-Twomey FYI

PR #4385, opened 2026-01-26
[7Dev]Vineet Kumar Product Model - Modification to Interest Rate and Equity Qualification functions (Updated Expectations)

Chuks-Ajeh:
expectations PR for #4382

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -8,6 +8,24 @@ _What is being released?_
 
 This release includes a modification to the Qualify_AssetClass_InterestRate and Qualify_AssetClass_Equity logic to avoid qualifying FX Products as Interest Rate or Equity.
 
-_Review directions_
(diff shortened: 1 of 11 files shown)


rosetta-source/src/main/resources/functions/repo-and-bond/bond-execution-func-output.json
```diff
@@ -4,10 +4,6 @@
     "primitiveInstruction" : {
       "execution" : {
         "product" : {
-          "taxonomy" : [ {
```
hugohills-regnosys, line 7: These expectations are wrong

hugohills-regnosys: Fixed

PR #4432, opened 2026-02-03
Oscar Twomey Asset Model - Aligning AssetType with Asset and Security

regnosys-prod-user:


CDM-ReleaseManagement-OT:
@tomhealey-icma can you please review?

rosetta-source/src/main/rosetta/base-staticdata-asset-common-enum.rosetta
```diff
@@ -256,14 +256,13 @@ enum CreditRiskEnum: <"Represents an enumeration list to identify tranched or un
...
+enum SecurityTypeEnum: <"Represents an enumeration list to indentify the type of an instrument.">
     Debt <"Identifies an instrument as a fixed income instrument of debt issued and securitized as a tradable asset.">
     Equity <"Identifies an instrument as an Equity value of holding of shares in listed company.">
     Fund <"Identifies an instrument as representing holding in an investment fund.">
     Warrant <"Identifies an instrument as a Warrant that give the right, but not the obligation, to buy or sell a security — most commonly an equity — at a certain price before expiration, or to receive the cash equivalent.">
     Certificate <"Identifies an instrument as one that that offers a derivative-based economic return which is not structured as a bond, an equity or a warrant. Note that this security type is not a Certificate of Deposit (aka CD).">
     LetterOfCredit <"Identifies an instrument as a letter of credit or documentary credit/ bankers commercial credit.  A payment mechanism used in international trade to provide economic guarantee of payment by a creditworthy issuer for payment of exported goods.">
```
chrisisla: I don't think a Letter of Credit can be classified as a type of security... an instrument yes, but not a security.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-enum.rosetta
```diff
@@ -256,14 +256,13 @@ enum CreditRiskEnum: <"Represents an enumeration list to identify tranched or un
     TranchedCreditRisk <"Indicates tranched credit risk, including securitizations.">
     UntranchedCreditRisk <"Indicates tranched credit risk, including repackagings.">
 
-enum InstrumentTypeEnum: <"Represents an enumeration list to indentify the type of an instrument.">
+enum SecurityTypeEnum: <"Represents an enumeration list to indentify the type of an instrument.">
     Debt <"Identifies an instrument as a fixed income instrument of debt issued and securitized as a tradable asset.">
     Equity <"Identifies an instrument as an Equity value of holding of shares in listed company.">
     Fund <"Identifies an instrument as representing holding in an investment fund.">
```
chrisisla: A fund is not a type of security... not sure it should be a type of instrument either. A fund can hold securities, but is not a type of security in its own right, so not sure this should be an option here?

rosetta-source/src/main/rosetta/base-staticdata-asset-common-enum.rosetta
```diff
@@ -256,14 +256,13 @@ enum CreditRiskEnum: <"Represents an enumeration list to identify tranched or un
...
     Warrant <"Identifies an instrument as a Warrant that give the right, but not the obligation, to buy or sell a security — most commonly an equity — at a certain price before expiration, or to receive the cash equivalent.">
     Certificate <"Identifies an instrument as one that that offers a derivative-based economic return which is not structured as a bond, an equity or a warrant. Note that this security type is not a Certificate of Deposit (aka CD).">
```
chrisisla: The description for this implies that this should not be considered a security either

rosetta-source/src/main/rosetta/base-staticdata-asset-common-enum.rosetta
Oscar-Twomey: Fund exists in the `Security` type already. 

From what I am reading online, a fund is considered an investment vehicle. Not sure how this would translate into the product model however, unless a new type was added. Or else fund exists as an Asset type, but not sure if this is correct either.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-enum.rosetta
Oscar-Twomey: In which case, it should be added it to the Instrument choice type as well. 

And in keeping with this PR it should be added to AssetTypeEnum, which contains a flattened list of Instrument.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-enum.rosetta
Oscar-Twomey: Ok. It may be worth discussing on a call tomorrow - the purpose of this PR is to align the product and collateral models with minimal impact to the collateral model. 

The challenge here is that there seems to be different definitions of security and instrument between both models. To align them it would mean adding some new types to the instrument choice type and creating a new security type enum.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-enum.rosetta
Oscar-Twomey: Hi Chris, in line with the contribution, I have moved to `AssetTypeEnum`.

chrisisla (approved):
Putting the non-Security items under `AssetTypeEnum` makes sense as an interim change. I'm still not convinced that "Fund" should be under `SecurityTypeEnum` (I think this could be context though, as if it is an exchange traded fund then it could be an instrument, but still not a security I think...?) but it has been that way for a while - if someone queries it in the future we can revisit it then.

regnosys-prod-user:
_What is being released?_

Adding extra conditions.

> [!NOTE]

PR #4435, opened 2026-02-05
CDM Schema Publication

### LINKED ISSUE #4436: Publish CDM Schemas at a public URL
### Background

FDC3 would like to be able to represent CDM objects : https://github.com/finos/FDC3/issues/1604

### Proposal

Publish CDM Schemas at a public URL.  This will enable us to use them (I think) in the FDC3 project.

@iansloyan 

### Compatibility

[See @](https://github.com/finos/common-domain-model/pull/4435)

### Release

cdm.finos.org website.

### Additional Context

_No response_

robmoffat:
@iansloyan 

Fixed #4436


eacunaISDA:
@robmoffat - I moved your list of SCHEMA_VERSIONS to a centralized place, a separate file so your download script can reference that, and we only update the list in one place. I set up landing pages and navigation to the schema. As for the $id we have several options to include that in advance or later on, ideally it would be part of the schema generator.  

@minesh-s-patel - on your end please consider including $id in the json file to follow a path something like this:
```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "$id": "https://cdm.finos.org/schemas/6.0/cdm-base-datetime-AdjustableDate.schema.json",
  "$anchor": "cdm.base.datetime",
  "type": "object",
  "title": "AdjustableDate",

```
(we may want another folder below 6.0, before file itself), maybe make this change once we've finalized the URL pattern...

Alternatively, we can add the $id as part of the download script but it would be nice to have it also available in Maven central, also:

✅ Universal benefit - Every consumer of CDM schemas gets $id fields automatically
✅ Single source of truth - URL configuration lives with the schema generator
✅ Future-proof - New schemas always have $id from generation
✅ Proper JSON Schema practice - Schemas should have $id at creation time

website/docusaurus.config.js
```diff
@@ -116,6 +117,10 @@ module.exports = {
               label: 'Releases',
               href: 'https://github.com/finos/common-domain-model/releases',
             },
+            {
+              label: 'Schemas',
+              href: 'https://github.com/finos/common-domain-model/schemas',
```
chrisisla, line 121: Is this path correct?

PR #4466, opened 2026-02-18
Oscar Twomey Collateral Model - Addition of `IndexType`

regnosys-prod-user:


regnosys-prod-user:
# *Collateral Model - Addition of `IndexType`*

_Background_

In the collateral model, there is no way of specifying whether a security is a constituent of a particular index without also specifying an asset identifier for that index.

_What is being released?_

Addition of a new type `IndexType`. This allows for the definition of an index within collateral criteria without using the Index type in the Product Model. This means only the value in the `EquityIndexEnum` needs to be specified rather than an `AssetIdentifier`.

_Review Directions_

Changes can be reviewed in PR: [#4466](https://github.com/finos/common-domain-model/pull/4466)

> [!NOTE]

rosetta-source/src/main/rosetta/product-collateral-type.rosetta
```diff
@@ -360,3 +361,6 @@ type DeliveryAmount: <"A class to specify the application of Interest Amount wit
         one-of
 
 type SpecificAsset extends Asset: <"A single, specifically identified Asset chosen from the Asset data type">
+
+type IndexType: <"Specification a specific index.">
+    equityIndex EquityIndexEnum (0..1) <"An enum list for equity indices.">
```
llynhiavu: @Oscar-Twomey at the CRWG this week, wasn't the feedback (Nick Moger) to make it collateral specific?

collateralEquityIndexEnum

Oscar-Twomey: You're correct

Oscar-Twomey: @llynhiavu The change I am making here will only impact the collateral model. So as not to delay this I propose making the name change in a few weeks ahead of the next CRWG. This is because we will need to remove the attribute from EquityIndex which is in the product model and would require notifying the necessary WG.

llynhiavu (approved):

PR #4512, opened 2026-03-09
Chris Rayner Event Model - Transfer Qualifications

regnosys-prod-user:


rosetta-source/src/main/rosetta/event-qualification-func.rosetta
```diff
@@ -96,6 +96,7 @@ func Qualify_Cancellation: <"Qualification of an cancellation event.">
 
 func Qualify_CashAndSecurityTransfer: <"The qualification of a security settlement from the fact that (i) it is composed of a cashTransfer component and a securityTransfer component, and (ii) the cash and security move in the same direction.">
     [qualification BusinessEvent]
+    [deprecated]
```
Oscar-Twomey, line 99: Hi @chrisisla, is there a reason you have not removed the function completely? Both functions are flawed as per your analysis, so I think they should be removed.

rosetta-source/src/main/rosetta/event-qualification-func.rosetta
```diff
@@ -130,6 +131,7 @@ func Qualify_CashTransfer: <"The qualification of a cash transfer from the fact
                 ) all = True
             and beforeTransfers count + transferInstructions count = afterTransfers count
             and transferInstructions -> transferState -> transfer -> quantity -> unit -> currency exists
```
Oscar-Twomey, line 133: I think the line checking for currency can be removed if the asset transferred is cash.

rosetta-source/src/main/rosetta/event-qualification-func.rosetta
```diff
@@ -728,16 +731,20 @@ func Qualify_SecurityTransfer: <"The qualification of a security transfer from t
...
+    
     set is_event:
-        transfer -> asset -> Instrument -> Security exists
-            and transfer -> quantity -> unit -> financialUnit only exists
-
+        (businessEvent -> instruction
+                extract primitiveInstruction -> transfer only exists
+                ) all = True
+            and beforeTransfers count + transferInstructions count = afterTransfers count
+            and transferInstructions -> transferState -> transfer -> quantity -> unit -> financialUnit exists
```
Oscar-Twomey, line 745: Are you supposed to check the `financialUnit` is a Share also?

rosetta-source/src/main/rosetta/event-qualification-func.rosetta
chrisisla: Agreed. I have set as `[deprecated]` as an interim, giving people time to move away from the functions. I am happy to remove the functions if everyone is happy for me to do so.

rosetta-source/src/main/rosetta/event-qualification-func.rosetta
chrisisla: Agreed. The check for the 'asset` should be enough

rosetta-source/src/main/rosetta/event-qualification-func.rosetta
chrisisla: Similar to the Cash Transfer qualification, the check on the `asset` should be enough, so yes, I agree that we don't need to check the `financialUnit`.

Oscar-Twomey: In your issue you say that you need to check the financial unit is a Share - do you need to add this in?

chrisisla:
Now that the issue has been approved at the CRWG I will create a new PR and include the changes suggested.

PR #3666, opened 2025-04-28
Transferable party1/party2 refactor

tomhealey-icma:
This PR addresses  three issues:
1. Define who pays / receives (Party1 / Party2) for a Tradeable Product's embedded payout [#3590](https://github.com/finos/common-domain-model/issues/3590)
2. Add product issuer, [#3567](https://github.com/finos/common-domain-model/issues/3567)
3. AssetType is not included in Product although it's part of CollateralCriteria and collateral EligibilityQuery. [#3559](https://github.com/finos/common-domain-model/issues/3559)

To address these issues the following changes are being made:
1. Add new PartyRole "Issuer"
2. Add productParty Counterparty (0..1) to TransferableProduct
4. Add to AssetBase:
    assetPayer Party (0..1) 
    assetPayerRole PartyRole (0..1) 
    assetAncillaryParty AncillaryParty (0..*)
5. Add new function input tests, bond-execution-with-bond-details.
6. Add AssetType to AssetBase.

PR #4521, opened 2026-03-12
AssetBase Payer and Receiver Refactor

### LINKED ISSUE #3559: Relocate AssetBase
## Background ##

When a new Asset type was created in v6 there were four types of assets created Cash, Commodity, DigitalAsset and Instrument. Cash, Commodity, DigitalAsset all have Asset base as an attribute where as Instrument has three attributes, ListedDerivative, Loan and Security. Each of these types also has the attribute InstrumentBase. Under InstrumentBase there is the attribute AssetBase and the attribute InstrumentType which again as the attribute ListedDerivative. 

So the questions that arise are:
1) What is the meaning of Instrument->ListedDerivative->InstrumentType->ListedDerivative?
2) What is the difference between Instrument->ListedDerivative->InstrumentType->ListedDerivative and Instrument->Security->InstrumentType->ListedDerivative. 
3) Under ListedDerivative there are only three attributes deliveryTerm, optionType and strike. Thus implying LIstedDerivatives are either futures or simple options but missing many other parameters.

## Proposal ##

As a starting point here are two proposals:
1) AssetBase should be move to the AssetType level instead of InstrumentBase.
2) InstrumentTypes should be specific to the type of Instrument so that ListedDerivative, Loan and Security have mutually exclusive type identifies to eliminate ambiguity at the of type of Instrument. 
3) Remove Listed from the name of ListedDerivative as it's possible to have Transferable derivatives not listed. 

## Compatibility ##

This would be a breaking change to v6 and thus would go into v7.

### LINKED ISSUE #3567: Add product issuer
## Background ##

The function used for collateral eligibility checks, CheckIssuerName currently compares an issuer name from collateral eligibility against the Asset's legalEntity->name (string), which is held in the attribute "exchange" implying that this can only be used for exchange names. 

The product model needs a clear and unambiguous attribute for the issuer of a product that includes the LEI. It is also necessary to add other parties associated with the issuer. 

## Proposal ##

It's proposed that a ProductBase type is created that includes ProductParty (Party) and ProductPartyRole (PartyRole) and merge AssetBase into it so attributes identifier and taxonomy are all in the one place, and higher up in the tree. Product would extend ProductBase and thus apply to transferable and non-transferable products.
 

## Compatibility ##

This would go into version 7 as it is a breaking change in version 6.

### LINKED ISSUE #3590: Modify PayerReceiver to support any party as payer or receiver.
### Background

An InterestRatePayout defines the payerReceiver as a counterparty (Party1 or Party2) as defined in TradeableProduct. For trades with settlementPayout this works but when you add the interestRatePayout for Bonds with Product or Instrument->Underlier, the payer/receiver are incorrect and there is not way to link the payer to an Issuer. This is also related to [3567](https://github.com/finos/common-domain-model/issues/3567) that is missing an Issuer PartyRole. This problem also exits for equity dividends.

This has a big implication for anyone using the CDM and wants to access credit risk based on exposures to payers but it's also a trade processing problem.

### Proposal

The proposal is to change PayerReceiver attributes payer and received from CounterpartyEnum to Party/PartyRole. This would allow for settlement payouts to use buyer/seller and bond payouts to use issuer/holder. 

### Compatibility

This is a breaking change in v6 so it would go into v7.

### Additional Context

_No response_

tomhealey-icma:
This PR addresses three issues and replaces PR 3666::

Define who pays / receives (Party1 / Party2) for a Tradeable Product's embedded payout https://github.com/finos/common-domain-model/issues/3590
Add product issuer, https://github.com/finos/common-domain-model/issues/3567
AssetType is not included in Product although it's part of CollateralCriteria and collateral EligibilityQuery. https://github.com/finos/common-domain-model/issues/3559
To address these issues the following changes are being made:

Add new PartyRole "Issuer"
Add productPartyRole CounterpartyRoleEnum (0..1) to TransferableProduct

Add to AssetBase:
party Party (0..1)
partyRole AssetPartyRole (0..1)
ancillaryPartyRole AssetAncillaryPartyRole (0..*)
assetType AssetTypeEnum (1..1)

Remove from AssetBase:
exchange
relatedExchange

Add new function input tests, bond-execution-with-bond-details.

Update  flpm ingest functions.


tomhealey-icma:
All the changes have been submitted and the build is successful. The CVE error is unrelated to my changes. Request cdm maintainer review.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -17,15 +17,33 @@ type AssetBase: <"The base data type to specify common attributes for all Assets
...
+    ancillaryPartyRole AssetAncillaryPartyRole (0..*) <"The role of a party with indirect participation in the asset (i.e. transfer agent).">
+    assetType AssetTypeEnum (1..1) <"Represents a class to allow specification of the asset product type.">
 
-    condition ExchangeListed: <"If Exchange is specified, it must be an exchange-listed Instrument.">
-        if exchange exists then isExchangeListed
 
-    condition RelatedExchange: <"Related Exchange should only be specified if an Exchange is also specified.">
-        if exchange is absent then relatedExchange is absent
+    condition ExchangeRole: <"If Exchange is specified, it must be an exchange-listed Instrument.">
+         if (partyRole -> role = Exchange or (ancillaryPartyRole->role filter item = AssetAncillaryPartyRoleEnum.Exchange then exists))
```
Oscar-Twomey: The second statement can be re-written for clarity

`ancillaryPartyRole -> role any = Exchange`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -17,15 +17,33 @@ type AssetBase: <"The base data type to specify common attributes for all Assets
...
+
+    condition RelatedExchange: <"If isExchangeList is false then RelatedExchange ancillaryPartyRole is absent.">
+         if isExchangeListed is absent then ancillaryPartyRole -> role any <> RelatedExchange and
+             ancillaryPartyRole -> role any <> Exchange
+
+    condition Issuer: <"If the asset type is a security then the role must be issuer.">
+        if assetType = AssetTypeEnum.Security
+            then partyRole -> role = Issuer
+
+    condition LenderBorrower: <"If the asset type is a Loan then the role must be Lender.">
```
Oscar-Twomey: This is missing a clause for the borrower.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -17,15 +17,33 @@ type AssetBase: <"The base data type to specify common attributes for all Assets
...
+        if assetType = AssetTypeEnum.Loan
+            then partyRole -> role =  Lender
+
+    condition NoAncillaryPartyRole: <"Party,partyrole and ancillaryPartyRole are all absent.">
```
Oscar-Twomey: What is this condition trying to do?

It currently enforces no party, partyRole, or ancillaryPartyRole can be set on the type.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -102,7 +120,7 @@ type Cash extends AssetBase: <"An Asset that consists solely of a monetary holdi
         taxonomy is absent
 
     condition NoExchange: <"Cash cannot be a listed Asset.">
-        exchange is absent and isExchangeListed is absent
+        isExchangeListed is absent
```
Oscar-Twomey, line 120: I think the `is absent` is a bit ambiguous, and would be better written as `isExchangeListed = False` .

If `isExchangeListed` was absent it would be fine. However, if `isExchangeListed` was False, the condition would fail as it would not be absent.

rosetta-source/src/main/rosetta/product-template-type.rosetta
```diff
@@ -22,6 +22,7 @@ import cdm.product.qualification.*
 
 type TransferableProduct extends Asset: <"A TransferableProduct is a type of financial product which can be held or transferred, represented as an Asset with the addition of specific EconomicTerms.">
     economicTerms EconomicTerms (1..1) <"The price forming features, including payouts and provisions.">
+    productPartyRole CounterpartyRoleEnum (1..1) <"Overrides the parties on the trade to indentify the party with the obligation to pay dividends or coupons, e.g issuer.">
```
Oscar-Twomey: Typo - indentify

Should be identify.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -17,15 +17,33 @@ type AssetBase: <"The base data type to specify common attributes for all Assets
...
-    condition ExchangeListed: <"If Exchange is specified, it must be an exchange-listed Instrument.">
-        if exchange exists then isExchangeListed
 
-    condition RelatedExchange: <"Related Exchange should only be specified if an Exchange is also specified.">
-        if exchange is absent then relatedExchange is absent
+    condition ExchangeRole: <"If Exchange is specified, it must be an exchange-listed Instrument.">
+         if (partyRole -> role = Exchange or (ancillaryPartyRole->role filter item = AssetAncillaryPartyRoleEnum.Exchange then exists))
+                  then isExchangeListed = True
+
+    condition ExchangeListed: <"if isExchangeListed is true then partyRole is Exchange.">
```
Oscar-Twomey, line 30: In this condition you're saying:
`if exchangeListed is True, then Exchange must exists as an ancillaryPartyRole`

However, in the `ExchangeRole` condition above, because of the OR, you're allowing for exchangeListed to be True if the PartyRole is Exchange, even if ancillaryPartyRole does not contain Exchange. So both conditions are conflicting in a sense.

<br><br>


Furthermore, in the `RelatedExchange` condition below... 
`if isExchangeListed is absent then ancillaryPartyRole must not = RelatedExchange & Exchange` 

RelatedExchange is not mentioned in the ExchangeListed condition, but given the way it is written, it suggests that RelatedExchange does determine whether an Asset is exchange listed.

Oscar-Twomey (approved):

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
```diff
@@ -776,6 +776,59 @@ func MapExchangeIdToLegalEntity:
...
+    Party {
+        partyId: fpmlExchangeId extract MapPartyIdentifier2,
+        name: MapStringWithScheme(
+                    fpmlExchangeId -> value,
+                    fpmlExchangeId -> exchangeIdScheme
+                ),
+        ...
+    }
+
+func MapPartyIdentifier2:
```
Oscar-Twomey: It would be better to rework the existing `MapPartyIdentifier` function to support ExchangeId rather than duplicating the function.

```
func MapPartyIdentifier:
    inputs:
        value fpml.NonEmptyScheme (1..1)
        partyIdScheme fpml.NonEmptyURI (1..1)
    output:
        partyIdentifier PartyIdentifier (0..1)

    set partyIdentifier:
        PartyIdentifier {
            identifier: MapStringWithScheme(
                        value,
                        partyIdScheme
                    ),
            identifierType: MapPartyIdentifierTypeEnum(partyIdScheme)
        }

File: rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta [hunk 8]
```diff
@@ -776,6 +776,59 @@ func MapExchangeIdToLegalEntity:
...
+ func MapExchangeIdToAssetParty:
+     inputs:
+         fpmlExchangeId fpml.ExchangeId (0..1)
+     output:
+         party Party (0..1)
+
+set party:
+    Party {
+        partyId: fpmlExchangeId extract MapPartyIdentifier2,
+        name: MapStringWithScheme(
```
Oscar-Twomey: Both the name and the partyId will be set with the same values. Only the partyId should be mapped in this case if there is no definitive exchange name being passed in.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
```diff
@@ -776,6 +776,59 @@ func MapExchangeIdToLegalEntity:
             ...
         }
 
+ func MapExchangeIdToAssetParty:
```
Oscar-Twomey: The function should be renamed `MapExchangeIdToParty`

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-pricequantity-func.rosetta
```diff
@@ -1868,6 +1868,12 @@ func MapIdentifiedAssetToAssetListedDerivative:
     output:
         asset Asset (0..1)
 
+    alias assetParty:
+          fpmlExchangeId extract MapExchangeIdToAssetParty
+
+    alias relatedAssetParty:
```
Oscar-Twomey: This alias `relatedAssetParty` is not used in any of the logic and can be removed.

Oscar-Twomey: And the same for the functions below

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-pricequantity-func.rosetta
```diff
@@ -1877,9 +1883,12 @@ func MapIdentifiedAssetToAssetListedDerivative:
                             identifier: MapAssetIdentifierList(fpmlIdentifiedAsset),
                             isExchangeListed: if fpmlExchangeId exists
                                     then True,
-                            exchange: fpmlExchangeId extract MapExchangeIdToLegalEntity,
-                            relatedExchange: fpmlRelatedExchangeId
-                                    extract MapExchangeIdToLegalEntity,
+                            party: assetParty,
+                            partyRole: assetParty extract MapAssetPartyRole,
+                            ancillaryPartyRole: fpmlRelatedExchangeId
```
Oscar-Twomey, line 1941: This gives a syntax warning, and can be re-written a bit clearer as well:

```
  ancillaryPartyRole: fpmlRelatedExchangeId
          filter value <> fpmlExchangeId -> value
          then extract
              MapAssetAncillaryPartyRole(
                      MapExchangeIdToAssetParty
                  ),
```

There is no need for the square brackets, and the item -> value is also unnecessary as there is a distinction between fpmlRelatedExchangeId & fpmlExchangeId you are comparing.

You only need to use extract once, and then the functions can be wrapped.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey, line 16: You will also need some new conditions here to enforce the asset type & party role. E.g.,

```
condition PartyRole: <"If party role is lender or borrower, the asset type must be a loan.">
        if [AssetPartyRoleEnum -> Lender, AssetPartyRoleEnum -> Borrower] any = partyRole -> role then assetType = Loan

File: rosetta-source/src/main/resources/functions/repo-and-bond/bond-execution-with-bond-details-func-input.json
Comment (Oscar-Twomey) [inline rosetta-source/src/main/resources/functions/repo-and-bond/bond-execution-with-bond-details-func-input.json:1]: @tomhealey-icma did you add these repo and bond samples?

File: rosetta-source/src/main/resources/ingest/config/test-pack-translate-fpml-confirmation-to-trade-state-fpml-5-10-incomplete-products-bond-options.json [hunk 12]
```diff
@@ -10,7 +10,7 @@
     "assertions" : {
       "inputPathCount" : 65,
       "outputPathCount" : 72,
-      "modelValidationFailures" : 4,
+      "modelValidationFailures" : 6,
```
Oscar-Twomey: @tomhealey-icma can you please leave a comment on these validation diffs.

rosetta-source/src/main/resources/base-staticdata-asset-commodity-enum.rosetta
Oscar-Twomey, line 1: @PayalKhanna what are these changes?

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: We've lacked a clear definition of what Listed means. I think we can limit the conditions to:
1) isExchangeListed = true, then party must be exchange but ancillaryPartyRole exchange not required
2) If partyRole = Exchange then isExchangeListed = true. 

Related exchange is defined as other than the primary so it's not a condition that should determine isExchangeListed.

Also removing the ancillary party role condition that said you can't have an ancillary party without a party.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: Agree.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: Removing this condition. To restrictive to say you can't have an ancillary party without a party since their role and purpose are different. The new conditions are:
    condition ExchangeRole: <"If Exchange is specified, it must be an exchange-listed Instrument.">
         if partyRole -> role any = Exchange
                  then isExchangeListed = True

         if isExchangeListed then partyRole -> role any = Exchange

    condition Issuer: <"If the asset type is a security then the role must be issuer.">
        if assetType = AssetTypeEnum.Security
            then partyRole -> role any = Issuer

    condition LenderBorrower: <"If the asset type is a Loan then the role must be Lender.">
        if assetType = AssetTypeEnum.Loan
            then partyRole -> role any =  Lender or partyRole -> Borrower

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: changing to :
    condition NoExchange: <"Cash cannot be a listed Asset.">
        isExchangeListed is absent

rosetta-source/src/main/rosetta/product-template-type.rosetta
tomhealey-icma: done

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: Changing conditions to:
    condition ExchangeRole: <"If Exchange is specified, it must be an exchange-listed Instrument.">
         if partyRole -> role any = Exchange
                  then isExchangeListed = True

         if isExchangeListed then partyRole -> role any = Exchange

    condition Issuer: <"If the asset type is a security then the role must be issuer.">
        if assetType = AssetTypeEnum.Security
            then partyRole -> role any = Issuer

    condition LenderBorrower: <"If the asset type is a Loan then the role must be Lender.">
        if assetType = AssetTypeEnum.Loan
            then partyRole -> role any =  Lender or partyRole -> Borrower

rosetta-source/src/main/resources/base-staticdata-asset-commodity-enum.rosetta
Oscar-Twomey: Maybe @tomhealey-icma you could also review? It looks like there have been new files added?

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -17,15 +17,33 @@ type AssetBase: <"The base data type to specify common attributes for all Assets
...
+         if isExchangeListed then ancillaryPartyRole -> role any = Exchange
+
+    condition RelatedExchange: <"If isExchangeList is false then RelatedExchange ancillaryPartyRole is absent.">
+         if isExchangeListed is absent then ancillaryPartyRole -> role any <> RelatedExchange and
+             ancillaryPartyRole -> role any <> Exchange
+
+    condition Issuer: <"If the asset type is a security then the role must be issuer.">
+        if assetType = AssetTypeEnum.Security
```
chrisisla: you don't need to specify the enum anymore so this is clearer as `if assetType = Security`.

For a Security will this condition enforce that we have a `partyRole` set to "Issuer"? If so then this is not valid as the setting of the Issuer should be optional.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
chrisisla: for the last line should that also be "any" for the borrower too? i.e. 
`then partyRole -> role any =  Lender or partyRole -> role any = Borrower`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
chrisisla: for cash `isExchangedListed` makes no sense so it would be better if it was `absent`. If the condition checked for both being `absent` or being `False` that might be the compromise?

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
```diff
@@ -125,6 +125,18 @@ enum PartyRoleEnum: <"The enumerated values for the party role. The enumerated v
     TripartyAgent <"A third party entity that acts as an intermediary and provides automated clearing, settlement, allocation, valuation and reporting services.">
     ThirdPartyCustodian <"A third-party providing custody, settlement, segregation and reporting services.">
 
+enum AssetPartyRoleEnum: <"The enumerated values that specify the role of party with an direct obligation to the asset (i.e. Lender, Issuer).">
```
chrisisla: "...enumerated values that specify the role of **a** party with **a** direct obligation..."

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
```diff
@@ -125,6 +125,18 @@ enum PartyRoleEnum: <"The enumerated values for the party role. The enumerated v
...
+    Exchange <"The exchange or platform where the security has been issued and the exchange has a direct economic role.">
+    Borrower <"The entity that receives cash or assets from a lender with the expectation that cash or assets will be returned at a future agreed date and price.">
+    Issuer <"The entity that issues a financial product and has the obligation to pay coupons, dividends and principal to the product holder.">
+    Lender <"The entity that provides cash or assets to a borrower with the expectation that cash or assets will be returned at a future agreed date and price.">
+
+enum AssetAncillaryPartyRoleEnum: <"The enumerated values that specify the role of a party with an indirect obligation related to an asset as agent, registrar or other.">
+    Registrar <"The party responsible to maintaining records of asset owners and holders.">
+    TransferAgent <"The party responsible to maintaining records of owners and processing changes in ownership.">
+    Exchange <"The exchange or platform where the security has been issued and the exchange has a indirect role.">
```
chrisisla: "...has a**n** indirect role.">

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
```diff
@@ -125,6 +125,18 @@ enum PartyRoleEnum: <"The enumerated values for the party role. The enumerated v
...
+    RelatedExchange <"An exchange or platform other than the primary exchange where the security has been issued and the exchange has a indirect role.">
```
chrisisla: "has a**n** indirect role.">

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
```diff
@@ -125,6 +125,18 @@ enum PartyRoleEnum: <"The enumerated values for the party role. The enumerated v
...
     ThirdPartyCustodian <"A third-party providing custody, settlement, segregation and reporting services.">
 
+enum AssetPartyRoleEnum: <"The enumerated values that specify the role of party with an direct obligation to the asset (i.e. Lender, Issuer).">
+    Exchange <"The exchange or platform where the security has been issued and the exchange has a direct economic role.">
+    Borrower <"The entity that receives cash or assets from a lender with the expectation that cash or assets will be returned at a future agreed date and price.">
+    Issuer <"The entity that issues a financial product and has the obligation to pay coupons, dividends and principal to the product holder.">
+    Lender <"The entity that provides cash or assets to a borrower with the expectation that cash or assets will be returned at a future agreed date and price.">
+
+enum AssetAncillaryPartyRoleEnum: <"The enumerated values that specify the role of a party with an indirect obligation related to an asset as agent, registrar or other.">
+    Registrar <"The party responsible to maintaining records of asset owners and holders.">
```
chrisisla: "responsible **for** maintaining"

rosetta-source/src/main/rosetta/product-template-type.rosetta
chrisisla: Should `productPartyRole` not be optional? If the issuer pays the dividend to the borrower and the borrower then passes that back to the lender then there is no need to specify the issuer on the trade.
Also, taking a securities lending trade as an example, the counterparties on the trade will be the lender and the borrower i.e. will be Party1 and Party2. So is it correct that you are using `CounterpartyRoleEnum` here if you want to specify the issuer too?

rosetta-source/src/main/resources/functions/repo-and-bond/bond-execution-with-bond-details-func-input.json
tomhealey-icma: yes

rosetta-source/src/main/resources/ingest/config/test-pack-translate-fpml-confirmation-to-trade-state-fpml-5-10-incomplete-products-bond-options.json
tomhealey-icma: These must be auto generated.

chrisisla (changes requested):
A few typos and a couple of questions, particularly on the condition that seems to enforce the addition of an issuer to a Security

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: Yes all securities must have an issuer.

chrisisla: Agreed that every security will have an issuer but I don't think it should be mandatory to always have to provide it.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: Here's a new set of conditions:
    condition ExchangeRole: <"If Exchange is specified, it must be an exchange-listed Instrument.">
         if partyRole -> role any = Exchange
                  then isExchangeListed = True

         if isExchangeListed then partyRole -> role any = Exchange

    condition Issuer: <"If the asset type is a security then the role must be issuer.">
        if assetType = AssetTypeEnum.Security
            then partyRole -> role any = Issuer

    condition LenderBorrower: <"If the asset type is a Loan then the role must be Lender.">
        if assetType = AssetTypeEnum.Loan
            then partyRole -> role any =  Lender or partyRole -> Borrower

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: fixed

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: Why? This is specifically for mapping to asset party, no other party.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: This was done for alignment with existing practice.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-pricequantity-func.rosetta
tomhealey-icma: Done

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
tomhealey-icma: done

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
tomhealey-icma: done

tomhealey-icma:
@chrisisla 
Should productPartyRole not be optional? If the issuer pays the dividend to the borrower and the borrower then passes that back to the lender then there is no need to specify the issuer on the trade.
Also, taking a securities lending trade as an example, the counterparties on the trade will be the lender and the borrower i.e. will be Party1 and Party2. So is it correct that you are using CounterpartyRoleEnum here if you want to specify the issuer too?

Whether the asset is prurchased standalone or used in a security lending trade, the issuer that has the obligation to pay coupons and redemptions doesn't change. The asset borrow or lender is not the same as a sec lending trade. So in the end, like a repo trade there are three parties, buyer, seller and asset issuer.

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
tomhealey-icma: done

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
tomhealey-icma: changed condition to:
    condition NoExchange: <"Cash cannot be a listed Asset.">
        isExchangeListed is absent

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: and what about ?
    condition LenderBorrower: <"If the asset type is a Loan then the role must be Lender.">
        if assetType = AssetTypeEnum.Loan
            then partyRole -> role any =  Lender or partyRole -> role any = Borrower

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: This is open to discussion as to whether the conditions should be required.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: Have you tried what you are proposing? It doesn't build because of input mismatch.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: I am not seeing the fix

tomhealey-icma: Added:
    condition LenderBorrower: <"If the asset type is a Loan then the role must be Lender.">
        if assetType = AssetTypeEnum.Loan
            then partyRole -> role any =  Lender or partyRole -> role any = Borrower

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: Done: 
    condition LenderBorrower: <"If the asset type is a Loan then the role must be Lender.">
        if assetType = AssetTypeEnum.Loan
            then partyRole -> role any =  Lender or partyRole -> role any = Borrower

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
tomhealey-icma: done

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: The `any` is not required here because both `partyRole` and `role` are single cardinality.

Did you mean to make `partyRole` single cardinality?

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: Specifying the enum is not required, so it could be written as `  if assetType = Loan`

As above, the `any` is not required anymore. It can either be removed, or the cardinality needs to be updated. I remember `partyRole` being multi-cardinality previously?

Image

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
Oscar-Twomey: The output is a `Party` so its specifically for mapping to a party. 

Its currently only used to map an asset -> party, but its still just a `Party` type.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
Oscar-Twomey: I haven't seen this done - usually there is a name in the type as well which gets mapped to the name. 

Image

<br>

Image

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
Oscar-Twomey: Yes, you would only need to update the inputs. 

Image

<br>

Image

<br>

Alternatively, you could update the function name to something like `MapExchangeIdToPartyIdentifier`

pom.xml
Oscar-Twomey, line 1: @tomhealey-icma do you know what this is?

test-output2.txt
Oscar-Twomey, line 1: @tomhealey-icma can this be deleted?

rosetta-source/src/main/resources/ingest/config/test-pack-translate-fpml-confirmation-to-trade-state-fpml-5-10-incomplete-products-bond-options.json
Oscar-Twomey: Yes, but do you know which validation failures they are - are they the ones you added? 

Or have these changes inadvertently caused validation failures?

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
```diff
@@ -776,6 +777,59 @@ func MapExchangeIdToLegalEntity:
...
+        partyId PartyIdentifier (1..1)
+
+    set partyId -> identifier:
+        MapStringWithScheme(
+            fpmlExchangeId -> value,
+            fpmlExchangeId -> exchangeIdScheme
+        )
+
+
+func MapAssetPartyRole:
```
Oscar-Twomey, line 971: It would be better practice to pass the asset party role as an argument to this function, that way it can be reused in each case.

Same for `MapAssetAncillaryPartyRole` below.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: Agree. Done

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: Removed any

test-output2.txt
tomhealey-icma: Deleted.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: I prefer to change the func name to MapExchangeIdToPartyIdentifier.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: This is the original design so I'm not going to change that now:
                  "partyRole" : {
                    "partyReference" : {
                      "value" : {
                        "partyId" : [ {
                          "identifier" : {
                            "value" : "XNAS",
                            "meta" : {
                              "scheme" : "http://www.fpml.org/schemes/4-1/exchangeId"
                            }
                          },
                          "meta" : {
                            "globalKey" : "292da8"
                          }
                        } ],
                        "name" : {
                          "value" : "XNAS",
                          "meta" : {
                            "scheme" : "http://www.fpml.org/schemes/4-1/exchangeId"
                          }
                        },
                        "meta" : {
                          "globalKey" : "9abd9150"
                        }
                      }
                    },
                    "role" : "Exchange"
                  },
                  "assetType" : "Security",
                  "securityType" : "Equity"
                }
              }

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: While the output is a party the input is specifically an ExchangeId, fpmlExchangeId fpml.ExchangeId (0..1), it intended as a party reference to assets. If you were to use this as a general party reference mapping there is no Exchange in PartyRole. I would mostly prefer to change the name to MapExchangeIdToExchangeParty.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-pricequantity-func.rosetta
tomhealey-icma: Done. Removed brackets, removed  item->value and replaced with function call with two parameters.

pom.xml
tomhealey-icma: This was a custom heap size probable added by Intellij. I'll remove it.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: done

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: Maybe worth discussing on next week's CRWG.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: Tying this to my other comment as I still don't see the removal of `any`

> The `any` is not required here because both `partyRole` and `role` are single cardinality.
> Did you mean to make `partyRole` single cardinality?

Oscar-Twomey: The any can be removed on all of the conditions: 
- `ExchangeRole`
- `ExchangeListed`
- `Issuer`
- `LenderBorrower` (done in [this comment](https://github.com/finos/common-domain-model/pull/4521/changes#r3173091238))

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: - `is absent` checks whether it is empty
- `= False` checks whether it is equal to False

It would be best written as: 

`if isExchangeListed exists then isExchangeListed = False`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -16,16 +16,27 @@ choice Asset: <"An Asset is defined as something that can be owned and transferr
...
     taxonomy Taxonomy (0..*) <"Defines the taxonomy of an object by combining a taxonomy source (i.e. the rules to classify the object) and a value (i.e. the output of those rules on the object.">
-    isExchangeListed boolean (0..1) <"Defines whether the Asset is listed on a public exchange.">
-    exchange LegalEntity (0..1) <"If the Asset is listed, defines the public exchange of the listing.">
-    relatedExchange LegalEntity (0..*) <"Provides the related Exchanges, if applicable.">
-    assetType AssetTypeEnum (1..1) <"Specifies the type of asset.">
+    isExchangeListed boolean (0..1) <"Defines whether the asset is listed on a public exchange.">
+    party Party (0..*) <"The party with the obligation to pay dividends, coupons, or other payouts to the asset holder.">
+    partyRole AssetPartyRole (0..1) <"The role of the payer specific to the asset type, i.e. bond, equity.">
+    ancillaryPartyRole AssetAncillaryPartyRole (0..*) <"The role of a party with indirect participation in the asset (i.e. transfer agent).">
+    assetType AssetTypeEnum (1..1) <"Represents a class to allow specification of the asset product type.">
```
Oscar-Twomey: The description for this should remain as is - its not clear what is meant by asset product type. 

Also, it should be "Specifies the...". There is a move away from the use of class as it is not a Rune concept.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-pricequantity-func.rosetta
Oscar-Twomey: Specifying the enum can also be removed from the logic.

image

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
Oscar-Twomey: Its a bit of an unnecessary function - its only usage is in `MapExchangeIdToAssetParty`. It would make more sense to reuse `MapPartyIdentifier`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: Also, there is still an `AssetTypeEnum.Security` on the issuer condition. This can simply be `Security`

image

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
Oscar-Twomey: In the samples I'm seeing both the ID and name are now being set where it was only 1 previously

image

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: Does it make sense to add for `assetType = ListedDerivative then partyRole must be Exchange`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: Is it worth adding a condition saying for Cash, partyRole is absent?

Oscar-Twomey: As per [this comment](https://github.com/finos/common-domain-model/pull/3666#discussion_r2314301816)

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
Oscar-Twomey: `borrower` can be removed as an attribute from Loan as it is now being specified in `AssetBase`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: I don't see a reason to add a condition that has no conflict. Why can't a 'cash' asset cannot have an issuer?

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: The example I provided was how it was previously which is why I replicated it.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: added

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: done

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-pricequantity-func.rosetta
tomhealey-icma: Done

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: done

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
tomhealey-icma: Done

rosetta-source/src/main/rosetta/product-template-type.rosetta
tomhealey-icma: A pass through coupon or dividend doesn't change who has the obligation on the asset.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: done.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -16,16 +16,31 @@ choice Asset: <"An Asset is defined as something that can be owned and transferr
...
-    condition RelatedExchange: <"Related Exchange should only be specified if an Exchange is also specified.">
-        if exchange is absent then relatedExchange is absent
+    condition ExchangeRole: <"If Exchange is specified, it must be an exchange-listed Instrument.">
+         if partyRole -> role = Exchange
+                  then isExchangeListed = True
+
+    condition ExchangeListed: <"if isExchangeListed is true then partyRole is Exchange.">
+         if isExchangeListed then partyRole -> role = Exchange
+
+    condition Issuer: <"If the asset type is a security then the role must be issuer.">
```
chrisisla: `partyRole` is optional. I'm not sure if this condition will make it mandatory for anything set as `assetType` "Security" and must be set to "Issuer". As we discussed before, I agree all Securities will have an Issuer, I don't think this should be mandatory in `AssetBase`

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
```diff
@@ -16,16 +16,31 @@ choice Asset: <"An Asset is defined as something that can be owned and transferr
...
+        if assetType = Security
+            then partyRole -> role = Issuer
+
+    condition Issuer: <"If the asset type is a security then the role must be issuer.">
```
chrisisla: Think this condition name should change (as we now have two conditions with the same name) and the description in the comment does not describe the purpose of the condition

rosetta-source/src/main/rosetta/product-template-type.rosetta
```diff
@@ -22,6 +22,7 @@ import cdm.product.qualification.*
 
 type TransferableProduct extends Asset: <"A TransferableProduct is a type of financial product which can be held or transferred, represented as an Asset with the addition of specific EconomicTerms.">
     economicTerms EconomicTerms (1..1) <"The price forming features, including payouts and provisions.">
+    productPartyRole CounterpartyRoleEnum (1..1) <"Overrides the parties on the trade to identify the party with the obligation to pay dividends or coupons, e.g issuer.">
```
chrisisla, line 25: Think I queried this before as well, not convinced you can use `CounterpartyRoleEnum` here. If you are describing the collateral on a trade, which is a `TransferableProduct`, then using "Party1" or "Party2" is ambiguous here, especially if you are in fact referring to the Issuer who is not a counterparty to the trade?

chrisisla (changes requested):
I think I raised several of these before but I can't see my original comments 🙁

rosetta-source/src/main/resources/base-staticdata-asset-commodity-enum.rosetta
Oscar-Twomey: Hi, @tomhealey-icma just following up on these files. They look new, not sure if you added them?

tests/pom.xml
```diff
@@ -151,6 +151,29 @@
                         </execution>
                     </executions>
                 </plugin>
+                <plugin>
```
Oscar-Twomey, line 154: @tomhealey-icma is this change to the pom file required?

rosetta-source/src/main/resources/functions/business-event/quantity-change/full-termination-equity-swap-func-output.json
```diff
@@ -30,38 +30,40 @@
...
-                  }
-                }
-              },
-              "settlementDate" : {
-                "adjustedDate" : {
-                  "value" : "2001-09-24",
-                  "meta" : {
-                    "globalKey" : "3e8a58"
-                  }
+            "quantity" : {
```
Oscar-Twomey: @tomhealey-icma it looks like this sample is using an earlier version of CDM. Unscheduled/Scheduled Transfer was added to replace Transfer Expression in [7.0.0-dev.104](https://github.com/finos/common-domain-model/releases/tag/7.0.0-dev.104)

rosetta-source/src/main/resources/ingest/output/fpml-confirmation-to-trade-state/fpml-5-10-incomplete-products-commodity-derivatives/com-ex11-physical-oil-pipeline-heating-oil-fixed-price.json
```diff
@@ -182,16 +182,15 @@
             "scheme" : "http://www.fpml.org/coding-scheme/external/iso17442"
           }
         },
-        "identifierType" : "LEI",
```
Oscar-Twomey, line 161: It looks like this mapping has been lost?

chrisisla:
Emailed @tomhealey-icma showing where I see the duplicate conditions

rosetta-source/src/main/rosetta/product-template-type.rosetta
tomhealey-icma: It was agreed at CRWG meeting, May 12 2026, to change CounterpartyRoleEnum to a new name in a future PR.

tomhealey-icma:
All changes complete and all tests pass. Note: an unrelated upstream change to updateClearingFuncInputJson created duplicate key, clearing-svc" that is cause an error in CdmTestPackCreator.

chrisisla (dismissed):
Thanks for making the changes Tom, I'm happy with all the model changes 🙂 

The one thing that we do still need to resolve before we can merge this is the introduction of the new rosetta files directly under rosetta-source/src/main/resources. I don't think these should be in here and have been added by accident? Other than that I would say this change is good to go.

tomhealey-icma:
> Thanks for making the changes Tom, I'm happy with all the model changes 🙂
> 
> The one thing that we do still need to resolve before we can merge this is the introduction of the new rosetta files directly under rosetta-source/src/main/resources. I don't think these should be in here and have been added by accident? Other than that I would say this change is good to go.

The extraneous files have been removed.

chrisisla (dismissed):
Thanks for removing the spurious rosetta files @tomhealey-icma !

tests/pom.xml
Oscar-Twomey: @tomhealey-icma @chrisisla can you please clarify whether this was intended?

tomhealey-icma: This is the result of a merge with upstream changes to the pom that I did not make. I've confirm that the dev 106 pom and my pom are identical.

Oscar-Twomey (approved):

chrisisla (approved):

PR #4530, opened 2026-03-16
Introduce `thin` Maven Profile for Minimal Runtime Distribution (7.x)

### LINKED ISSUE #4523: Introduce `thin` Maven Profile for Minimal Runtime Distribution
### Background

The current `cdm-java` build process bundles supplemental resources including model samples, mapping analytics, and build-specific configurations into the primary runtime artifact. While useful for development, this "bloated" JAR (observed at ~65 MB in distribution) increases the memory footprint and expands the security surface area for production, cloud-native deployments.

Aligning with industry best practices for Thin JARs, we should decouple core generated logic from these non-runtime assets to improve downstream consumer efficiency and supply chain security.

### Proposal

ntroduce an opt-in Maven profile designed to produce a secondary, "hardened" artifact. To ensure the default build remains untouched, this profile will generate an additional JAR using a Maven Classifier.

#### Naming Alternatives (For TAWG Discussion)

* **`thin`** (Proposed Default): Standard terminology for code-only artifacts.
* **`slim`**: Common in containerization/Docker contexts.
* **`core`**: Differentiates the heart of the model from its extensions.
* **`runtime`**: Explicitly defines the artifact's purpose for production.

### Compatibility

This is fully backward-compatible.

* The default `mvn install` or `mvn deploy` will continue to produce the standard artifact.
* Consumers opting into the `thin` profile will see a significant reduction in size (est. 30–40%) and a simplified security/compliance audit trail.

### Release

Next available 6.x and 7.x milestones.

### Additional Context

_No response_

manel-martos:
# Build Modernization — Introduce `thin` Maven Profile for Minimal Runtime Distribution

_Background_

The current `cdm-java` (rosetta-source) distribution bundles a significant volume of supplemental resources into the primary runtime artifact. This results in an artifact size of approximately 65 MB in Maven Central, which leads to unnecessary memory overhead and an expanded security surface area in production and cloud-native environments.

By decoupling the core generated model logic from these supplemental files, this PR provides a "Thin JAR" alternative that improves deployment efficiency, reduces class-loading times, and simplifies security compliance audits.

_What is being released?_

* Introduction of the `thin` Maven Profile: An opt-in build configuration that generates a lean runtime artifact.
* Secondary Artifact with Maven Classifier: Implementation of the `thin` classifier, allowing the build to generate a secondary, lean JAR (e.g., `cdm-java-thin.jar`) alongside the standard distribution ensuring full backward compatibility for existing pipelines.
* Targeted Resource Exclusions: The following resources are excluded from the `thin` artifact to minimize bloat:

```xml
<excludes>
    <exclude>build-resources/**</exclude>
    <exclude>cdm-sample-files/**</exclude>
    <exclude>config/**</exclude>
    <exclude>functions/**</exclude>
    <exclude>ingest/**</exclude>
    <exclude>ingestions/**</exclude>
    <exclude>mapping-analytics/**</exclude>
    <exclude>release/**</exclude>
    <exclude>result-json-files/**</exclude>
    <exclude>schemas/**</exclude>
    <exclude>**/rosetta/*.rosetta</exclude>
</excludes>

```

_Review Directions_

Maintainers and users can verify the build locally by running:
`mvn clean install -Pthin -DskipTests`

This should produce both the standard JAR and the new `-thin.jar` artifact in the `target` directory and local `.m2` repository.


manel-martos:
The contribution has been synced as discussed during the last TAWG. I will handle the CDM 6.x branch upgrade separately.

@davidalk @hugohills-regnosys - could you please confirm whether the `rosetta-config.yml` resource is required at runtime? As shown in this contribution, it is still included in the `thin` descriptor.

hugohills-regnosys:
> The contribution has been synced as discussed during the last TAWG. I will handle the CDM 6.x branch upgrade separately.
> 
> @davidalk @hugohills-regnosys - could you please confirm whether the `rosetta-config.yml` resource is required at runtime? As shown in this contribution, it is still included in the `thin` descriptor.

It is not required at runtime so could be removed. However is required for it to be run in Rosetta, as it is a small file could we leave it in?

PR #4582, opened 2026-04-07
Chris Rayner Event Model - Transfer Qualifications

### LINKED ISSUE #4075: Update Qualify_CashTransfer to check for Cash
### Background

The Qualify_CashTransfer function is used to confirm that a `transfer` event is transferring cash. This currently checks that the `transferInstruction` contains a `quantity` that has a `currency` within it.

Image

### Proposal

Since the Asset Refactoring, the `Qualify_SecurityTransfer` function has been updated to check the `Asset` as well as the `quantity` criteria to determine that this is a transfer of securities. The `Qualify_CashTransfer` should be updated similarly i.e. check that the `Asset` is of type `Cash`:

Image

This aligns better with `Qualify_SecurityTransfer` and is a more robust check:

Image

### Compatibility

This should be backwards compatible.

### Release

CDM 7

### Additional Context

_No response_

### LINKED ISSUE #4122: What is Qualify_SecuritySettlement actually qualifying?
### Background

The `Qualify_SecuritySettlement` function will return True if you have security transfers and cash transfers in different directions on the same day.

This is ignoring three important aspects of security _settlement_:
1. The `collateralType` - if this is a trade against non-cash then there will be no cash transfer. The current function will thus never be True for non-cash trades.
2. The `quantity` - settlement can be partial or full i.e. either some or all of the securities/collateral can be transferred between the payer and the receiver. Settlement infers that the obligations of the trade have been fully met so the current name of this functions implies it is confirming full settlement, which it does not do.
3. The `deliveryMethod` - if the trade is set as DVP then yes, you should have both cash and security transfers on the same day - but they should also be atomic i.e. in the same transaction. If the trade is FOP then you may not have cash and securities being delivered on the same day, so again this function will not work for FOP trades.

The purpose of this issue is to request feedback from the community about how we should proceed specifically in relation to this function. For example, is anyone actually using this function, and if so, how?

### Proposal

There are a couple of ways to go here, I've illustrated 2 below but welcome suggestions on how to improve this.

**Option1 - Deprecate the function**
Settlement as a term is ambiguous - as I mention above, there are partial and full settlements - but the term should be associated to the contractual obligations related to the transfer of the asset and the transfer of the collateral of a trade being fulfilled.  Currently the function does not do this, for the reasons I mentioned above.

In order to prevent confusion, it may be better to deprecate the function and promote the use of `Qualify_CashTransfer` and `Qualify_SecurityTransfer` to perform and monitor the transfers of securities and cash collateral. The application can then determine when all shares/cash have been received/delivered and set the settlement status accordingly.

**Option 2 - Enhance the function**
As per the Issue description, the function could be enhanced to take into account the `collateralType` and not check for the presence of any cash transfers if the trade is set as "NonCash". In order to qualify as _settlement_ and not just a _transfer_, the function would also need to go through the `transferHistory` on the `tradeState` and determine whether the full quantity/value had been delivered/received.

**N.B.** This ties in with a proposal in #4071, where the suggestion is that functions can be added to the model to derive the settlement status of a trade based upon its `transferHistory`.

### Compatibility

If we decide to deprecate the function then this would be backwards compatible.

If we decide to enhance the function that this would be backwards incompatible.

### Release

CDM 7

### Additional Context

This issue is aligning the model with what happens in security settlement now.

### LINKED ISSUE #4368: Review Qualify_SecurityTransfer
### Background

Whilst working on the transfer event it was found that the qualification functions do not work as anticipated. 

### Task Details

Review the functionality of `Qualify_SecurityTransfer` to ensure that it truly qualifies that the event being processed is indeed a transfer of securities.

### Parent Feature

#4367

### Additional Context

_No response_

### LINKED ISSUE #4369: Review Qualify_CashAndSecurityTransfer
### Background

Whilst working on the transfer event it was found that the qualification functions do not work as anticipated.

### Task Details

Review the functionality of Qualify_CashAndSecurityTransfer to ensure that it truly qualifies that the event being processed is indeed a transfer of both cash and securities

### Parent Feature

#4367

### Additional Context

_No response_

regnosys-prod-user:


regnosys-prod-user:
# *Event Model - Transfer Qualifications*

_Background_

There are currently 4 main qualification functions in the CDM that qualify transfer events. These are: 
- `Qualify_CashAndSecurityTransfer` 
- `Qualify_CashTransfer` 
- `Qualify_SecuritySettlement` 
- `Qualify_SecurityTransfer` 
 
Recent work on the transfer event has highlighted that these functions do not always work as anticipated. The purpose of this release is to update the functions to qualify the events correctly and to align their processing.

_What is being released?_

Changes: 
- `Qualify_CashAndSecurityTransfer` 
  - This function has been removed. 
  - Initial suggestion was to add the `[deprecated]` annotation but it was agreed at the CRWG this function should be removed entirely.
- `Qualify_CashTransfer` 
  - Added a check for the `asset` being `Cash`. 
  - Removed check for `currency` as the fact that `asset` is `Cash` is enough to qualify this as cash. 
- `Qualify_SecuritySettlement` 
  - This function has been removed. 
  - Initial suggestion was to add the `[deprecated]` annotation but it was agreed at the CRWG this function should be removed entirely. 
- `Qualify_SecurityTransfer` 
  - modified to align processing with `Qualify_CashTransfer`. 
  - Added check for `asset` being `Instrument -> Security` 
  - Removed check for `financialUnit` as the fact that `asset` is `Security` is enough to qualify this as a security.

_Review Directions_

Changes can be reviewed in PR: [#4582](https://github.com/finos/common-domain-model/pull/4582)

> [!NOTE]

ReleaseManagement-BP:
Closed as new contribution was required, in #4718

PR #3968, opened 2025-08-06
Paul Hands Legal Agreement - Master - Isla - Added more elections to the isla GMSLA namespace

regnosys-prod-user:

PR #4653, opened 2026-04-28
[master] Paul Hands Legal Agreement - Master - Isla - Added more elections to the isla GMSLA namespace

### LINKED ISSUE #3967: Further enhancement of Skeleton Work for GMSLA
### Background

Now that Foundations is merged, it is a good time to start adding in some (not all) of the extra elections that will help describe a Legal Agreement "GMSLA" in more detail in the CDM representative space. This should be iterated over, and slowly build to a full picture of representing this type of contract.

### Proposal

Add more elections to the isla namespace, with enumerated values. I propose we keep these distinct and separate to begin with, so each agreement can grow organically. At a time in the future when definitions that overlap can be spotted, we can revisit to make shared types and enums, as necessary

### Compatibility

Yes backwards compatible, only adding new types into a skeleton namespace

### Release

Prod, but understand this is iterative

### Additional Context

_No response_

ReleaseManagement-BP:
Release PR for https://github.com/finos/common-domain-model/pull/3968


chrisisla (dismissed):
The function outputs have changed which look okay as they are adding `settlementTerms` which seems logical.
The changes the Math functions look to only be formatting so again should be fine.
The only changes that Paul has made are to the `legaldocumentation` namespace so if we're worried then this PR only needs to commit the files updated in that namespace. Looks good though.

ReleaseManagement-BP:
Hey @paulhands @chrisisla We've had issues with releasing this PR, we're getting errors on the builds because this PR introduces a type Collateral in namespace cdm.legaldocumentation.master.isla.  There is an existing type Collateral in namespace cdm.product.collateral.  Although Rune and the Java / Python code gen supports the same type name in different namespaces, the other code generator (e.g. DAML etc) do not.

The choices are:

- Rename the type in the model
- Fix the code generators (probably quite a lot of work) - we should raise a CDM GitHub stating which code generators do not support this
- Switch off the code generators that do not support namespaces

So with that, I wanted to see if it was feasible at all to go with the first option and rename the type in the model?

Let me know your thoughts

chrisisla:
Hi @ReleaseManagement-BP 
We've updated the name so this should now build okay. 
Thanks!

ReleaseManagement-BP:
Hey @chrisisla please could you look to approve this? Can be merged in once that is done

chrisisla (approved):

PR #4669, opened 2026-05-04
CDM Documentation / Schemas - Navigation: Update link

eacunaISDA:
The footer link previously held the URL we had intended for schemas if served part of the repo, but this is not applicable for now, so we need to use the existing link to the schemas made available via the CDM microsite.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -119,7 +119,8 @@ module.exports = {
             },
             {
               label: 'Schemas',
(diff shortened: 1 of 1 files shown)


ReleaseManagement-BP:
@eacunaISDA Due to Inactivity, I am reaching out to see if this is still required. If I do not hear back in the next week, I will close this due to inactivity. If you would like this re-opened, please reach out to me.

PR #4672, opened 2026-05-05
Chris Rayner Product Model - Remove AssetPayoutDividendTermsValidation condition

### LINKED ISSUE #4625: Remove AssetPayoutDividendTermsValidation condition from EconomicTerms
### Background

In `EconomicTerms` there is a condition named `AssetPayoutDividendTermsValidation` which enforces the inclusion of `terminationDate` if there is an `AssetPayout` which includes `dividendTerms`. 

This is not correct, as trades that have no term applied to them should also be able to set the dividend terms i.e. specify how any dividends that fall during the lifecycle of a trade are to be handled.

### Proposal

Remove the condition.

### Compatibility

This will be backwards incompatible as it will be allowing attributes to be set that previous were not allowed.

### Release

CDM V7

### Additional Context

_No response_

regnosys-prod-user:


regnosys-prod-user:
# *Product Model - Remove AssetPayoutDividendTermsValidation condition*

_Background_

In `EconomicTerms` there is a condition named `AssetPayoutDividendTermsValidation` which enforces the inclusion of `terminationDate` if there is an `AssetPayout` which includes `dividendTerms`. This is not correct, as trades that have no term applied to them should also be able to set the dividend terms i.e. specify how any dividends that fall during the lifecycle of a trade are to be handled.

_What is being released?_

The condition `AssetPayoutDividendTermsValidation` has been removed from `EconomicTerms`

_Review Directions_

Changes can be reviewed in PR: [#4672](https://github.com/finos/common-domain-model/pull/4672)

> [!NOTE]

PR #4690, opened 2026-05-07
Chris Rayner Event Model - Support for Rate Changes on Securities Lending trades

### LINKED ISSUE #3640: Support for Rate Changes on Securities Lending trades
### Background

**What is a rate change in securities lending?**
A Rate Change is where the interest rate or lending fee applied to a trade is adjusted, which affects the cost of borrowing for the borrower and the return for the lender.

The rate can be defined in one of two ways:
- Fixed figure agreed between the borrower and the lender at trade initiation. Changes in this rate could be due to increased demand for the securities on loan for example.
- A benchmark Index rate is agreed to be used on the trade at initiation. Changes in the rate associated to an index are generally due to fluctuations in market conditions or changes in the overall interest rate environment i.e. based upon an observation.

For securities lending trades, even though a change in the interest rate constitutes a change to the terms of the contract, a rate change results in an amendment to the existing contract rather than a cancellation.

**Analysis of existing CDM functionality**
The CDM contains several possible ways to support rate changes on securities lending trades, the main 3 are discussed here:

_On Demand Rate Change processing_
This was added for repo processing. When the rate is changed it closes the current contract and opens a new one with the new rate. This is consistent with repo processing, but, as mentioned previously, for securities lending the contract is not cancelled, so this processing cannot be used for our use case.

_QuantityChange and TermsChange processing_
The interest rate is held in the payouts and on the `TradeLot`. It would be possible to use a `QuantityChangeInstruction` to change the rate, and a `TermsChangeInstruction` to change the effective date. However, by changing the rate and the effective date we will be losing the original rate and date which are required for reporting and billing purposes.

_Reset processing_
The `Reset` functionality can be used to store details of adjustments to the trade terms. These are recorded on the `TradeState` which allows a lineage of adjustments to be stored. However, this expects an `InterestRatePayout`, which we will only have if the trade is against cash. It also expects an `Observation`, which will only be the case where the trade is using a benchmark index. Also, for securities lending, a rate change is not considered to be the same as a `Reset`, as a rate change is a reaction to a market adjustment whereas a reset is a scheduled event written into the terms of the agreement.

_(N.B. a reset might occur if there is a schedule tied to a regular benchmark rate adjustment)_

**Conclusion**
The existing processing does not meet our requirements. 

### Proposal

The requirement is to be able to record changes in the rate on securities lending trades. Multiple rates and their effective dates will need to be stored. Event functions are required to maintain this new rate change history, as well as qualify that these events are rate changes.

**Types**
New types to store the rate change details are proposed as follows:
- `LoanRateChange` which will hold:
  - A mandatory `newRate` as a number representing the new rate being applied
  - A mandatory `timestamp` as a new `zonedDateTime` to hold the date/time the rate is effective from
  - An optional `observation` where any observation that was the driver for the change in rate can be described (primarily for benchmark index changes)

Image

- `TradeState` -> `loanRateChangeHistory` which will be a list of rate changes with their effective dates, providing the lineage required for billing

Image

For processing of the rate change event we will need a new primitive instruction:
- `LoanRateChangeInstruction` which will hold:
  - A mandatory `loanRateChange`, containing the new rate and its date
  - A mandatory `replace` Boolean, allowing the rate change history already stored on the `TradeState` to either be appended to or replaced (following similar logic from the `Valuation` processing).

Image

There will also be some changes to existing types and enums required which will be determined during the development.

**Functions**
Two new functions will be required:
- `Create_LoanRateChange`
  - This will add the `loanRateChange` passed in to the `TradeState` -> `loanRateChangeHistory` list
  - It will be able to be called by `Create_BusinessEvent`
- `Qualify_LoanRateChange`
  - This will simply look for the inclusion of a `loanRateChange` in the `instruction`.

### Compatibility

Everything being added is new and so should be backwardly compatible.

### Release

This change is required on the current development version only i.e. 7dev.

### Additional Context

This proposal has been derived by the ISLA CDM WG in conjunction with existing production users of the CDM.

regnosys-prod-user:


regnosys-prod-user:
# *Event Model - Support for Rate Changes on Securities Lending trades*

_Background_

The processing of the `QuantityChange` primitive instruction is being enhanced to support the entry and maintenance of a schedule of interest rates in a `PriceSchedule` using the `datedValue` list.

_What is being released?_

The key changes are as follows: 

Function `UpdatePriceAmountForEachMatchingQuantity` 
- Now expects a `PriceQuantity` rather than a `Price`as input 
- Calls a new `UpdateDatedValue` function to update a `datedValue` entry if one already exists for the effective date of the price change 
- Calls a new `InsertDatedValue` function to insert a new entry in the `datedValue` list if one does not exist already for the given effective date 
- The original functionality to update the `value` in a `PriceSchedule` remains unchanged 

Function `PriceUnitEquals`
 - Has been enhanced to take the `priceSubType` into account when comparing two `PriceSchedule` instances 

Function `UpdateDatedValue` 
- New function to update only those entries in a `datedValue` list that have a date that matches the `effectiveDate` passed in. 
- Note: There is an existing function `UpdateDatedValues` which will update any entries in `datedValue` whose date is greater than or equal to the `effectiveDate`. This is not the required processing for price (rate) changes so it was felt that creating a separate function would have less impact on existing use cases. 

Function `InsertDatedValue` 
- New function to insert a new entry into a `datedValue` list based upon the `effectiveDate` that was passed in. 
- Supports prefixing where the `effectiveDate` is prior to the earliest date in the `datedValue` list, and appending where the `effectiveDate` is after the latest date in the `datedValue` list 
- Note: as per the comment on the type, the `datedValue` list is expected to be maintained in ascending date order

_Review Directions_

Changes can be reviewed in PR: [#4690](https://github.com/finos/common-domain-model/pull/4690)

> [!NOTE]

rosetta-source/src/main/rosetta/product-common-settlement-func.rosetta
```diff
@@ -59,20 +59,50 @@ func UpdatePriceAmountForEachMatchingQuantity: <"Updates any price from the Pric
...
+        direction QuantityChangeDirectionEnum (1..1) <"Enum specifying how the updated amounts should be applied, e.g. add, subtract or replace.">
     output:
         updatedPrice PriceSchedule (0..1)
             [metadata location]
 
-    alias changedAmount:
+    alias changeMatching: <"Locate the first PriceQuantity item where the attributes in the existing and the new price match.">
         change
-            filter PriceUnitEquals(item, price)
+            filter PriceUnitEquals(item -> price only-element, price)
```
Oscar-Twomey: Only element is assuming the change will only have one price schedule inside it - is this intentional?

Oscar-Twomey: This could be an alternative...but I haven't tested it :) 

```
alias changeMatching: 
    change
            filter c [
                c -> price
                    extract PriceUnitEquals(item, price)
                    then only-element
            ]
            then first
```

This would filter the list of changes (f) where an item in the list of prices within matches the price (and assuming there's only 1 match per change it uses only-element.)
 
That way if there's multiple price schedules it should still work.

The logic would return a single change with a matching price

rosetta-source/src/main/rosetta/base-math-func.rosetta
```diff
@@ -135,6 +135,69 @@ func UpdateAmount: <"Updates an amount based on the given QuantityChangeDirectio
...
+        effectiveDate date (0..1)
+    output:
+        newdatedValues DatedValue (0..*)
+
+    add newdatedValues: <"Update only the item(s) whose date exactly match the effective date.">
+        datedValues
+            extract
+                DatedValue {
+                    date: item -> date,
+                    value: if item -> date = effectiveDate
```
Oscar-Twomey, line 190: You could use the `UpdateAmount` function above and pass in `item -> value`, `changeAmount`, and the Direction as arguments.

rosetta-source/src/main/rosetta/base-math-func.rosetta
```diff
@@ -135,6 +135,69 @@ func UpdateAmount: <"Updates an amount based on the given QuantityChangeDirectio
...
+
+    alias insertDatedValue: <"Create the datedValue we will insert. If there are no prior dateValues then this is a new entry at the beginning of the list which is only valid when this is a Replace (as you cannot Increase or Decrease an unknown number). If we have datedValues prior to the effective date then take the value from the last of the datedValues whose date is prior to the effective date and apply the requested change to that figure (i.e. Replace, Increase or Decrease).">
+        if previousDatedValue = empty and direction = Replace then
+            DatedValue {
+                date: effectiveDate,
+                value: changeAmount
+            }            
+        else
+            previousDatedValue
+                extract
```
Oscar-Twomey: Very minor, but the `previousDatedValue extract` isn't necessary here - you can just go straight into creating the `DateValue` type.

rosetta-source/src/main/rosetta/base-math-func.rosetta
Oscar-Twomey: Same for the `UpdateDatedValues` function below

rosetta-source/src/main/rosetta/base-math-func.rosetta
Oscar-Twomey: And replace item with previousDatedValue

rosetta-source/src/main/rosetta/product-common-settlement-func.rosetta
chrisisla: I've tested the new suggested approach with my test cases and it does work for those scenarios okay 👍 
I will update accordingly

rosetta-source/src/main/rosetta/base-math-func.rosetta
chrisisla: Good call, I will update both the `UpdateDatedValue` and `UpdatedDatedValues` functions.

rosetta-source/src/main/rosetta/base-math-func.rosetta
chrisisla: I had a bit of trouble with this. If I took the `extract` out then I was getting new `DatedValue` items created with the value as null. This was because where `previousDatedValue` was `empty`, the `extract` was catering for the fact there were no dates to loop through and thus not adding any new items. When I took the `extract` out, null entries started to appear.

To correct this I have updated the test on `previousDatedValue` to check for `is absent` and added a test in the else clause for `previousDatedValue exists`. This works okay.

regnosys-prod-user:
_What is being released?_

Changes from review comments

> [!NOTE]

PR #4693, opened 2026-05-07
[master] Chris Rayner Event Model - Support for Rate Changes on Securities Lending trades

### LINKED ISSUE #3640: Support for Rate Changes on Securities Lending trades

ReleaseManagement-BP:
Release PR for https://github.com/finos/common-domain-model/pull/4690


ReleaseManagement-BP:
Hi @llynhiavu please could you look at reviewing this PR when you have a chance?

llynhiavu (approved):
Reviewed and have no comment. Looks okay to me.

rosetta-source/src/main/resources/functions/repo-and-bond/cancellation-output.json
```diff
@@ -1404,6 +1404,12 @@
                   "value" : "GBP"
                 }
               },
+              "datedValue" : [ {
```
Oscar-Twomey, line 1407: @llynhiavu @chrisisla 

This doesn't look right. 

These samples have both a date & a dated value - is that correct?

Also, the dated value doesn't have a date - that seems wrong

chrisisla: Within a `datedValue` there should be both a `date` and a `value`, you should not end up with one without the other.

It is possible to have both a `value` and a `datedValue` in a `PriceShedule` for example. In this situation the `value` is taken to be the initial value.

Oscar-Twomey: Although its valid to have both, in this sample it doesn't look correct as there is only 1 price. 

I expect that because this is a single price (rather than a schedule) there is no date - hence it shouldn't be populating dated value

Oscar-Twomey: Its possible it has something to do with the defaulting on `UpdateQuantityAmountForEachMatchingQuantity`

PR #4706, opened 2026-05-11
[master] Chris Rayner Product Model - Remove AssetPayoutDividendTermsValidation condition

### LINKED ISSUE #4625: Remove AssetPayoutDividendTermsValidation condition from EconomicTerms

ReleaseManagement-BP:
Release PR for https://github.com/finos/common-domain-model/pull/4672


ReleaseManagement-BP:
Hi @tomhealey-icma please could you look at reviewing this PR when you have a chance?

tomhealey-icma (approved):

PR #4742, opened 2026-05-20
Added settlement party to settlement base

tomhealey-icma:
This enhancement adds settlement party and settlement party role to settlement  base.


regnosys-ops:
No Expectations generated for this PR. Make sure build was run with tests and has passed before merging.

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
```diff
@@ -141,6 +145,10 @@ enum AncillaryRoleEnum: <"Defines the enumerated values to specify the ancillary
     CalculationAgentOptionalEarlyTermination <"Specifies the party responsible for performing calculation agent duties associated with an optional early termination.">
     CalculationAgentMandatoryEarlyTermination <"Specifies the party responsible for performing calculation agent duties associated with an mandatory early termination.">
     CalculationAgentFallback <"Specifies the party responsible for deciding the fallback rate.">
+    SecuritiesSettlementCentre <"Specifies the party that processes the transfer and delivery of securities between issuers, counterparties, agents and other service providers.">
+    CashSettlementCentre <"Specifies the party that processes the transfer and delivery of cash between issuers, counterparties, agents and other service providers.">
+    PlaceOfDeposit <"Specifies the party responsible for recording securities ownership, holding physical certificates and facilitating securities transfers.">
```
llynhiavu: @tomhealey-icma 
1. At first glance I wasn't sure a PlaceOfDeposit was a Role (like a calculation agent or custodian, which are clearly parties) and seemed out of place. That said, you also added enum values using similar terminology (e.g. cash settlement centre") to specify place/institution which I suppose makes sense (so no issue I suppose. was just explaining my thought process while reviewing)

2. there an overlap in definition between SecuritiesSettlementCentre and PlaceOfDeposit. They both "facilitate/process ... securities transfers." is that alright? won't cause confusion among users?

3. implementation looks good to me otherwise.

llynhiavu (commented):
minor terminology comments.

tomhealey-icma:
As discussed at the CRWG meeting, I updated the condition to optional and added another condition to check for one-of SettlementCentre, SecuritiesSettlementCentre or CashSettlementCentre. I did not remove the deprecated SettlementCentreEnum. Branch had been updated and build successful

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
lolabeis: I'm not sure these different roles, while all relevant in an Asset context, make sense in the context of defining "economic terms". I was expecting only a single new enum value here In AncillaryRoleEnum: SettlementCentre.

Remember that this enum value would only appear in EconomicTerms > ... > Payout > SettlementTerms, in which the custodian or place of deposit aren't relevant.

In relation to those different settlement centre values (security vs cash), can you clarify:
Do those define the settlement centre for buy/sell transactions in the security (and therefore relevant for the economic terms of a Trade in that security), or for the security's own lifecycle event like coupon payment (and therefore relevant for the economic terms of an Asset / TransferableProduct)?

rosetta-source/src/main/rosetta/base-staticdata-party-enum.rosetta
```diff
@@ -141,6 +145,10 @@ enum AncillaryRoleEnum: <"Defines the enumerated values to specify the ancillary
     CalculationAgentOptionalEarlyTermination <"Specifies the party responsible for performing calculation agent duties associated with an optional early termination.">
     CalculationAgentMandatoryEarlyTermination <"Specifies the party responsible for performing calculation agent duties associated with an mandatory early termination.">
     CalculationAgentFallback <"Specifies the party responsible for deciding the fallback rate.">
+    SecuritiesSettlementCentre <"Specifies the party that processes the transfer and delivery of securities between issuers, counterparties, agents and other service providers.">
```
lolabeis: `SecuritySettlementCentre`, singular preferred for attribute names or enum values.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
```diff
@@ -373,10 +373,26 @@ type SettlementBase: <"A base class to be extended by the SettlementTerms class.
...
     settlementDate SettlementDate (0..1) <"The date on which the settlement amount will be paid, subject to adjustment in accordance with any applicable business day convention. This component would not be present for a mandatory early termination provision where the cash settlement payment date is the mandatory early termination date.">
     settlementCentre SettlementCentreEnum (0..1) <"Optional settlement centre as an enumerated list: Euroclear, Clearstream.">
+        [deprecated]
```
lolabeis: I did check and this attribute doesn't appear to be used, neither as inbound (for ingestion) or outbound (for DRR). Therefore it should be safe to remove.

We can keep it as deprecated in CDM 7, but should then immediately proceed to delete it in CDM 8.

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
```diff
@@ -373,10 +373,26 @@ type SettlementBase: <"A base class to be extended by the SettlementTerms class.
...
     settlementProvision SettlementProvision (0..1) <"Optionally defines the parameters that regulate a settlement.">
     standardSettlementStyle StandardSettlementStyleEnum (0..1) <"Settlement Style.">
+    settlementParty Party (0..*) <"Settlement entities defined within the product.">
+    settlementPartyRole PartyRole (0..*) <"Roles of the settlement entities such as agents, custodians, etc.">
```
lolabeis: These 2 attributes (`settlementParty` and `settlementPartyRole`) should be removed.

The whole point of introducing the `AncillaryRoleEnum` attribute is so that those parties **do not need** to be defined here. Instead, they can be specified using existing attributes:
* TradableProduct -> ancillaryParty
* Asset -> ancillaryPartyRole

rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
```diff
@@ -373,10 +373,26 @@ type SettlementBase: <"A base class to be extended by the SettlementTerms class.
...
+    tradableSettlementCentre AncillaryRoleEnum (0..*) <"Specifies the settlement ancillary party of the tradable product.">
+    assetSettlementCentre AssetAncillaryPartyRoleEnum (0..*) <"Specifies the settlement ancillary party of the asset.">
```
lolabeis: Why do these 2 attributes (`tradableSettlementCentre` and `assetSettlementCentre`) need to be multi-cardinality?

This will live inside Payout > SettlementTerms. Can a payout have more than one settlement centre? Note that the deprecated `settlementCentre` attribute was (0..1).

This multi-cardinality complicates the logic of the SettlementCentreParty condition.

lolabeis (commented):
Detailed review comments in-line.

Also the latest commit seems to have misplaced some enum files in rosetta-source/src/main/resources, which should be deleted.

tomhealey-icma:
@lolabeis - Responses to your comments:
1) ICMA members, specifically the ICSDs noted that they need separate and distinct identifiers for Security Settlement and Cash Settlement, thus the need for separate enums and multi-cardinality. 
2) Place of Deposit is distinct as the holder of the asset and distinct from the entity does the processing.
3) Settlement Centre are holders of record even after issuance so both for coupon payments and any trades (transfers), unlike IRS or FX.
4) I will remove the old SettlementCentreEnum
5) I will remove the settlement party attributes.

tomhealey-icma:
I've made the latest updates and successfully rebuilt. Ready for approval.

lolabeis:
Caught-up with @tomhealey-icma, summary of our discussion:
* We don't need a `tradableSettlementCentre` attribute, only the asset one is relevant
* No need for new enum values in `AncillaryRoleEnum`
* 4 new enum values in AssetAncillaryPartyRoleEnum:
  * CashSettlementCentre
  * SecuritySettlementCentre
  * PlaceOfDeposit
  * Custodian
* Rather than a multi-cardinality `assetSettlementCentre`, we need 2 optional attributes of single-cardinality each:
  * `cashSettlementCentre`, whose value should be `CashSettlementCentre`
  * `securitySettlementCentre` whose value should be `SecuritySettlementCentre`
* Do not modify the snippets in the documentation for older versions of the model.
* Remove the extraneous enum files in /resources

ReleaseManagement-BP:
Superseded by #4914

Issue #4584, opened 2026-04-07
Python: Top Level Package Name Collision Prevention


dschwartznyc:
### Background

Repeating issue [181](https://github.com/finos/rune-python-generator/issues/181) from the Rune-Python-Generator:

The generated python package structure follows strictly the model defined namespace hierarchy. As the model is created in rune, the modelers have the full freedom to choose namespace names, which potentially can lead to clashes with existing 3rd party python packages.

An example is the current cdm wheel, which contains a top-level package named fpml.

### Proposal

- Nest all generated top-level packages in a package which is unique to rune/finos. The top-level package name could be `rune` or `finos`.
- Provide a command line switch in the generator CLI which controls the name of this top-level package name. Possibly defaulting to `finos` or `rune`.

As an example, let's consider the rune namespace namespace `cdm.base.staticdata.asset.common`. Assuming the top-level package name is `finos`, then the namespace will translate in the python package `finos.cdm.base.staticdata.asset.common`. Also the namespace `fpml` will translate to the python package name `finos.fpml`

**NOTE**: The above is about the structure of the packages in the wheel, and not about the name of the cdm wheel, which should stay as it is (`finos-cdm`).

### Compatibility

The solution would not be backwards compatible.

### Release

CDM 6.x.x +

dschwartznyc:
There were two questions raised in the [April 9th TAWG](https://github.com/finos/common-domain-model/discussions/4573).

1. **Why is this necessary?**
To remove the potential for the new CDM packages to have namespace clashes with existing publicly available Python projects.  For example, there is already an [FPML package available at PyPI.org](https://pypi.org/project/fpml/).  Additionally, with Microsoft's use of the name [common domain model](https://learn.microsoft.com/en-us/common-data-model/data-lake) in its Azure product, there similarly may be packages which use the "cdm" namespace.

2. **Can we use the Java style fully qualified names (org.finos.cdm...)?**
As supported by the below this is not a Python standard approach.

**PEP 8 – The "Official" Style Guide**

[PEP 8](https://www.google.com/search?q=https://peps.python.org/pep-0008/%23package-and-module-names) is the definitive reference for Python code. It states:

> "Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged."

[...]

Issue #4224, opened 2025-11-28
Python Function Support


dschwartznyc:
### Background

The Rune Python Generator does not support generation of Python Code from Rune function definitions (see [Rune Python Generator issue #108](https://github.com/finos/rune-python-generator/issues/108) and [Rune Python Rune Issue #13](https://github.com/finos/rune-python-runtime/issues/13).  This becomes increasingly challenging for CDM as there is increased function use in the model including as part of a Type's conditions. 

### Proposal

Integrate changes once made to the Rune Python Generator ([Issue #108](https://github.com/finos/rune-python-generator/issues/108)) and the Rune Python Runtime ([Issue #13](https://github.com/finos/rune-python-runtime/issues/13)).

### Compatibility

The addition of this new functionality should be backwards compatible.

### Release

CDM 5.x, 6.x, 7

PR #4750, opened 2026-05-21
[5.x.x] Re-enables CDM's Python code generation using the new python generator/runtime

### LINKED ISSUE #3239: Proposed change to the CDM - Python deployment to PIP
The Python library generated from the in production CDM Rune definitions should be deployed to [The Python Package Index (PyPI) ](https://pypi.org) in addition to the current Maven Central deployment.  Doing so would promote adoption of CDM by expanding its access to all Python users.  The proposal is to limit distribution to only in production versions of CDM pruning prior releases in a manner TBD and to be aligned with Python industry standard practices.

@plamen-neykov @manel-martos

dschwartznyc:
**Summary**

Re-enables CDM's Python code generation using the [newly released rune-python-generator](https://github.com/finos/rune-python-generator/pull/184) and moves Python artifact distribution from Maven Central to [PyPI](https://pypi.org/), the Python community standard (Issue [#3239](https://github.com/finos/common-domain-model/issues/3239)).

**What's New**

The updated generator:

*   Generates Python from Rune defined functions ([Python Function Support #4224](https://github.com/finos/common-domain-model/issues/4224))
*   Protects against namespace collisions ([Python: Top Level Package Name Collision Prevention #4584](https://github.com/finos/common-domain-model/issues/4584))
*   Adds support for previously missing collection and list operators: sort, min, max, reduce, distinct, flatten, reverse, sum, one-of
*   Fixes circular references in inheritance and attribute handling
*   Completes support for type aliases
*   Refactors object, attribute, and expression generation

**Breaking Changes**

*   Consistent with the outcome of the [recent CDM community survey](https://github.com/finos/common-domain-model/issues/3239#issuecomment-4313502510), distribution to Maven Central is discontinued. Downstream consumers should migrate to PyPI.

**Known Limitations**

*   No support for functions natively defined in Java
*   No support for the recent [code list changes](https://github.com/finos/common-domain-model/pull/3819)


codefresh.yml
```diff
@@ -241,20 +240,23 @@ steps:
...
+    fail_fast: false
+    image: python:3.12-slim
+    working_directory: ./rosetta-source/src/generated/python
+    when:
+      condition:
+        all:
+          isRelease: "${{IS_RELEASE}} == true"
+          pythonBuildSuccess: steps.BuildPython.result == 'success'
+    environment:
+      - TWINE_PASSWORD="${{PYPI_TOKEN}}"
```
hugohills-regnosys: I think it the end we needed to remove these quotes:
```
TWINE_PASSWORD=${{PYPI_TOKEN}}
```

dschwartznyc: done

hugohills-regnosys (approved):

PR #4792, opened 2026-06-01
Oscar Twomey Asset - Moving Price Quantity Triangulation to `PriceQuantity`

### LINKED ISSUE #4109: Quantity x Price triangulation rule should apply whenever a quantity is specified as a no. units
### Issue Summary

A PriceQuantityTriangulation rule is implemented at the TradableProduct level to check the internal consistency of the price and quantities provided.

When quantity is specified both as a number of asset's units _and_ as an equivalent monetary quantity, the rule is meant to check that:
* No units x price = monetary quantity

However, the condition's implementation is inconsistent with the asset's quantity being specified as a no. units. The no. units could be anything from commodity, shares or even currency.

There are several issues with the current implementation:
1. The logic filters on priceType = CashPrice, which corresponds to a price specified as a simple "payment" with no reference to an underlying asset. Instead, it should filter on AssetPrice and ExchangeRate, when the asset quantity can be specified as a no. units.
2. The logic narrowly applies to the case when unit is a no. shares. It should apply broadly.
3. The condition should apply directly at the PriceQuantity level, rather than at the TradableProduct level and then iterating across PriceQuantity items.

### Steps to Reproduce

* Inspect the `PriceQuantityTriangulation` condition at the `TradableProduct` level
* Inspect `TradeLot` and `PriceQuantity`: no such condition exists
* Inspect the logic in the condition's underlying `func` components:
  * `PriceQuantityTriangulation`
  * `CashPriceQuantityNoOfUnitsTriangulation`

### Expected Behaviour

1. The logic should filter on priceType = AssetPrice and ExchangeRate, when the asset quantity can be specified as a no. units.
2. The logic should apply broadly when an asset quantity is specified.
3. The condition should apply directly at the PriceQuantity level.

### Actual Behaviour

1. The logic filters on priceType = CashPrice, which corresponds to a price specified as a simple "payment" with no reference to an underlying asset.
2. The logic narrowly applies to the case when unit is a no. shares.
3. The condition applies at the TradableProduct level and then iterates across PriceQuantity items.

### Compatibility

Fixing the bug will make the condition more stringent and is therefore backward incompatible

### Release

CDM 7

### Additional Context

This bug is related to #3541, because the same triangulation rule should apply to the Transfer object, when that Transfer is also specified as the asset's no. units and a monetary equivalent.

regnosys-prod-user:


regnosys-prod-user:
# *Asset - Moving Price Quantity Triangulation to `PriceQuantity`*

_Background_

A `PriceQuantityTriangulation` condition is implemented at the `TradableProduct` level to check the internal consistency of the price and quantities provided. There are limitations to the existing implementation and the placement of the condition

_What is being released?_

- Move the condition to `PriceQuantity` so the check happens on the prices &amp; quantities rather than on the trade lot 
- Remove the `CashPriceQuantityNoOfUnitsTriangulation` function and place the logic directly in the `PriceQuantityTriangulation` function
- Remove references to the unit type and apply the condition to a broader use case

_Review Directions_

Changes can be reviewed in PR: [#4792](https://github.com/finos/common-domain-model/pull/4792)

> [!NOTE]

chrisisla (commented):
By moving the check into `PriceQuantity` I believe this will now check all `priceQuantity` objects rather than just those in the `tradeLots`. Looking at what we have in the payouts and collateral objects I think the triangulation will still work, but have you considered when a price is affected by margins, collateral haircuts or clean vs dirty price?

rosetta-source/src/main/rosetta/product-template-func.rosetta
```diff
@@ -93,24 +93,37 @@ func CompareTradeLotToAmount: <"Compare the Quantity amount in TradeLot to the g
             then extract [ CompareNumbers(item -> value, op, amount) ]
                 all = True
 
-func PriceQuantityTriangulation: <"Defines all the scenarios which triangulation can be helpful validation between Prices and Quantities.">
+func PriceQuantityTriangulation: <"Checks the relationship between price, multiplier, notional and number of units, and validates the result of their product.">
```
lolabeis: This function belongs in the `observable.common` namespace.

rosetta-source/src/main/rosetta/product-template-func.rosetta
```diff
@@ -93,24 +93,37 @@ func CompareTradeLotToAmount: <"Compare the Quantity amount in TradeLot to the g
...
-        tradeLots TradeLot (0..*)
+        price PriceSchedule (0..*)
+        quantity NonNegativeQuantitySchedule (0..*)
     output:
         success boolean (1..1)
 
+    alias notional:
+        FilterQuantityByCurrencyExists(quantity)
+            extract value
+            then distinct only-element
```
lolabeis: In the ExchangeRate case, there will be 2 quantities with currency, so that logic isn't robust.

We need to apply a smarter filtering using the 2 unit attributes (`unit` and `perUnitOf`) in assetPrice, namely:
- for the `notional` alias: unit should match `unit`
- for the `noOfUnits` alias: unit should match `perUnitOf`

This logic will work in both the ExchangeRate and AssetPrice cases.

rosetta-source/src/main/rosetta/product-template-func.rosetta
```diff
@@ -93,24 +93,37 @@ func CompareTradeLotToAmount: <"Compare the Quantity amount in TradeLot to the g
...
+
+    alias multiplier:
+        quantity
+            extract multiplier -> value
+            then only-element default 1.0
+
+    alias noOfUnits:
+        quantity
+            extract value
+            then only-element
```
lolabeis: See above comment, needs filtering

rosetta-source/src/main/rosetta/product-template-func.rosetta
```diff
@@ -93,24 +93,37 @@ func CompareTradeLotToAmount: <"Compare the Quantity amount in TradeLot to the g
...
+
+    alias assetPrice:
+        price
+            extract value
+            then only-element
```
lolabeis: This should filter on AssetPrice (or ExchangeRate), as in the original function's logic.

rosetta-source/src/main/rosetta/product-template-func.rosetta
```diff
@@ -93,24 +93,37 @@ func CompareTradeLotToAmount: <"Compare the Quantity amount in TradeLot to the g
...
-                extract [
-                    if priceQuantity -> price -> priceType contains AssetPrice
-                            or priceQuantity -> price -> priceType contains ExchangeRate
-                    then CashPriceQuantityNoOfUnitsTriangulation(
-                                priceQuantity -> quantity,
-                                priceQuantity -> price
-                            )
-                    else True
-                ] all = True
+        if price -> priceType contains [AssetPrice, ExchangeRate]
```
lolabeis: Incorrect use of `contains` (LHS should be a list, RHS a singleton).

But this if statement is redundant anyway because price is already filtered on AssetPrice or ExchangeRate, so `assetPrice exists` suffices.

regnosys-prod-user:
_What is being released?_

Updating functions for review comments

> [!NOTE]

CDM-ReleaseManagement-OT:
Superseded by #4849

PR #4810, opened 2026-06-04
DSL 10.0.0-dev.14

### LINKED ISSUE #4308: `distinct count = 1` bug for different zeros
### Issue Summary

A discrepancy has been identified regarding the `distinct count = 1` combination of operators. When processing a list of values that are numerically zero but populated using different data types (_i.e._, an integer 0 and a float 0.0), the operation `distinct count = 1` returns False, whereas it is expected to return True.

### Steps to Reproduce

To replicate the scenario described above, two functions (_e.g._, `AllEqualZeros` and `DistinctCount1` rules) have been created utilizing different operator combinations: one employing `all = 0` and the other employing `distinct count = 1`.

Image

As observed in the image above, the `interestRatePayout` has multiple cardinality; therefore, it can contain multiple values.

Subsequently, we reference the reporting rules within the specific type (in this instance, the `CFTCTransactionReport` type) to evaluate the results.

Image

In the sample used as example, we populated one `value` as `0` and another as `0.0`.

Image

### Expected Behaviour

Since both values are essentially equal to zero, we would expect the report to indicate **All values are zero** in both cases.

### Actual Behaviour

However, while we obtain the expected result, _i.e._ **All values are zero**, when using `all = 0`, we receive **NOT all values are zero** when applying `distinct count = 1`.

Image

### Compatibility

As this issue has been observed within the leg ordering functions, we have noted that it causes them to malfunction, leading to an erroneous selection of legs. Furthermore, if this behavior extends to other functions utilizing the `distinct count = 1` operator, it may result in incorrect reporting across those respective use cases.

### Release

Since this likely necessitates a change in DSL, it is possible that all versions of the CDM will be affected.

### Additional Context

_No response_

### LINKED ISSUE #4637: Replace choice type path with switch or as operator
### Background

Path operator for choice types creates ambiguity and potential misuse when path traversal is applied directly to choice types themselves rather than to guaranteed attributes. A warning is being added to the DSL to prohibit the usage of choice types via path operators.

### Proposal

Update any logic in CDM to use a `Switch` or `As` statement where it currently uses a path operator

```
asset switch
  Basket then <item is now of type Basket>,
  Loan then <item is now of type Loan>,
  Security then <item is now of type Security>,
  default then <optional default case>
```

Sometimes using a `switch` to access single `choice` type option is verbose and affects readability.

Consider `choice` type below:
```
choice Payout: 
    AssetPayout
    InterestRatePayout 
    OptionPayout 
    PerformancePayout
```

To access a single option, the syntax would be:
```
payout switch 
    PerformancePayout then item,
    default empty
```

Accessing single options is a common use-case, and to maintain model readability, the `as` keyword could be used:
```
payout as PerformancePayout
```

### Parent Feature

#4628

### Additional Context

This is related to Rune-dsl issues
- [#1198](https://github.com/finos/rune-dsl/issues/1198) Choice type path validation
- [#1231](https://github.com/finos/rune-dsl/issues/1231) `as` keyword for accessing sub `type` and single `choice` type options

SimonCockx:
# _Infrastructure - Dependency Update_

Version updates include:
  * `DSL` `10.0.0-dev.13` - Rune formatting bug fix. See DSL release notes: [10.0.0-dev.13](https://github.com/finos/rune-dsl/releases/tag/10.0.0-dev.13)
  * `DSL` `10.0.0-dev.14` - Introduces the `as` operator, fixes a bug to do with `distinct`, and removes deprecated `synonym` syntax. See DSL release notes: [10.0.0-dev.14](https://github.com/finos/rune-dsl/releases/tag/10.0.0-dev.14)

This release fixes https://github.com/finos/common-domain-model/issues/4308, and also deletes an unused legacy unit test.

_Review Directions_

Changes can be reviewed in PR: #4810


hugohills-regnosys:
The `as` syntax is still pending review in the TAWG, and not supported in the Python code gen yet, so modellers should not use this syntax in the CDM until approved.

hugohills-regnosys (approved):

PR #4812, opened 2026-06-05
[master] RecordKeeping Ingest - Inflation Swap Mappings (Update Expectations)

### LINKED ISSUE #4747: Mapping Translate 2.0 Gap - InflationRateSpecification fields not being mapped
### Issue Summary

`MapInflationRateSpecification` in Translate 2.0 mappings sets the following elements to empty: `inflationLag`, `indexSource`, `mainPublication`, `interpolationMethod` and `fallbackBondApplicable`. 

Image

However, these elements were implemented in the legacy synonyms implementation.

Image

### Steps to Reproduce

Ingest the `sample v1_inflation asset swap ex03 par par` from the test pack `fpml 5 10 incomplete products inflation swaps` using the legacy mappings and and compare the result with Translate 2.0 ingestion functions.

### Expected Behaviour

Inflation features should be mapped into the CDM object, as in this result obtained using legacy mappings:

Image

### Actual Behaviour

Inflation features lack in the CDM message, when this is built using Translate 2.0:

Image

### Compatibility

Bakcward compatible.

### Release

CDM 6 Prod and 7 dev.

### Additional Context

_No response_

erincassidy1:


AkashP-REGnosys (approved):

ReleaseManagement-BP:
Hi @chrisisla @llynhiavu @manel-martos would one of you be able to review this PR please with the aim of having this merged before the code freeze on the 15th June

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-product-swap-func.rosetta
```diff
@@ -301,33 +301,40 @@ func MapInflationRateSpecification:
...
+            indexSource: MapStringWithScheme(
+                        indexSource -> value,
+                        indexSource -> rateSourcePageScheme
+                    ),
+            mainPublication: MapStringWithScheme(
+                        mainPublication -> value,
+                        mainPublication -> mainPublicationScheme
+                    ),
+            interpolationMethod: interpolationMethod -> value to-enum InterpolationMethodEnum,
+            fallbackBondApplicable: fpmlInflationRateCalculation -> fallbackBondApplicable,
```
chrisisla: We're doing an `extract` on `fpmlInflationRateCalculation` so rather than `fpmlInflationRateCalculation -> fallbackBondApplicable` shouldn't this be `item -> fallbackBondApplicable` or just `fallbackBondApplicable` ?

erincassidy1: Hi @chrisisla, good spot! I've removed reference to `fpmlInflationRareCalculation`. Mapping now shows -
`fallbackBondApplicable`: `fallbackBondApplicable`. Thanks.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-product-swap-func.rosetta
```diff
@@ -301,35 +301,32 @@ func MapInflationRateSpecification:
...
+                                floorRateSchedule,
+                                fpmlLeg
+                            ),
+                    initialRate: MapInitialRate(initialRate, floatingRateIndex),
+                    finalRateRounding: MapFinalRateRounding(finalRateRounding),
+                    rateTreatment: rateTreatment to-enum RateTreatmentEnum,
+                    inflationLag: MapOffset(inflationLag),
+                    indexSource: MapStringWithScheme(
+                                indexSource -> value,
+                                indexSource -> rateSourcePageScheme
```
chrisisla, line 320: Just want to confirm this is the correct scheme, as I am not that familiar with FpML. If it is then all good 🙂

chrisisla (approved):
One very small question on a scheme but I'm pretty sure it will be the correct one.

Only other thing is that several more Rosetta files have been added to the PR which appear to only be formatting changes. I'm assuming these won't be included? Seeing as they are only formatting changes it is probably okay to include them.

Main change to the ingestion function looks good though, happy to approve.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-product-swap-func.rosetta
erincassidy1: Thanks @chrisisla! `scheme` in FpML is generally used to identify the source from which a `value` is taken. Where a `scheme` is available in the ingestion input, we map it to CDM using function `MapStringWithScheme`.

Regarding the formatting changes, when creating the release PR, we have an option to automatically format the PR so it aligns with CDM modelling conventions.

This functionality was temporarily broken in the master branch last week but has since been fixed. As a result, when I made the latest commit yesterday, the PR was reformatted as expected.

PR #4849, opened 2026-06-15
Oscar Twomey Asset - Moving Price Quantity Triangulation to PriceQuantity

### LINKED ISSUE #4109: Quantity x Price triangulation rule should apply whenever a quantity is specified as a no. units

regnosys-prod-user:


regnosys-prod-user:
# *Asset - Moving Price Quantity Triangulation to PriceQuantity*

_Background_

A `PriceQuantityTriangulation` condition is implemented at the `TradableProduct` level to check the internal consistency of the price and quantities provided. There are limitations to the existing implementation and the placement of the condition

_What is being released?_

- Move the condition to `PriceQuantity` so the check happens on the prices & quantities rather than on the trade lot
- Remove the `CashPriceQuantityNoOfUnitsTriangulation` function and place the logic directly in the `PriceQuantityTriangulation` function 
- Remove references to the unit type and apply the condition to a broader use case

_Review Directions_

Changes can be reviewed in PR: [#4849](https://github.com/finos/common-domain-model/pull/4849)

> [!NOTE]

PR #4854, opened 2026-06-16
Erin Cassidy Asset - Correcting AssetBase Conditions

regnosys-prod-user:

PR #4856, opened 2026-06-16
[master] - Erin Cassidy Asset - Correcting AssetBase Conditions (Updated Expectations)

### LINKED ISSUE #4794: AssetBase now makes PartyRole mandatory
### Issue Summary

A recent change to AssetBase includes two new attributes Party and PartyRole, both with (0..1) cardinality. The associated conditions make the PartyRole mandatory. For example, the condition IssuerRole is defined as
        _condition IssuerRole: <"If the asset type is a security then the role must be issuer.">
        if assetType = Security then partyRole -> role = Issuer_
If PartyRole is omitted from the Security object (e.g. as a Payout underlier), the constructed trade object fails validation because this condition cannot be evaluated.

If this is actually the intention, then Party and PartyRole should have cardinality (1..1). Otherwise, there should be an 'partyRole exists and ...' clause added to the conditions.

### Steps to Reproduce

None

### Expected Behaviour

If Party / PartyRole are omitted, there should be no validation errors.

### Actual Behaviour

Validation failure on constructed trade object where the underlying security does not have partyRole defined.

Validation FAILURE on [BusinessEvent.instruction(0).primitiveInstruction.execution.priceQuantity(0).observable.value.Asset.Instrument.Security] for [DATA_RULE] [AssetBaseIssuerRole] because [[Security->getPartyRole] cannot be compared to [AssetPartyRoleEnum] [Issuer]]

### Compatibility

No - new feature.

### Release

Dev master branch

### Additional Context

_No response_

regnosys-ops:
Update Expectation for PR #4854


CDM-ReleaseManagement-EC:
All validation failure differences are expected. Failures relating to `AssetBaseIssuerRole` have dropped from FpML ingestion test packs in this PR as a result of fixing the bug highlighted in issue https://github.com/finos/common-domain-model/issues/4794.

chrisisla (commented):
Change looks good @erincassidy1 , thank you for putting this PR together!

I saw your note on the Issue about the number of validation failures now decreasing, this makes sense.

The change you have made will fix the reported issue okay but I think there is another condition that could also cause a problem. If you look at `condition ExchangeListed` then I think that should also test for `partyRole exists`, otherwise `partyRole` will also be mandatory when `isExchangeListed` is true.

So perhaps:
```
        if isExchangeListed and partyRole exists then partyRole -> role = Exchange
```

erincassidy1:
Thanks for taking a look @chrisisla!

I've added your suggestion for `condition ExchangeListed`, it now reads: 

- `condition ExchangeListed: <"if isExchangeListed is true then partyRole is Exchange.">
if isExchangeListed and partyRole exists then partyRole -> role = Exchange`

I've also included a new condition `AssetPartyRoleExists` for a `TransferableProduct` as per @lolabeis and @2old2randr's comments on issue https://github.com/finos/common-domain-model/issues/4794: 

- `condition AssetPartyRoleExists: <"A TransferableProduct must contain an asset and a partyRole.">
asset ->> partyRole exists`

Can you please review again when you can?

chrisisla (approved):

Issue #4847, opened 2026-06-15
Trade time is never assigned


2old2randr:
### Issue Summary

Although both ExecutionInstruction and TradeState have a tradeTime attribute, the function Create_Execution never assigns the field when constructing the TradeState object.

The fix is simple - add two lines to the Create_Execution function:
    set execution -> trade -> tradeTime: <"Assign the tradeTime input to the execution object.">
        instruction -> tradeTime
 

### Steps to Reproduce

Create an ExecutionInstruction with the tradeTime attribute set and run Create_Execution

### Expected Behaviour

The output of Create_Execution (a TradeState object) should have the trade time populated with the value contained in the input ExecutionInstruction.

### Actual Behaviour

The tradeTime attribute of the output TradeState is null.

### Compatibility

I think it should be backward compatible since this is missing behaviour.

### Release

CDM 7.x.x but it can also probably be applied to previous versions.

PR #4861, opened 2026-06-17
Fix/assign trade time create execution

2old2randr:


2old2randr:
Fixed Create_Execution to assign trade time from execution instruction (issue #4847)

PR #4913, opened 2026-07-01
Exclude dev versions from PyPI

### LINKED ISSUE #4912: Exclude `dev` versions from PyPI
### Background

CDM currently publishes a release for every `dev` version. With this huge volume of releases (currently 130 for CDM 7), it's difficult to tell the versions apart, making it hard to choose which Python version to use. This will impede adoption and stability.

### Proposal

The proposal is to limit PyPI releases to those that are truly releases.  Skipping those labelled `dev` will reduce noise and limit confusion.  

Note that this would not preclude release of `alpha`, `beta`, and `release candidates` should CDM choose to adopt those labels.

### Compatibility

This has no effect on backwards compatibility since there will be no change to production releases.

### Release

CDM 6.x.x and 7.x.x.

### Additional Context

_No response_

dschwartznyc:
CDM currently publishes a release for every dev version. With this huge volume of releases (currently 130 for CDM 7), it's difficult to tell the versions apart, making it hard to choose which Python version to use. This will impede adoption and stability.

This PR will to limit PyPI releases to those that are truly releases. Skipping those labelled dev will reduce noise and limit confusion.

Note that this would not preclude release of alpha, beta, and release candidates should CDM choose to adopt those labels.

This has no effect on backwards compatibility since there will be no change to production releases.

The PR closes Issue #4912


dschwartznyc:
@llynhiavu @chrisisla @hugohills on the back of the TAWG discussion today, could one of you please review and approve?

The additional questions were captured in #4934

chrisisla (approved):
As agreed at the TAWG, we'll stop feeding dev releases to Pypi in order to (a) avoid clutter and (b) conform to the Pypi standard of only holding actual releases (i.e. major/minor) and not dev ones.

We have an Issue open to review this further and there are some other options on the table about how we can provide bundled dev releases (e.g. alpha, beta, release candidates).

If the CDM python community come back asking for the reintroduction of dev releases then we can reassess this decision.

Approving this change on that basis🙂

dschwartznyc:
@ReleaseManagement-BP is this ready for merge and to be part of the next 8.x.x dev release?

ReleaseManagement-BP:
@dschwartznyc Yes just merged this. Thanks for flagging, this will go into the next 8dev release

PR #4914, opened 2026-07-01
SettlementBase Refactor

### LINKED ISSUE #2213: Deprecate SettlementCenterEnum

tomhealey-icma:
This is a new PR to replace, https://github.com/finos/common-domain-model/pull/4742. A new PR was required because of problems with the prior merge and unwanted changes.

The last comments on this were:

We don't need a tradableSettlementCentre attribute, only the asset one is relevant
No need for new enum values in AncillaryRoleEnum
4 new enum values in AssetAncillaryPartyRoleEnum:
CashSettlementCentre
SecuritySettlementCentre
PlaceOfDeposit
Custodian
Rather than a multi-cardinality assetSettlementCentre, we need 2 optional attributes of single-cardinality each:
cashSettlementCentre, whose value should be CashSettlementCentre
securitySettlementCentre whose value should be SecuritySettlementCentre
Do not modify the snippets in the documentation for older versions of the model.
Remove the extraneous enum files in /resources

### DIFF
--- a/rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
+++ b/rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
@@ -372,11 +372,20 @@ type SettlementBase: <"A base class to be extended by the SettlementTerms class.
         [docReference ICMA GMRA namingConvention "Contractual Currency"
             provision "As defined in GMRA paragraph 2(k)/ paragraph 7(a) All the payments made in respect of the Purchase Price or the Repurchase Price of any Transaction shall be made in the currency of the Purchase Price (the Contractual Currency) save as provided in paragraph 10(d)(ii). Notwithstanding the foregoing, the payee of any money may, at its option, accept tender thereof in any other currency, provided, however, that, to the extent permitted by applicable law, the obligation of the payer to pay such money will be discharged only to the extent of the amount of the Contractual Currency that such payee may, consistent with normal banking procedures, purchase with such other currency (after deduction of any premium and costs of exchange) for delivery within the customary delivery period for spot transactions in respect of the relevant currency."]
     settlementDate SettlementDate (0..1) <"The date on which the settlement amount will be paid, subject to adjustment in accordance with any applicable business day convention. This component would not be present for a mandatory early termination provision where the cash settlement payment date is the mandatory early termination date.">
-    settlementCentre SettlementCentreEnum (0..1) <"Optional settlement centre as an enumerated list: Euroclear, Clearstream.">
     settlementProvision SettlementProvision (0..1) <"Optionally defines the parameters that regulate a settlement.">
     standardSettlementStyle StandardSettlementStyleEnum (0..1) <"Settlement Style.">
+    securitySettlementCentre AssetAncillaryPartyRoleEnum (0..1) <"Specifies the settlement centre of a security.">
+    cashSettlementCentre AssetAncillaryPartyRoleEnum (0..1) <"Specifies the settlement centre of cash.">
 
-//ICMA-P2
+    condition SecuritySettlementCentre:
+        if securitySettlementCentre exists
+          then   securitySettlementCentre = AssetAncillaryPartyRoleEnum -> SecuritySettlementCentre
+
+    condition CashSettlementCentre:
+        if cashSettlementCentre exists
+          then  cashSettlementCentre = AssetAncillaryPartyRoleEnum -> CashSettlementCentre
+
+
 type SettlementProvision: <"Defines parameters that regulate a settlement, for instance whether this settlement should be netted with other ones or broken-down into smaller amounts.">
     shapingProvisions ShapingProvision (0..1) <"Defines the parameters that are necessary to 'shape' a settlement, i.e. break it down into smaller amounts.">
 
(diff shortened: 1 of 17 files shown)


rosetta-source/src/main/rosetta/product-common-settlement-type.rosetta
```diff
@@ -372,11 +372,20 @@ type SettlementBase: <"A base class to be extended by the SettlementTerms class.
...
     settlementDate SettlementDate (0..1) <"The date on which the settlement amount will be paid, subject to adjustment in accordance with any applicable business day convention. This component would not be present for a mandatory early termination provision where the cash settlement payment date is the mandatory early termination date.">
-    settlementCentre SettlementCentreEnum (0..1) <"Optional settlement centre as an enumerated list: Euroclear, Clearstream.">
     settlementProvision SettlementProvision (0..1) <"Optionally defines the parameters that regulate a settlement.">
     standardSettlementStyle StandardSettlementStyleEnum (0..1) <"Settlement Style.">
+    securitySettlementCentre AssetAncillaryPartyRoleEnum (0..1) <"Specifies the settlement centre of a security.">
+    cashSettlementCentre AssetAncillaryPartyRoleEnum (0..1) <"Specifies the settlement centre of cash.">
+
+    condition SecuritySettlementCentre:
+        if securitySettlementCentre exists
+          then   securitySettlementCentre = AssetAncillaryPartyRoleEnum -> SecuritySettlementCentre
```
lolabeis: You don't need to use the enum name in the path, can simply write:
```
securitySettlementCentre = SecuritySettlementCentre
```

Same for cashSettlementCentre

lolabeis (commented):
Sending my comments here for audit trail, following my 1:1 exchanges with Tom:
 - Keep settlementCentre attribute as deprecated in CDM 7, we can remove it altogether in CDM 8
 - Need condition on payouts (at least SettlementPayout and OptionPayout) that if underlier is a security, then the 2 enum attributes should be populated (Note: that doesn't impose that the corresponding parties are populated in the asset itself)

Also see small inline comment.

tomhealey-icma:
I'll put settlementCentre back as [deprecated]. 

At this point I no longer see any need for securitySettlementCentre and cashSettlementCentre as they can be defined on the Asset and having securitySettlementCentre = SecuritySettlementCentre adds no useful information. 

I prefer to remove both attributes entirely.

lolabeis:
> At this point I no longer see any need for securitySettlementCentre and cashSettlementCentre as they can be defined on the Asset and having securitySettlementCentre = SecuritySettlementCentre adds no useful information.
> 
> I prefer to remove both attributes entirely.

@tomhealey-icma I understand your point, but I think the attributes still have their use as a referencing mechanism.

Say an implementor is building a Transfer-generating mechanism from Payout: for, instance a SettlementPayout for an underlying Security that spins out a Cash and a Security Transfer in opposite directions. They would want to attach the respective Settlement Centre (taken from the asset) to each Transfer. Without the enum attributes, that implementor would need to know that if they're generating a Security Transfer, they need to fetch the party that's associated to the SecuritySettlementCentre value in the asset. While that may sound obvious to a subject matter expert, it would be based on implicit knowledge.

By contrast if the enum value was associated to a specific `securitySettlementCentre` attribute, then the referencing mechanism to get that settlement centre becomes explicit and leads the implementor to the party associated to the SecuritySettlementCentre value in the underlying asset.

ReleaseManagement-BP:
@tomhealey-icma Comment added by @lolabeis on the main issues following the CRWG, please see https://github.com/finos/common-domain-model/issues/2213

ReleaseManagement-BP:
Hey @tomhealey-icma I can see that:

'FunctionInputCreationTest.testBondExecutionInput' 

is failing from your latest commit. Is this the test case that needed updating that we spoke about on the CRWG? Are you able to resolve this, or do you need help?

lolabeis (approved):
Noted the increase in validation failure, because the attributes aren't populated for some security-related trades. Will look to fix that in samples in later versions.

ReleaseManagement-BP:
# *Settlement Centre Refactoring and Party-Based Settlement Centre Support*

_Background_

Prior versions of the CDM SettlementBase included an enum, SettlementCentre that only supported two enums using a separate enum, SettlementCentreEnum. SettlementCentre is primarily used by securities and assets like bonds and equities. Once a settlement centre is defined at issuance, it does not change except by corporate action. Settlement centres should be defined using a party reference that includes and LEI, and it's also necessary to separate settlement centres for security and cash.

_What is being released?_

This PR refactored SettlementBase by adding new attributes, securitySettlementCentre and cashSettlementCentre, both are AssetAncillaryPartyRoleEnum with new enums values. Use of these attributes requires adding the settlement centre party to the asset.

_Review Directions_

Changes can be reviewed in PR: [#4914](https://github.com/finos/common-domain-model/pull/4914)

PR #4928, opened 2026-07-07
[6.x.x] - feat: #4774 Ingestion mapping for FpML executionNotification to CDM ExecutionInstruction

### LINKED ISSUE #4774: Missing ingestion mapping for FpML executionNotification to CDM ExecutionInstruction
### Issue Summary

The CDM Ingestion Engine currently fails to correctly map FpML executionNotification messages representing the initial stage of a trade lifecycle. Because an initial execution lacks a prior trade state, it should be encapsulated in a WorkflowStep containing a proposedEvent with an ExecutionInstruction. However, the engine currently maps a tradeState into a proposedEvent without primitiveInstruction. It also defaults to qualifying these messages as a Contract Formation event due to a lack of dedicated routing and mapping functions for initial execution events. 

### Steps to Reproduce

1. Feed an FpML executionNotification message (representing a brand-new trade execution with no prior state) into the CDM Ingestion Engine
2. Target the WorkflowStep output to capture the initial trade lifecycle stage
3. Observe the generated CDM object structure and subsequent event qualification

### Expected Behaviour

The ingestion engine (translate functions) should recognize the executionNotification as the start of the event chain and route it into a workflowStep that contains:
- `proposedEvent -> instruction -> executionInstruction`
and doesn’t contain a tradeState. After that, the event must qualify as an execution.

This requires the following engineering updates in the mapping functions:
- `MapWorkflowStep` modification to populate `proposedEvent -> instruction -> execution`
- New function - Implement `MapTradeToExecutionInstruction` to bridge the FpML block directly to the CDM `ExecutionInstruction` object

### Actual Behaviour

The ingestion engine lacks the necessary mapping functions to route the data without a pre-existing state reference. Instead of generating an `ExecutionInstruction`, it routes the executionNotification to a `WorkflowStep `where the `proposedEvent `is incorrectly qualified as Contract Formation and contains the trade. 

Image

This is the current translate function that needs to be modified:

Image

### Compatibility

This fix is backward-compatible:
- It introduces a brand-new entry point and routing logic for a previously unsupported lifecycle stage
- It should not alter existing mapping behaviors for executionAdvice or other workflows that successfully result in Contract Formation

### Release

CDM 6 prod and 7 dev

### Additional Context

_No response_

gpan-isda:
Added a new Execution intent to follow the workflow step to skip the before Trade State in case of an execution notification FpML message.


gpan-isda:
# *CDM Model: FpML executionNotification ingestion mapping to CDM ExecutionInstruction*

_What is being released_

An ingestion mapping has been added for the FpML `executionNotification` message so that its `Trade` element is now mapped onto a CDM `ExecutionInstruction`, in place of the previous mapping which produced a `ContractFormationInstruction`. This aligns the outcome of ingesting an `executionNotification` with the semantics of a newly executed trade, rather than a contract formation event that requires a pre-existing trade state.

To support this, a new enumerated value `Execution` has been added to the `EventIntentEnum` in the `cdm.event.common` namespace, with the description *"The intent is to execute a new trade."* This new intent is used by the ingestion functions to identify that the incoming FpML message represents a new trade execution.

The following ingestion functions have been updated in the `cdm.ingest.fpml` namespace:

- `MapExecutionNotificationToWorkflowStep` (in `ingest-fpml-confirmation-message-func.rosetta`) — now remaps the resolved FpML intent to `Execution` when the raw intent is `ContractFormation` or absent, so that `executionNotification` messages are routed through the new execution path.
- `MapWorkflowStep` (in `ingest-fpml-confirmation-workflowstep-func.rosetta`) — when `intent = Execution`, the workflow step's `Instruction` is now built with an `execution` `PrimitiveInstruction` (and no `before` `TradeState`), instead of the generic primitive instruction with a `before` state.
- A new function `MapTradeToExecutionInstruction` has been added to map an FpML `Trade` (together with the party and quotation lists) directly to a CDM `ExecutionInstruction`, populating `product`, `priceQuantity`, `counterparty`, `ancillaryParty`, `parties`, `partyRoles`, `executionDetails`, `tradeDate` and `tradeIdentifier`.

Expected output JSON files for existing FpML confirmation and record-keeping samples that exercise `executionNotification` messages have been regenerated to reflect the new `ExecutionInstruction`-based output.

_Review directions_

In the CDM Portal open the Textual Browser, search for `EventIntentEnum` in the `cdm.event.common` namespace and observe the new `Execution` value. Then search for `MapExecutionNotificationToWorkflowStep`, `MapWorkflowStep` and `MapTradeToExecutionInstruction` in the `cdm.ingest.fpml` namespace to review the updated ingestion logic. Ingesting any of the updated FpML `executionNotification` sample files (e.g. `pkg-ex01-pkge-execution-notification`, `pkg-ex55-execution-notification`, `msg-ex19-cds-execution-allocations`) will now produce a `WorkflowStep` whose `Instruction` contains an `execution` `PrimitiveInstruction` rather than a `contractFormation` one.

Inspect Pull Request: [#4928](https://github.com/finos/common-domain-model/pull/4928) and [#5061](https://github.com/finos/common-domain-model/pull/5061)

Inspect Issue: [#4774](https://github.com/finos/common-domain-model/issues/4774)

ReleaseManagement-BP:
Hi @hugohills-regnosys @manel-martos please could you review this?

docs/event-model.md
```diff
@@ -865,6 +865,7 @@ enum EventIntentEnum:
    CreditEvent
    Decrease
    EarlyTerminationProvision
+   Execution
```
hugohills-regnosys: I don't think we should be adding an intent just for mapping if it wasn't previously necessary because that's more of a modeling decision.

hugohills-regnosys: I see that adding `Execution` to the `EventIntentEnum` was clearly stated in the accompanying issue, [#4774](https://github.com/finos/common-domain-model/issues/4774#issuecomment-4659647104), and that was approved by the CRWG. My only concern is that it is used only for mappings, and is not required in the model, for example in qualification functions.  Maybe it's fine.  @manel-martos what is your view?

manel-martos: @hugohills-regnosys , we are investigating this. My initial concern with removing the `Execution` intent is that `MapIntent` may no longer provide sufficient granularity to distinguish between `Execution` and `ContractFormation`, and therefore may not populate the required `Execution` event correctly. This distinction is necessary for our event processing use case within the Translate 2.0 functionality.

We will discuss this internally and assess whether raising a separate issue to address the gap in `MapIntent` would provide a more generic solution.

manel-martos: @hugohills-regnosys - The mapping requirements are defined in the issue but we are dicussing again how some FpML messages (+ optional features) should map into the CDM event structures. We aim at building a proposal which only involves FpML Translate 2.0 upgrades avoiding the creation of the enum value.

ReleaseManagement-BP: @manel-martos is the plan to still have this new proposal ready for friday's release? Or shall I move this to the 6.26.0 release?

manel-martos:
Hi @gpan-isda ,

Thanks for implementing the requested `executionNotification` mapping.

Following a further internal review of the proposal, including [the discussion](https://github.com/finos/common-domain-model/pull/4928#discussion_r3585736430) with @hugohills-regnosys , we would like to suggest a small revision before approval.

The current approach adds `Execution` to `EventIntentEnum`, but this value is used only as an internal routing mechanism within the FpML ingestion logic. As `EventIntentEnum` is part of the CDM model, our preference would be to avoid extending it solely to support translation-specific control flow where the appropriate instruction can instead be determined directly from the source FpML content.

For the scope of this contribution, we propose removing the new `Execution` enum value and applying the following source-data-based mapping:

* An `executionNotification` without legal documentation should map to an `ExecutionInstruction`.
* An `executionNotification` with legal documentation should continue to map to a `ContractFormationInstruction`, preserving the existing behaviour.

Based on our understanding of the FpML message semantics, the absence of legal documentation in an `executionNotification` indicates the initial execution of a new trade. Where legal documentation is present, the existing contract-formation path should remain applicable.

The new `MapTradeToExecutionInstruction` function can remain and be invoked directly for an `executionNotification` where legal documentation is absent.

The intent-resolution logic in `MapIntent` may also need to be adjusted so that the current default scenario is handled correctly, while preserving the existing behaviour for other message types.

Please let us know if it would be helpful to discuss the proposed routing logic together.

manel-martos (changes requested):
Request defined in https://github.com/finos/common-domain-model/pull/4928#issuecomment-5069115687

manel-martos (dismissed):
Thanks @gpan-isda for the upgrades.
@hugohills-regnosys , the model looks cleaner to me. Could you please verify the translate logic around legal documentation presence?

gpan-isda:
Thank you @manel-martos. However I do think now that this needs a change as the function signature has changed to include an optional parameter and may constitute as a breaking change.

rosetta-source/src/main/resources/ingest/output/fpml-confirmation-to-workflow-step/fpml-5-10-processes/msg-package-price.json
```diff
@@ -1,670 +1,638 @@
 {
   "proposedEvent" : {
-    "intent" : "Clearing",
```
manel-martos, line 3: We acknowledge that the `intent` flags may not map directly between FpML and CDM. However, the changes affecting both samples under `fpml-5-10-processes` remain a blocker to accepting the contribution in its current form. Our current position is that both expected outputs should keep the flag unchanged, but we propose bringing this point to the WG for further discussion and alignment.

gpan-isda: Hi @manel-martos Thanks for the observation. I have reinstated the intent flag on these tests but please feel free to discuss this in WG.

manel-martos: The new expectation has been successfully tested against the DRR reporting pipeline retrieving the expected Action-Event type tuple. We consider this comment resolved.

tests/src/test/resources/ingest/diagnostics/product_diagnostics.md
manel-martos, line 1: @hugohills-regnosys - Not sure how these diagnostics files are maintained. Could you accept/resolve this expectation change?

manel-martos (dismissed):

tests/src/test/resources/ingest/diagnostics/product_diagnostics.md
ReleaseManagement-BP: @Chuks-Ajeh While Hugo is OOO is this something you'd be able to help with?

ReleaseManagement-BP: @gpan-isda / @manel-martos Chuks was able to look at this, and the reason these were not generated is because of the way @gpan-isda ran update expectations. So could you please run the following command to resolve the issue:

'mvn clean install  -P format,update-expectations'

manel-martos (approved):

tests/src/test/resources/ingest/diagnostics/product_diagnostics.md
hugohills-regnosys: When the above mvn update-expectations command was run, it looks like some of the diagnostics files did not get committed.  I have regenerated them in a separate PR with an explanation in the PR description, see https://github.com/finos/common-domain-model/pull/5061

Issue #4379, opened 2026-01-23
CDM Smart Contract Step 2: Determine the Unadjusted Calculation Period


lukebanas:
### Background

Following Step 1 of extracting the Floating Rate Index, the next requirement is to establish the temporal boundaries of the reset. This issue introduces the next executable stage in the Reset Smart Contract process: **Smart Contract Step 2 - Determine the Unadjusted Calculation Period**.

In the ISDA 2021 Interest Rate Derivatives Definitions, the Calculation Period constitutes the period of time over which interest accrues for the purpose of calculating a payment amount. While the CDM currently allows for the representation of these dates within the *CalculationPeriodDates* structure, it does not contain an executable logic to derive them deterministically.

**Note**: In this proposal we use the term “unadjusted” in conjunction with dates or periods that have not yet been adjusted by any applicable business day convention(s) or calculation methodology(ies).

To arrive at the desired Calculation Period, the following high-level steps should be followed:

1.  Generate a list of all the Unadjusted Calculation Periods for a given Trade
2.  Return the Unadjusted Calculation Period containing the unadjusted Reset Date.

_Note: The Unadjusted Reset Date is inherent to the Reset Smart Contract Process. It is known prior to beginning the Smart Contract because it is the key trigger that initiates it. Therefore, this input is not present in the Workflow Step, but instead added externally by an outside implementor. Given this Unadjusted Reset Date, after calculating a list of all Unadjusted Calculation Periods, the Unadjusted Calculation Period containing the Unadjusted Reset Date is returned as the output of this step._

### Proposal

**Objective**

Implement **Smart Contract Step 2 - Determine the Unadjusted Calculation Period** as an executable Smart Contract Step within the Reset Instruction Smart Contract. The step must:

*   Construct the schedule of Unadjusted Calculation Periods based on the *CalculationPeriodDates* CDM structure specified in the Confirmation.
*   Choose and return the relevant Unadjusted Calculation Period given the Unadjusted Reset Date specified for this Smart Contract, as well as the Unadjusted Reset Date.

**Scope**

To implement an executable Smart Contract Step.

*   **Step Name**: DetermineUnadjustedCalculationPeriod
*   **Instruction Function:** _Create_DetermineUnadjustedCalculationPeriodInstruction_
*   **Purpose:**
    *   Receive the following:
        *   The Calculation Period Dates from economic terms in the CDM Workflowstep
        *   The Unadjusted Reset Date
[...]

PR #4993, opened 2026-07-23
Instruction Composition Reset Step 3: Adjust the period

### LINKED ISSUE #4389: CDM Composition Step 3: Adjust the Period
### Background

The purpose of this issue is to implement **Composition Step 3 - Adjust the (Calculation) Period**. Please refer to previous issues for an explanation of the Instruction Composition process.

This is the third Composition Step in the Interest Rate Reset Instruction Composition process, the purpose of which is to check that the start and end dates of the Calculation Period do not fall on non business days when the interest cannot be paid, and adjust them to good business days, thus persisting the Adjusted Calculation Period.

#### **Objective**

Implement **Composition Step 3 - Adjust Calculation Period** as an executable Composition Step within the Reset Instruction Composition. The step must:

*   Obtain the list of non-business days corresponding to the Unadjusted Calculation Period from Step 2 by using a ‘placeholder function’ that fetches these dates given a business center and a period of time. 
    
*   Adjust the Unadjusted Calculation Period from Step 2 given a specified business convention.
    

### **Scope**

To implement an executable Composition Step. To do this, the process is broken down into two parts:

1.  Use of the placeholder function

*   **Function Name:** _GetNonBusinessDates_
*   Takes either a list of one or more dates or a period, along with the applicable business centre to provide a list of non-business days to use for adjusting.
*   This function will be addressed in a separate issue

2.  Creating the Instruction

*   **Step Name:** AdjustPeriod
*   **Instruction Function:** _Create_AdjustPeriodInstruction_
*   **Purpose:** 
    *   Receive the following:
        *   The Unadjusted Calculation Period
        *   The non-business days applicable to the Calculation Period
        *   The Business Day Convention for the Calculation Period
    *   Using this information, adjust the Calculation Period.
    *   Output the resulting adjusted period in an _AdjustPeriodInstruction_ type 
        

#### **Inputs** to _**Create_AdjustPeriodInstruction**_

*   The **Unadjusted Calculation Period**
*   The **Calculation Period Non-business Days**, taken from the placeholder function by using the business centers specific to the Unadjusted Calculation Period
*   The **Calculation Period Business Day Convention**, extracted from the Workflow Step
    

### **Process**

1.  The placeholder function is used to provide a list of non-business days (for a given business center) applicable to the Period
2.  The output of the above function is the list of non-business days that occur within the Unadjusted Calculation Period from Step 2. This is part of the input to the creation of the Instruction, detailed next.
3.  The instruction is created using the _Create_AdjustPeriodInstruction_ function with the below input:
    
    *  unadjustedPeriod CalculationPeriodBase (1..1)
    *  nonBusinessDays date (0..*)
    *  businessConvention BusinessDayConventionEnum (1..1)
        
4.  Given these inputs, the Calculation Period (start and end date) is adjusted as per the  applicable business day convention:
    
    *  If “Following”, then the non-business dates are adjusted to the first following day that is a business day.
    *  If “ModFollowing”, then the non-business dates are adjusted to the first following day that is a business day unless that day falls in the next calendar month, in which case that date will be the first preceding day that is a business day.
    *  If “Preceding”, then the non-business dates are adjusted to the first preceding day that is a business day.
        
5.  **The resulting Adjusted Calculation Period is finally returned as the output of this instruction**
    

Image

### Compatibility

### Release

TBC

### Additional Context

CDM Smart Contract (Instruction Composition) Task Force

SuperMonanana:
# _Instruction Composition Reset Step 3: Adjust the period_

_Background_

Building upon the functional milestones introduced in the previous release for Step 2 (#4379), the CDM Smart Contract Taskforce is advancing the state-machine execution capabilities within the Instruction Composition Framework. This release delivers the components required to execute Step 3 (Adjust Calculation Period) of the 8-step Interest Rate Reset processing workflow. By introducing this phase, the model can now deterministically apply calendar adjustments and financial business day conventions to unadjusted schedules. For further information, see issue #4389.

_What is being released?_

This release provides the core business day alignment logic, holiday calendar resolution hooks, and state accumulator mutations necessary to transition an unadjusted calculation period into its adjusted lifecycle state. It incorporates payload bindings for Step 3 into the reusable instruction containers and expands the historical overlay tracking mechanisms to capture the final adjusted period bounds.

_Functions_

Changes in the cdm.base.datetime namespace:
- `AdjustDateToBusinessDayConvention`: Core mapping function that routes an unadjusted date through a selected business day convention matrix when a calendar clash is detected.
- `ApplyFollowing`: Recursive utility evaluating calendar forward-shifts to find the next valid business day.
- `ApplyPreceding`: Recursive utility evaluating calendar backward-shifts to find the previous valid business day.
- `ApplyModFollowing`: Evaluation function ensuring forward-shifted dates do not cross month boundaries, defaulting to preceding lookbacks where necessary.

- `Create_AdjustPeriodInstruction`: An operational mapping function that compiles adjusted starting and ending coordinates for a calculation period using defined business conventions and holiday datasets.
- `GetNonBusinessDates`: An externalized code-implementation hook to query and extract non-business dates from financial calendar providers using either distinct date vectors or defined boundary periods.

- `UpdateResetCompositionState`: (Updated) Expanded to implement "Step 3" overlay logic. It conditionally captures the incoming adjustPeriod payload to update the active adjustedCalculationPeriod parameter.

_Types and Choices_

- `CompositionStepInstructions` (Updated): Expanded to include adjustPeriod, allowing the global step orchestrator to wrap period adjustment data structures.

- `AdjustPeriodInstruction`: Introduced as the official type for Step 3.
- `ResetInstructionState` (Updated): Expanded to hold its third phase state variable (adjustedCalculationPeriod), allowing the cumulative workflow to track adjusted parameters over its lifecycle.

_Review Directions_

Navigate to the following paths to inspect the new models:
- rosetta-source/src/main/rosetta/instructionComposition-type.rosetta: Review the updated CompositionStepInstructions container to verify the inclusion of the Step 3 instruction payload mapping.
- rosetta-source/src/main/rosetta/instructionComposition-func.rosetta: Examine the updated field overlay logic inside UpdateResetCompositionState handling adjusted period state tracking.
- rosetta-source/src/main/rosetta/instructionComposition-reset-type.rosetta: Inspect the new AdjustPeriodInstruction type structure and the updated tracking properties embedded within ResetInstructionState.
- rosetta-source/src/main/rosetta/instructionComposition-reset-func.rosetta: Inspect the schedule compilation logic inside Create_AdjustPeriodInstruction and the GetNonBusinessDates interface.

### DIFF
--- a/rosetta-source/src/main/rosetta/base-datetime-func.rosetta
+++ b/rosetta-source/src/main/rosetta/base-datetime-func.rosetta
@@ -394,3 +394,69 @@ func AdjustableDatesResolution: <"Prioritization of unadjustedDate over adjusted
         if adjustableDates -> unadjustedDate exists
         then adjustableDates -> unadjustedDate
         else adjustableDates -> adjustedDate
+
+func AdjustDateToBusinessDayConvention: <"Applies a specific business day convention to a single date if it's a non-business day and adjusts it. Note: Currently, only FOLLOWING, PRECEDING, and MODFOLLOWING conventions are implemented. Other conventions will raise an error indicating they are not yet supported.">
+    inputs:
+        unadjustedDate date (1..1)
+        nonBusinessDates date (0..*)
+        businessConvention BusinessDayConventionEnum (1..1)
+    output:
+        adjustedDate date (1..1)
+
+    condition SupportedConvention: <"Only FOLLOWING, PRECEDING, MODFOLLOWING, and NONE are currently implemented. Unsupported conventions are blocked to prevent returning silently unadjusted dates.">
+        businessConvention = BusinessDayConventionEnum -> FOLLOWING
+            or businessConvention = BusinessDayConventionEnum -> PRECEDING
+            or businessConvention = BusinessDayConventionEnum -> MODFOLLOWING
+            or businessConvention = BusinessDayConventionEnum -> NONE
+
+    set adjustedDate:
+        if (nonBusinessDates contains unadjustedDate) = False
+        then unadjustedDate
+        else businessConvention switch
+            FOLLOWING then
+                AdjustDateToFollowingBusinessDay(unadjustedDate, nonBusinessDates),
+            PRECEDING then
+                AdjustDateToPrecedingBusinessDay(unadjustedDate, nonBusinessDates),
+            MODFOLLOWING then
+                AdjustDateToModfollowingBusinessDay(unadjustedDate, nonBusinessDates),
+            default unadjustedDate
+
+func AdjustDateToFollowingBusinessDay: <"Finds the next available business day by moving forward in time.">
+    inputs:
+        unadjustedDate date (1..1)
+        nonBusinessDates date (0..*)
+    output:
+        nextBusinessDay date (1..1)
+
+    set nextBusinessDay:
+        if (nonBusinessDates contains unadjustedDate) = False
+        then unadjustedDate
+        else AdjustDateToFollowingBusinessDay(AddDays(unadjustedDate, 1), nonBusinessDates)
+
+func AdjustDateToPrecedingBusinessDay: <"Finds the previous available business day by moving backward in time.">
+    inputs:
+        unadjustedDate date (1..1)
+        nonBusinessDates date (0..*)
+    output:
+        previousBusinessDay date (1..1)
+
+    set previousBusinessDay:
+        if (nonBusinessDates contains unadjustedDate) = False
+        then unadjustedDate
+        else AdjustDateToPrecedingBusinessDay(AddDays(unadjustedDate, -1), nonBusinessDates)
+
+func AdjustDateToModfollowingBusinessDay: <"Applies the 'Following' convention, but reverts to 'Preceding' if the adjustment crosses a month boundary.">
+    inputs:
+        unadjustedDate date (1..1)
+        nonBusinessDates date (0..*)
+    output:
+        adjustedDate date (1..1)
+
+    alias followingDate: AdjustDateToFollowingBusinessDay(unadjustedDate, nonBusinessDates)
+
+    alias IsDifferentMonth: (followingDate -> month) <> (unadjustedDate -> month)
+
+    set adjustedDate:
+        if IsDifferentMonth
+        then AdjustDateToPrecedingBusinessDay(unadjustedDate, nonBusinessDates)
+        else followingDate
--- a/rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
+++ b/rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
@@ -1,6 +1,8 @@
 namespace cdm.event.instructioncomposition.reset
 version "${project.version}"
 
+import cdm.base.datetime.*
+import cdm.base.staticdata.codelist.*
 import cdm.event.instructioncomposition.*
 import cdm.product.asset.*
 import cdm.product.common.schedule.*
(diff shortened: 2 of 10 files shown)


ReleaseManagement-BP:
Hi @tomhealey-icma @llynhiavu @chrisisla please can you review this step when you have time? Thank you

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
```diff
@@ -348,3 +348,63 @@ func AdjustableDatesResolution: <"Prioritization of unadjustedDate over adjusted
...
+
+func AdjustDateToBusinessDayConvention: <"Applies a specific business day convention to a single date if it's a holiday and adjusts it.">
+    inputs:
+        unadjustedDate date (1..1)
+        holidayDates date (0..*)
+        businessConvention BusinessDayConventionEnum (1..1)
+    output:
+        adjustedDate date (1..1)
+
+    set adjustedDate:
```
chrisisla, line 412: So this is only supporting 3 of the business day conventions for now, and for the ones that it does not support it will just return the unadjusted date. Although not ideal, I do understand that for your use case you may not need the other conventions, but I think it would be worth at least adding a comment here to explain this, as users will be getting back the `unadjusted` date believing that it has actually been adjusted for these missing conventions.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
```diff
@@ -46,6 +48,54 @@ func Create_DetermineUnadjustedCalculationPeriodInstruction: <"This function gen
     set determineUnadjustedCalculationPeriodInstruction -> unadjustedResetDate: <"Sets the unadjusted reset date to the output.">
         unadjustedResetDate
 
+// Code Implementation Function - External call to fetch the list of non business dates
+func GetNonBusinessDates: <"This function, given the list of business centers applicable and a period or a list of dates, extracts the list of non business dates for the given period or list of dates and returns the corresponding list of non business dates. The function should receive as input a list of dates or a period, but not both at the same time. That is, given either a list of periods or a period (start date + end date), the function will return their non business days.">
```
chrisisla, line 52: From the notes this `[code implementation]` function is a hook to allow implementors to use their own code, is that correct? Are we looking to provide a default implementation? If not then I don't think this function will work as is, because we're not building `nonBusinessDates` anywhere. Not sure on how to proceed with this.

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
```diff
@@ -348,3 +348,63 @@ func AdjustableDatesResolution: <"Prioritization of unadjustedDate over adjusted
         if adjustableDates -> unadjustedDate exists
         then adjustableDates -> unadjustedDate
         else adjustableDates -> adjustedDate
+
+func AdjustDateToBusinessDayConvention: <"Applies a specific business day convention to a single date if it's a holiday and adjusts it.">
+    inputs:
+        unadjustedDate date (1..1)
+        holidayDates date (0..*)
```
chrisisla: May want to rename this from `holidayDates` to `nonBusinessDates` to match the naming in the other top level functions? It depends on whether this really is just focusing on "holidays" like Eid, Christmas or Bank Holidays, or whether it is actually about any day that is not a business day e.g. Sunday (in some regions 😉) . If you do rename the input parameter then it is worth doing for all these functions I reckon 👍

chrisisla (commented):
Some questions for now please 🙂

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
lagarciath: This was part of the initial scope for the TF, so it only supports these three conventions for now. We'll add an inline comment explaining that unsupported conventions fallback to returning the unadjusted date so users aren't caught off guard

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
lagarciath: Sure

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: As previously agreed during the TF meetings, this function is designed strictly as an infrastructure hook. It gives implementors a continuous pipeline to use with SmartContract/InstructionEnvelope.

We aren't providing a default implementation here, just the framework so external systems can plug in their own logic (similar to how date adjustments connect to an external system)

chrisisla: Okaydoke. So if someone were to try to call this function now (i.e. without implementing their own `GetNonBusinessDates` code implementation) what would happen?

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
SuperMonanana: We've added an inline comment and also a condition, so the user will get an exception with clear message if they put the unimplemented convention

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
SuperMonanana: Done

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: If someone calls this function as-is without a custom `GetNonBusinessDates` implementationm the function will return **null (or empty)**, which means the input required for the subsequent step won't exist, effectively halting that specific execution path.

Business day calendars (holidays, weekends, regional variations) are external data dependencies that fall outside the scope of CDM core data. Therefore, providing this information is the responsibility of the implementing system. This function serves as an infrastructure interface, it's included in this contribution to demonstrate a complete end-to-end flow. However, implementors can either plug in their own logic here or bypass this function entirely by sourcing the non-business dates through their own preferred mechanism before passing them into the next step of the process.

chrisisla (dismissed):
Small point on the Java - but IMO nothing that should impede this being approved - is that the function names and variables are still referring to holidays rather than non-business days. Not critical at all 👍

SuperMonanana:
@chrisisla Good spot, thanks! Renamed in Java tests as well

SuperMonanana:
7.x.x PR: https://github.com/finos/common-domain-model/pull/5057

chrisisla (dismissed):

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
```diff
@@ -348,3 +348,69 @@ func AdjustableDatesResolution: <"Prioritization of unadjustedDate over adjusted
...
+        else businessConvention switch
+            FOLLOWING then
+                ApplyFollowing(unadjustedDate, nonBusinessDates),
+            PRECEDING then
+                ApplyPreceding(unadjustedDate, nonBusinessDates),
+            MODFOLLOWING then
+                ApplyModFollowing(unadjustedDate, nonBusinessDates),
+            default unadjustedDate
+
+func ApplyFollowing: <"Finds the next available business day by moving forward in time.">
```
llynhiavu: Is the name of the function ApplyFollowing() explicit enough on its own? 

it makes sense within the context of func AdjustDateToBusinessDayConvention

This is in core namespace after all. So, a developer looking at the list of functions, may not immediately understand what e.g. ApplyFollowing does. 

Suggestions:
ApplyBusinessDayConventionFollowing or 
AdjustDateToBusinessDayConventionFollowing
AdjustDateToFollowingBusinessDay (slight preference for this)

I don't have a strong objection (I can see that clarifying the function name also clutters it at the same time) but raising the comment anyway for feedback from others.

Same comment applies to other func ApplyPreceding, ApplyModFollowing

e,g, AdjustDateToPrecedingBusinessDay, AdjustDateToModFollowingBusinessDay

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
```diff
@@ -348,3 +348,69 @@ func AdjustableDatesResolution: <"Prioritization of unadjustedDate over adjusted
...
+        nonBusinessDates date (0..*)
+    output:
+        nextBusinessDay date (1..1)
+
+    set nextBusinessDay:
+        if (nonBusinessDates contains unadjustedDate) = False
+        then unadjustedDate
+        else ApplyFollowing(AddDays(unadjustedDate, 1), nonBusinessDates)
+
+func ApplyPreceding: <"Finds the previous available business day by moving backward in time.">
```
llynhiavu: See similar comment R378

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
```diff
@@ -348,3 +348,69 @@ func AdjustableDatesResolution: <"Prioritization of unadjustedDate over adjusted
...
+        nonBusinessDates date (0..*)
+    output:
+        previousBusinessDay date (1..1)
+
+    set previousBusinessDay:
+        if (nonBusinessDates contains unadjustedDate) = False
+        then unadjustedDate
+        else ApplyPreceding(AddDays(unadjustedDate, -1), nonBusinessDates)
+
+func ApplyModFollowing: <"Applies the 'Following' convention, but reverts to 'Preceding' if the adjustment crosses a month boundary.">
```
llynhiavu: See similar comment R378

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
SuperMonanana: Thanks for the suggestion! I've renamed these three functions

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
SuperMonanana: Fixed

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
SuperMonanana: Fixed

llynhiavu (approved):
renamed functions are more intuitive now.

ReleaseManagement-BP:
Hey @SuperMonanana please could you update this branch please? Once it's up to date I can merge this in

SuperMonanana:
@ReleaseManagement-BP Done!

PR #5000, opened 2026-07-27
Add FpML coding scheme links to codelist aliases

### LINKED ISSUE #3929: Migration of CDM Reference Data to Lists Prevents Modeler Access
### Issue Summary

The upcoming migration of some reference data from ENUMs to externally imported lists will make it more difficult for modelers to access these values within CDM's constraints and functions. Additionally, as these lists are updated to reflect external changes, inconsistencies may be introduced into those conditions and functions.

### Steps to Reproduce

As described in issue [#3512](https://github.com/finos/common-domain-model/issues/3512), the intention behind this change (see PR [#3819](https://github.com/finos/common-domain-model/pull/3819)) is to improve CDM’s synchronization with external reference data. The implementation moves reference data from ENUMs, which are part of the model, to externally defined lists, which are not. As a result, the migrated reference data will be no longer directly visible in the model and may change without notice when external sources are updated.

This lack of visibility will make it difficult to use migrated reference data to define different model behaviors. For example, with the planned migration of BusinessCenterEnum, it would be challenging to define specific behavior based on a business centers such as "AEAB" (Abu Dhabi) or "AUSY" (Sydney) if these values are not visible in the model. Additionally, keeping business logic up to date could be difficult if the external lists change, especially if a referenced value is removed. This could lead to inconsistencies in the CDM model; for instance, if a function refers to "AEAB" and it is removed from the external list, the function may fail.

### Expected Behaviour

Reference data values should be easily visible to modelers, whether using modeling or coding tools, and consistency checks should be performed before any updates.

### Actual Behaviour

After migration, reference data is no longer visible in the model. It is unclear whether consistency checking is performed as part of data updates.

### Compatibility

Implementation of the PR is a breaking change.

### Release

Assume that this change is intended for CDM 7.

### Additional Context

Updating rosetta-technology.io could improve the visibility of reference data updates.

SaaiAravindhRaja:
## Pull Request Summary ##

Adds FpML Coding Schemes links to the migrated FpML code-list type aliases in `base-staticdata-codelist-type.rosetta`, so modelers can open the source scheme definitions while the values remain externally managed.

Issue: https://github.com/finos/common-domain-model/issues/3929

## Content ##
**This Pull Request contains (tick each list item once completed):**

- [x] A link to an Issue describing the work, background and any associated documentation
- [ ] Been reviewed by one of the following Working Groups:
  - [ ] Contribution Review Working Group (CRWG)
  - [ ] Steering Working Group (SWG)
  - [ ] Technology Architecture Working Group (TAWG)
  - [ ] A CDM Domain Working Group maintained by FINOS
  - [ ] An External Industry Association CDM Working Group e.g. run by ISLA, ISDA, ICMA
- [ ] Been assessed for compliance with CDM Design Principles (where applicable)
- [ ] A Release Note that describes the changes, which contains:
  - [ ] A section explaining the business or technical reasoning for the change
  - [ ] A section detailing the changes themselves
  - [ ] A section on any changes that break backwards compatibility
  - [ ] A link to this Pull Request as a cross reference

Unchecked items can be completed during review if the working group wants the change tracked as a formal specification update. This PR follows the lightweight visibility improvement suggested by maintainers and does not change validation behavior.

## Supporting Documentation ##

The added comments point to the current FpML Coding Schemes page section anchors for each canonical scheme. `InformationProvider` uses the `information-provider` scheme section, not the separate commodity information provider section.

### DIFF
--- a/rosetta-source/src/main/rosetta/base-staticdata-codelist-type.rosetta
+++ b/rosetta-source/src/main/rosetta/base-staticdata-codelist-type.rosetta
@@ -29,33 +29,47 @@ typeAlias FpMLCodingScheme(domain string):
         if item exists
         then ValidateFpMLCodingSchemeDomain(item, domain)
 
-typeAlias BusinessCenter: FpMLCodingScheme(domain: "business-center")
+typeAlias BusinessCenter: <"FpML coding scheme: https://www.fpml.org/spec/coding-scheme/fpml-schemes.html#s5.19">
+    FpMLCodingScheme(domain: "business-center")
 
-typeAlias FloatingRateIndex: FpMLCodingScheme(domain: "floating-rate-index")
+typeAlias FloatingRateIndex: <"FpML coding scheme: https://www.fpml.org/spec/coding-scheme/fpml-schemes.html#s5.114">
+    FpMLCodingScheme(domain: "floating-rate-index")
 
-typeAlias CommodityBusinessCalendar: FpMLCodingScheme(domain: "commodity-business-calendar")
+typeAlias CommodityBusinessCalendar: <"FpML coding scheme: https://www.fpml.org/spec/coding-scheme/fpml-schemes.html#s5.39">
+    FpMLCodingScheme(domain: "commodity-business-calendar")
 
-typeAlias InflationRateIndex: FpMLCodingScheme(domain: "inflation-index-description")
+typeAlias InflationRateIndex: <"FpML coding scheme: https://www.fpml.org/spec/coding-scheme/fpml-schemes.html#s5.123">
+    FpMLCodingScheme(domain: "inflation-index-description")
 
-typeAlias CreditEventType: FpMLCodingScheme(domain: "credit-event-type")
+typeAlias CreditEventType: <"FpML coding scheme: https://www.fpml.org/spec/coding-scheme/fpml-schemes.html#s5.76">
+    FpMLCodingScheme(domain: "credit-event-type")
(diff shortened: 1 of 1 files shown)


chrisisla:
@llynhiavu and @manel-martos - As FpML and codelist experts would you mind reviewing this please? Just need to check the URLs are going to the correct section of the FpML schema descriptions (and they open in Rosetta okay).

manel-martos (commented):
I think it would be useful to include links to the relevant specifications, although I understand there could be concerns around keeping those links up to date.

One option worth exploring would be to reference the actual coding-scheme file definition instead. It is directly accessible and also acts as the URI identifier for the codelist.

For example:
`https://www.fpml.org/spec/coding-scheme/fpml-schemes.html#s5.19`
could be complemented or replaced with:
`https://www.fpml.org/coding-scheme/business-center-9-4.xml`

We think the description added should always match the actual code-list version found under the project's resources directory which are used at validation layer.

chrisisla:
Hi @manel-martos 
Either link would work, as this is really just a workaround to make it so modellers can review the contents of the coding schemes more easily (and I suggested this really in relation to use within Rosetta, which is why I suggested the spec pages as they are more human readable). I'm happy either way though.
I take your point about keeping the links up to date with the scheme versions - this could end up being a modelling overhead that we're actually trying to avoid by using codelists in the first place!
@llynhiavu - do you have any comments? 
Thanks!

PR #5042, opened 2026-08-13
Scope ExchangeListed condition to non-Security assets

### LINKED ISSUE #5022: `AssetBase` conditions cannot be satisfied for exchange-listed securities.
### Issue Summary

The `partyRole exists` guards added in #4794 (merged via #4856) fixed validation for assets that omit `partyRole`. However, for an exchange-listed `Security` that carries one, two of the guarded conditions now contradict each other.

From `base-staticdata-asset-common-type.rosetta` (In 7.0.0 and [currently in main](https://github.com/finos/common-domain-model/blob/master/rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta#L28-L34)):

```
        if isExchangeListed and partyRole exists
        then partyRole -> role = Exchange

    condition IssuerRole: <"If the asset type is a security then the role must be issuer.">
        if assetType = Security and partyRole exists
        then partyRole -> role = Issuer
```

So, for any `Security` with `isExchangeListed = True` and a `partyRole` present, both conditions fire; the first requires the role to be `Exchange`, while the second requires the role to be `Issuer`.

For context, I'm using this to model a simple ETF. Ideally, I would set `partyRole` to be `Issuer` with the issuing fund's legal entity. However, this probably also applies to any other exchange-listed security (e.g. equities, bonds).

---

There's a secondary effect that it causes me to grapple with: Because these contradict, I have to leave `partyRole` unset if I'm to use 7.0.0, but then I can't model a `TransferableProduct` (see [here](https://github.com/finos/common-domain-model/blob/7.0.0/rosetta-source/src/main/rosetta/product-template-type.rosetta#L27-L28)):

```
    condition AssetPartyRoleExists: <"A TransferableProduct must contain an asset and a partyRole.">
        asset ->> partyRole exists
```

And since `Position` requires `product (1..1)`, this leaves no valid way to express a plain holding of my ETF via `Position`.

### Steps to Reproduce

Construct any `Security` with `isExchangeListed = True` and `partyRole` populated.  With either role:                                                                                                                                                             
 
- partyRole -> role = Issuer → violates ExchangeListed                                                                                                                                                  
- partyRole -> role = Exchange → violates IssuerRole

### Expected Behaviour

An exchange-listed security can carry its issuer as an `AssetPartyRole` (or the model documents where a listed security's issuer is meant to be expressed instead).

### Actual Behaviour

Any populated `partyRole` on an exchange-listed `Security` fails one of the two conditions; validation passes only when `partyRole` is omitted, which in turn makes `TransferableProduct.AssetPartyRoleExists` unsatisfiable for listed securities.

### Compatibility

A scoping fix looks backwards compatible. For example, `ExchangeListed` becomes:

```
condition ExchangeListed: <"If isExchangeListed is true then partyRole, where specified, must be Exchange, except for a Security, whose partyRole is governed by IssuerRole.">
      if isExchangeListed and partyRole exists and assetType <> Security
      then partyRole -> role = Exchange
```

which preserves the intended behavior for listed derivatives (where `ExchangeRole` requires `Exchange`) while letting `IssuerRole` govern securities.

### Release

\>7.0.0

### Additional Context

Happy to raise a PR for whichever resolution is preferred:

- the `ExchangeListed` scoping above,
- an `IssuerRole` precedence rule,
- or a documented "listed securities carry no `partyRole`" stance with a corresponding relaxation of `AssetPartyRoleExists`.

mattyoungberg:
Resolves #5022

# Asset Model – Resolve contradictory `AssetBase` conditions for exchange-listed securities

_What is being released_

The `partyRole exists` guards added in #4856 left two `AssetBase` conditions in direct contradiction for an exchange-listed `Security` that carries a `partyRole`: `ExchangeListed` requires the role to be `Exchange`, while `IssuerRole` requires the role of a `Security` to be `Issuer`. No exchange-listed `Security` with a populated `partyRole` can satisfy both, and omitting `partyRole` makes `TransferableProduct.AssetPartyRoleExists` unsatisfiable — so a plain holding of a listed security cannot be expressed via `Position` at all.

This change scopes the `ExchangeListed` condition to `assetType <> Security`, leaving a `Security`'s `partyRole` governed solely by `IssuerRole`. Listed derivatives are unaffected: `ExchangeRole` already requires their `partyRole` to be `Exchange`, and `RoleIsExchange` still forces `isExchangeListed = True` wherever the `Exchange` role appears. The change only relaxes a condition, so any instance that validates today continues to validate; no test pack expectations change.

_Alternatives considered_

- **Relax `IssuerRole` to also permit `Exchange`.** Note that the FpML ingestion mappings currently populate `partyRole` with the exchange (role = `Exchange`) on `Security` instances, so the existing test pack already carries `IssuerRole` violations on 138 exchange-listed securities across ~95 samples. Relaxing `IssuerRole` would clear those failures but requires regenerating `modelValidationFailures` expectations across the test pack, and weakens the issuer semantics of `IssuerRole`.
- **Document that listed securities carry no `partyRole` and relax `TransferableProduct.AssetPartyRoleExists`.** This is a larger behavioural change to `TransferableProduct` and removes the ability to state a listed security's issuer entirely.

The scoping fix here is the minimal, backwards-compatible resolution; the working group may still wish to address the FpML mapping of exchange IDs into `Security.partyRole` separately.

_Review directions_

In `base-staticdata-asset-common-type.rosetta`, inspect the `ExchangeListed` condition on `AssetBase` and observe the added `assetType <> Security` clause and updated description. Confirm that `IssuerRole`, `ExchangeRole` and `RoleIsExchange` are unchanged.

### DIFF
--- a/rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
+++ b/rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
@@ -25,8 +25,8 @@ type AssetBase: <"The base data type to specify common attributes for all Assets
     condition RoleIsExchange: <"If Exchange is specified, it must be an exchange-listed Instrument.">
         if partyRole -> role = Exchange then isExchangeListed = True
 
-    condition ExchangeListed: <"if isExchangeListed is true then partyRole is Exchange.">
-        if isExchangeListed and partyRole exists
+    condition ExchangeListed: <"If isExchangeListed is true then partyRole, where specified, must be Exchange - except for a Security, whose partyRole is governed by IssuerRole.">
+        if isExchangeListed and partyRole exists and assetType <> Security
         then partyRole -> role = Exchange
 
     condition IssuerRole: <"If the asset type is a security then the role must be issuer.">


tomhealey-icma:
I don't agree that this is the right change. The use of Exchange and Issuer conflates partyRole. PartyRole is meant to be the party with the obligation to pay dividends or coupons, which is the Issuer. So ExchangeListed does not automatically mean the Exchange has this obligation. I don't really know of any scenarios where an exchange would have that obligation. I would remove the Exchange condition and if there is a scenario where the Exchange has that obligation then they should be defined as a separate party.

rosetta-source/src/main/rosetta/base-staticdata-asset-common-type.rosetta
tomhealey-icma: I believe the ExchangeListed condition should be removed and Exchange should not be a role.

PR #5059, opened 2026-08-19
[7.x.x] - feat: #4774 Ingestion mapping for FpML executionNotification to CDM ExecutionInstruction

### LINKED ISSUE #4774: Missing ingestion mapping for FpML executionNotification to CDM ExecutionInstruction

BENP-REGNOSYS:
# *CDM Model: FpML executionNotification ingestion mapping to CDM ExecutionInstruction*

_What is being released_

An ingestion mapping has been added for the FpML `executionNotification` message so that its `Trade` element is now mapped onto a CDM `ExecutionInstruction`, in place of the previous mapping which produced a `ContractFormationInstruction`. This aligns the outcome of ingesting an `executionNotification` with the semantics of a newly executed trade, rather than a contract formation event that requires a pre-existing trade state.

To support this, a new enumerated value `Execution` has been added to the `EventIntentEnum` in the `cdm.event.common` namespace, with the description *"The intent is to execute a new trade."* This new intent is used by the ingestion functions to identify that the incoming FpML message represents a new trade execution.

The following ingestion functions have been updated in the `cdm.ingest.fpml` namespace:

- `MapExecutionNotificationToWorkflowStep` (in `ingest-fpml-confirmation-message-func.rosetta`) — now remaps the resolved FpML intent to `Execution` when the raw intent is `ContractFormation` or absent, so that `executionNotification` messages are routed through the new execution path.
- `MapWorkflowStep` (in `ingest-fpml-confirmation-workflowstep-func.rosetta`) — when `intent = Execution`, the workflow step's `Instruction` is now built with an `execution` `PrimitiveInstruction` (and no `before` `TradeState`), instead of the generic primitive instruction with a `before` state.
- A new function `MapTradeToExecutionInstruction` has been added to map an FpML `Trade` (together with the party and quotation lists) directly to a CDM `ExecutionInstruction`, populating `product`, `priceQuantity`, `counterparty`, `ancillaryParty`, `parties`, `partyRoles`, `executionDetails`, `tradeDate` and `tradeIdentifier`.

Expected output JSON files for existing FpML confirmation and record-keeping samples that exercise `executionNotification` messages have been regenerated to reflect the new `ExecutionInstruction`-based output.

_Review directions_

In the CDM Portal open the Textual Browser, search for `EventIntentEnum` in the `cdm.event.common` namespace and observe the new `Execution` value. Then search for `MapExecutionNotificationToWorkflowStep`, `MapWorkflowStep` and `MapTradeToExecutionInstruction` in the `cdm.ingest.fpml` namespace to review the updated ingestion logic. Ingesting any of the updated FpML `executionNotification` sample files (e.g. `pkg-ex01-pkge-execution-notification`, `pkg-ex55-execution-notification`, `msg-ex19-cds-execution-allocations`) will now produce a `WorkflowStep` whose `Instruction` contains an `execution` `PrimitiveInstruction` rather than a `contractFormation` one.

Inspect Pull Request: [#4928](https://github.com/finos/common-domain-model/pull/4928)

Inspect Issue: [#4774](https://github.com/finos/common-domain-model/issues/4774)

### DIFF
--- a/rosetta-source/src/main/resources/functions/business-event/exercise/exercise-cash-settled-func-input.json
+++ b/rosetta-source/src/main/resources/functions/business-event/exercise/exercise-cash-settled-func-input.json
@@ -50,7 +50,7 @@
       }
     },
     "before" : {
-      "@key" : "275f654",
+      "@key" : "3ec1dca4",
       "trade" : {
         "@key" : "700a3655",
         "product" : {
@@ -505,47 +505,9 @@
           }
         } ]
       },
-      "transferHistory" : [ {
-        "@key" : "be8ed309",
-        "transfer" : {
-          "@type" : "cdm.event.common.UnscheduledTransfer",
-          "quantity" : {
-            "value" : 100000,
-            "unit" : {
-              "currency" : {
-                "@data" : "EUR"
-              }
-            }
-          },
-          "asset" : {
-            "@type" : "cdm.base.staticdata.asset.common.Cash",
-            "identifier" : [ {
-              "identifier" : {
-                "@data" : "EUR"
-              },
-              "identifierType" : "CurrencyCode"
-            } ],
-            "assetType" : "Cash"
-          },
-          "settlementDate" : {
-            "unadjustedDate" : "2018-04-01",
-            "dateAdjustments" : {
-              "businessDayConvention" : "FOLLOWING"
-            }
-          },
-          "payerReceiver" : {
-            "payerPartyReference" : {
-              "@ref" : "4f018eab",
-              "@ref:external" : "party1"
-            },
-            "receiverPartyReference" : {
-              "@ref" : "be64334",
-              "@ref:external" : "party2"
-            }
-          },
-          "transferType" : "Premium"
-        }
-      } ]
+      "state" : {
+        "positionState" : "Executed"
+      }
     }
   } ],
   "intent" : "OptionExercise",
--- a/rosetta-source/src/main/resources/functions/business-event/exercise/exercise-partial-exercise-func-input.json
+++ b/rosetta-source/src/main/resources/functions/business-event/exercise/exercise-partial-exercise-func-input.json
@@ -35,7 +35,7 @@
       }
     },
     "before" : {
-      "@key" : "3fec6fa0",
+      "@key" : "a05324da",
       "trade" : {
         "@key" : "2e9c140b",
         "product" : {
@@ -458,47 +458,9 @@
           }
         } ]
       },
-      "transferHistory" : [ {
-        "@key" : "f833714b",
-        "transfer" : {
-          "@type" : "cdm.event.common.UnscheduledTransfer",
-          "quantity" : {
-            "value" : 100000,
-            "unit" : {
-              "currency" : {
-                "@data" : "EUR"
-              }
-            }
-          },
-          "asset" : {
-            "@type" : "cdm.base.staticdata.asset.common.Cash",
-            "identifier" : [ {
-              "identifier" : {
-                "@data" : "EUR"
-              },
-              "identifierType" : "CurrencyCode"
-            } ],
-            "assetType" : "Cash"
-          },
-          "settlementDate" : {
-            "unadjustedDate" : "2018-05-03",
-            "dateAdjustments" : {
-              "businessDayConvention" : "FOLLOWING"
-            }
-          },
-          "payerReceiver" : {
-            "payerPartyReference" : {
-              "@ref" : "4f018eab",
-              "@ref:external" : "party1"
-            },
-            "receiverPartyReference" : {
-              "@ref" : "be64334",
-              "@ref:external" : "party2"
-            }
-          },
-          "transferType" : "Premium"
-        }
-      } ]
+      "state" : {
+        "positionState" : "Executed"
+      }
     }
   } ],
   "intent" : "OptionExercise",
--- a/rosetta-source/src/main/resources/ingest/output/fpml-confirmation-to-workflow-step/fpml-5-10-incomplete-processes/pkg-ex02-swap-spread-single-trade-execution-notification.json
+++ b/rosetta-source/src/main/resources/ingest/output/fpml-confirmation-to-workflow-step/fpml-5-10-incomplete-processes/pkg-ex02-swap-spread-single-trade-execution-notification.json
@@ -2,12 +2,12 @@
   "@model" : "cdm",
   "@type" : "cdm.event.workflow.WorkflowStep",
   "@version" : "0.0.0.master-SNAPSHOT",
-  "@key" : "31e3a616",
+  "@key" : "6a94baf2",
   "proposedEvent" : {
     "intent" : "ContractFormation",
     "instruction" : [ {
-      "before" : {
-        "trade" : {
+      "primitiveInstruction" : {
+        "execution" : {
           "product" : {
             "taxonomy" : [ {
               "source" : "ISDA",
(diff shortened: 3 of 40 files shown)


ReleaseManagement-BP:
Hi @hugohills @manel-martos could either of you please review this PR? May take longer than initially expected, as there are additional file changes compared to https://github.com/finos/common-domain-model/pull/4928

rosetta-source/src/main/resources/ingest/output/fpml-confirmation-to-workflow-step/fpml-5-10-incomplete-processes/pkg-ex02-swap-spread-single-trade-execution-notification.json
```diff
@@ -2,12 +2,12 @@
   "@model" : "cdm",
   "@type" : "cdm.event.workflow.WorkflowStep",
   "@version" : "0.0.0.master-SNAPSHOT",
-  "@key" : "31e3a616",
+  "@key" : "6a94baf2",
   "proposedEvent" : {
     "intent" : "ContractFormation",
```
hugohills-regnosys: Shouldn't the `intent` be unset for executions?

tests/src/test/java/org/finos/cdm/functions/FunctionInputCreator.java
```diff
@@ -824,11 +826,24 @@ private void setDates(InterestRatePayout.InterestRatePayoutBuilder payout, Date
     }
 
     private TradeState getProposedEventInstructionBefore(String resourceName) throws IOException {
-        return ResourcesUtils.getObject(WorkflowStep.class, resourceName).getProposedEvent()
```
manel-martos, line 827: Hi @BENP-REGNOSYS,

This Java update wasn’t included in the original 6.x.x contribution, so it increases the scope of the expected changes. The changes look fine to me, but could you please confirm whether this update is also required for the 6.x contribution?

rosetta-source/src/main/resources/functions/business-event/exercise/exercise-cash-settled-func-input.json
```diff
@@ -505,47 +505,9 @@
           }
         } ]
       },
-      "transferHistory" : [ {
```
manel-martos, line 508: @BENP-REGNOSYS ,

can you verify if the removal of the "transferHistory" is expected in this test scenario?

rosetta-source/src/main/resources/functions/business-event/exercise/exercise-partial-exercise-func-input.json
```diff
@@ -458,47 +458,9 @@
           }
         } ]
       },
-      "transferHistory" : [ {
```
manel-martos, line 461: @BENP-REGNOSYS ,

also in this test case, "transferHistory" is removed from the expectation.

manel-martos (dismissed):
The unexpected changes have prompted a request for clarification, specifically around the `FunctionInputCreator.java` test scenario.

The test-pack updates triggered by this Java change will also need to be validated against downstream projects. In particular, the files under `ingest/output/fpml-confirmation-to-workflow-step/fpml-5-10-native-cdm-events/*` now show all events as executions. If these outputs are consumed by DRR, the enrichment unit tests will likely require corresponding integration updates.

Overall, the contribution looks safe to merge.

hugohills-regnosys:
> The unexpected changes have prompted a request for clarification, specifically around the `FunctionInputCreator.java` test scenario.
> 
> The test-pack updates triggered by this Java change will also need to be validated against downstream projects. In particular, the files under `ingest/output/fpml-confirmation-to-workflow-step/fpml-5-10-native-cdm-events/*` now show all events as executions. If these outputs are consumed by DRR, the enrichment unit tests will likely require corresponding integration updates.
> 
> Overall, the contribution looks safe to merge.

The `FunctionInputCreator` uses the ingested CDM samples to create samples that are used to test functions.

CDM 6.x.x supports both synonym ingest and function ingest.  The `FunctionInputCreator` on 6.x.x uses the ingested CDM samples based on synonym ingest.  As the synonym mappings were not updated inline with the function ingest mappings there were no diffs.

CDM 7.x.x only supports function ingest.  The `FunctionInputCreator` on 7.x.x uses the ingested CDM samples based on function ingest, which are now executions rather than contract formations.  It looks like the `ExecutionInstruction` does not provide a way to set the premium (modelled as a transfer) so it is lost during ingestion.  This should be raised as an CDM issue with `ExecutionInstruction`.

The issue where the intent is set to `ContractFormation` for an execution looks like it needs fixing on both 6.x.x and 7.x.x.

hugohills-regnosys (changes requested):
Need to fix the setting on `intent`, it is incorrectly set to `ContractFormation` for executions.

hugohills-regnosys:
> If these outputs are consumed by DRR, the enrichment unit tests will likely require corresponding integration updates.

Agreed that this will have an impact on DRR test packs, we should ensure that there are tasks in upcoming sprints to adjust them.

ReleaseManagement-BP:
Hi @gpan-isda, please see the comments above. Can you please look at fixing the setting on `intent` for `executions`?

gpan-isda:
> Hi @gpan-isda, please see the comments above. Can you please look at fixing the setting on `intent` for `executions`?

What should be the value to be set for `intent` ? While there is an event type as ExecutionInstruction, there is no corresponding `intent` enum (`EventIntentEnum`). A previous solution to this looked at adding an new enum as `Execution` but since this was change core CDM, the idea was rejected by the group.

hugohills:
I think intent should be blank (it's an optional attribute)

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

hugohills-regnosys (approved):

PR #5069, opened 2026-08-24
Reset Step 5: Determine Unadjusted observation Dates

SuperMonanana:
# _Instruction Composition Reset Step 5: Determine Unadjusted Observation Dates_

_Background_

This release implements Step 5 (Determine Unadjusted Observation Dates & Index Metadata Integration) of the 8-step lifecycle execution engine. A critical requirement for calculating interest rate resets—particularly across compound, average, overnight, and term rates—is the deterministic generation of observation date lists and the integration of Floating Rate Option metadata. This release establishes key date-generation utility logic, incorporates ISDA FRO Matrix lookup capabilities, and introduces observation date schedule compilation. For further information, see issue #4413.

_What is being released?_

This release introduces recursive date generation functions, ISDA FRO Matrix metadata extraction endpoints, and observation date calculation logic catering to Single Observation, Multiple Observation (Overnight/Compounded/Averaged), and Compounded Index methodologies. Additionally, it updates state accumulators and asset floating rate definitions to capture observation schedules and index rounding parameters.

_Functions_

Changes in the cdm.base.datetime namespace:
- `GenerateCalendarDateList`: A recursive utility function that generates a complete list of contiguous calendar dates between a start date and an end date (inclusive).

Changes in the cdm.event.instructioncomposition.reset namespace:
- `ExtractFloatingRateOptionData`: An operational interface function that extracts comprehensive benchmark index metadata (FloatingRateIndexDefinition) from the ISDA FRO Matrix for a specified trade date and floating rate index.
- `Create_DetermineUnadjustedObservationDatesInstruction`: Evaluates calculation period boundaries, index styles, categories, and confirmation methodologies to determine whether single, compounded index, or multiple observation schedules apply, returning the unadjusted observation dates alongside the index metadata.

Changes in the cdm.event.instructioncomposition namespace:
- `UpdateResetCompositionState`: Updated. Modified to implement "Step 5" state overlay logic. Conditionally checks for the presence of the determineUnadjustedObservationDates, incrementally appending unadjustedObservationDates to the cumulative state or preserving existing state values if absent.

_Types and Choices_

Changes in the cdm.event.instructioncomposition namespace:
- `CompositionStepInstructions` (Updated): Expanded to include determineUnadjustedObservationDates for Step 5 processing.

Changes in the cdm.event.instructioncomposition.reset namespace:
- `DetermineUnadjustedObservationDatesInstruction`: Introduced as the official type for Step 5.
[...]

Issue #3287, opened 2024-12-03
Conditions, Functions and Ensuring Consistency Across Different Programming Languages


dschwartznyc:
To ensure the potential for CDM implementations to consistently operate across different programming languages, conditions should either be defined in the CDM Rune model directly or, well documented and clearly defined as such if explicitly implemented in a programming language.   Further, wherever possible existing implementations in a programming language should be migrated to Rune directly.

Absent this consistency, there is clear potential for divergence in the behavior of the library across different languages.

As example, Trade has a condition that explicitly leverages a [Java implementation](https://github.com/finos/common-domain-model/blob/master/rosetta-source/src/main/java/cdm/product/template/functions/FpmlIrd8Impl.java).

```
condition FpML_ird_8: <"FpML validation rule ird-8 - If the same party is specified as the payer and receiver, then different accounts must be specified.">
        if tradableProduct -> product -> contractualProduct -> economicTerms -> payout -> interestRatePayout exists
        then FpmlIrd8(tradableProduct, account) = True
```

This issue should first be reviewed in the TAWG to confirm consensus support for a recommendation and then raised to the SWG for prioritization and on going governance.

Oblongs:
We have discussed this issue before and I feel that it is actually the function that needs to be documented as externally implemented, rather than the condition (i.e. the `func FpmlIrd8` in this case).

I actually raised this as a Rune issue previously:  https://github.com/finos/rune-dsl/issues/773

dschwartznyc:
@Oblongs thanks.

In addition to the concerns cited in the issue above, any externally implemented function needs to have a clear definition of its expected behavior and, ideally, include unit tests that confirm that it behaves as expected.  Otherwise implementation in other languages will be a challenge and CDM overall might face inconsistencies across different implementations.  

It seems that the best route to consistency is to implement functions in Rune directly wherever possible.

chrisisla:
Approved by the TAWG. Resource to now be allocated to do the work.

ReleaseManagement-BP:
@dschwartznyc @chrisisla is there any update on this work? Are resources still to be allocated? Or can this be closed?

chrisisla:
@ReleaseManagement-BP - this issue needs to remain open as it is not resolved as yet.

PR #5073, opened 2026-08-26
Reset Step 6: Adjust Observation Dates

### LINKED ISSUE #4414: CDM Smart Contract Step 6: Adjust the Observation Dates
### Background

The purpose of this issue is to implement **Smart Contract Step 6 - Adjust the Observation Dates**. Please refer to previous issues for an explanation of the Smart Contract process.

This is the sixth Smart Contract Step in the Interest Rate Reset Smart Contract process, the purpose of which is to determine and persist the list of one or more Adjusted Observation Dates that stem from the generated list of unadjusted observation dates of the fifth step.

Note: In this proposal we use the term “adjusted” in conjunction with dates or periods that have been adjusted by the applicable business day convention(s) or calculation methodology(ies).

By automating this step, downstream processes can function smoothly, as the observation of market rates hinges on the correct list of Adjusted Observation Dates obtained in this step.

### Objective

Implement **Smart Contract Step 6 – Adjust the Observation Dates** as an executable Smart Contract Step within the Reset Instruction Smart Contract. The step must:

1.  Use of the placeholder function (For Reset Date)
 *   **Function Name:** _GetNonBusinessDates_
 *   Takes either a list of one or more dates or a period, along with the applicable business centre to provide a list of non-business days to use for adjusting.
 *   This function will be addressed in a separate issue
 *   This is necessary in this step to be able to apply the correct business day Fixing Day Offset for Single Observation FROs.  
    

2.  Use of the placeholder function (For Unadjusted Observation Dates)

*   **Function Name:** _GetNonBusinessDates_
*   Takes either a list of one or more dates or a period, along with the applicable business centre to provide a list of non-business days to use for adjusting.
*   This function will be addressed in a separate issue
*   This is necessary in this step to be able to apply the correct business day Lookback Shift for Multiple Observation FROs.

_Note: Points (1) and (2) above are expected to call the same placeholder function. Depending on whether the Floating Rate Index requires Single or Multiple Observations, (1) or (2) will apply, respectively._  

3.  Process the step:

*   Take the Unadjusted Observation Dates obtained in Step 5.
*   If the Floating Rate Index requires Single Observation:
    *   Using the list of non-business dates related to the Reset Date, apply the Fixing Day Offset, if present, to the **single** Unadjusted Observation Date.
*   If the Floating Rate Index requires Multiple Observations:
    *   First, using the list of non-business dates related to the Observation Dates, apply the Fixing Day Offset, if present, to the start and end dates of the Unadjusted Observation Dates.
    * Second, Using the list of non-business dates related to the Observation Dates, apply the Lookback Shift, if present, to the fixing date adjusted start and end dates of the Unadjusted Observation Dates.
    *   Using the list of non-business dates related to the Observation Dates, adjust all of the dates in the middle following the _Preceding_ Business Day Convention.
    *   **Repeat** any day gaps in the list of dates with the previous date in the list in order to have the same number of dates as the number of calendar days in the period.
*   If the Floating Rate Index requires Index Level Observations:
    *   Return the start and end dates of the Unadjusted Observation Dates

### Scope

To implement an executable Smart Contract Step. To do this, the process is broken down into two parts:

1.  Use of the placeholder function:

*   **Function Name:** _GetNonBusinessDates_
*   **(1)** The Unadjusted Observation Dates from Step 5 and their Applicable Business Centers are used to obtain a list of applicable non business dates.
*   **(2)** The Reset Date and its Applicable Business Centers are used to obtain a list of applicable non business dates.
*   The results from (1) & (2) can differ, as the business centers applicable to the observation dates and the reset date can also be different.
*   This function will be addressed in separate issue. Note that the design and development of these functions are beyond the scope of this work and are included only for indicative purposes.  
    

2.  Creating the Instruction

*   **Step Name:** AdjustObservationDates
*   **Instruction Function**: _Create_AdjustObservationDatesInstruction_
*   **Purpose:**
    *   Receive the following:
        *   The Unadjusted Observation Dates
        *   The Fixing Day Offset, if present
        *   The Lookback Shift, if present
        *   The Category of the Floating Rate Index
        *   The Style of the Floating Rate Index
        *   The Method of the Floating Rate Index, if present
        *   The Non-Business Dates related to the Calculation Period, if required.
        *   The Non-Business Dates related to the Reset Date, if required.
    *   Depending on whether the Floating Rate Index is Single Observation, Multiple Observation, or Compounded Index, three different routes can be chosen:
        *   For **Multiple Observation**, apply first the Fixing Day Offset (if present) and second the Lookback Shift (if present), to the Observation Dates, and then adjust any dates within the period start and end to valid business days, filling any resulting gaps with the immediately preceding date
        *   For **Single Observation**, apply Fixing Day Offset (if applicable) to the Unadjusted Observation Date (the Adjusted Reset date).
        *   For **Index Level Observation**, return the first and last of the Unadjusted Observation Dates. Note that Index Level Observation dates have already been adjusted as part of the Calculation Period adjustments and do not require further adjustment.
    *   Store the resulting list of one or more Adjusted Observation Dates in an _AdjustObservationDatesInstruction_ type, which will be used to generate the Smart Contract Step and persist the output of this step in the Smart Contract History as the reference for subsequent steps

**Inputs** to **_Create_AdjustObservationDatesInstruction_**

*   The **Unadjusted Observation Dates**, extracted from the Smart Contract History in Step 5 (Determine Unadjusted Observation Dates)
*   The **Fixing Day Offset**, extracted from the Workflow Step if present
*   The **Lookback Shift**, extracted from the Workflow Step (if exists)
*   The **Category** of the Floating Rate Index
*   The **Style** of the Floating Rate index
*   The **Index Method** of the Floating Rate Index
*   The **Calculation Period Non Business Dates**, extracted from an external API call using the business centers specific to the Calculation Period
*   The **Reset Date Non Business Dates**, extracted from an external API call using the business centers specific to the Reset Date

### 

### Process

1.  The instruction is created using the _Create_AdjustObservationDatesInstruction_ function with the below input:
    *   unadjustedObservationDates date (0..\*)
    *   fixingDayOffset Offset (0..1)
    *   lookbackShift OffsetCalculation (0..1)
    *   category FloatingRateIndexCategoryEnum (1..1)
    *   style FloatingRateIndexStyleEnum (0..1)
    *   method FloatingRateIndexCalculationMethodEnum (0..1)
    *   nonBusinessDatesCalculationPeriod date (0..\*)
    *   nonBusinessDatesResetDate date (0..\*)
2.  To discern between Single Observation versus Multiple Observation vs Index Level Observation, the Category, Style, and Method are observed:
    *   If the Floating Rate Index is a Category of _Screen Rate_ and does **not** have a Style of _Overnight,_ and the Method is **absent**, then the Floating Rate Index is **Single Observation**. In this case, the Fixing Day Offset (if present) is applied to the single Unadjusted Observation Date, and the business-day shifted date is returned as the output of this instruction. To apply the Fixing Day Offset, the **Reset** non-business dates list is utilised.
    *   If the Floating Rate Index is **(1)** a Category of _Calculated Rate_, it has a Style of _Average FRO_, _Compounded FRO_, and a Method of _OIS Compounding_ or _Overnight Averaging_, or **(2)** a Category of _Screen Rate_ and a Style of _OvernightRate_ then the Floating Rate Index is **Multiple Observation**.. In this case:
        1.  First, the Fixing Day Offset (if present) is applied to the start and end dates of the Unadjusted Observation Dates. To apply the Fixing Day Offset, the **Calculation Period** non-business dates list is utilised.
        2. Second, the Lookback Shift (if present) is applied to the Fixing Day-adjusted start and end dates of the Unadjusted Observation Dates. To apply the Lookback Shift, the **Calculation Period** non-business dates list is utilised.
        3.  The dates contained within the shifted start and end dates of the observation dates are all adjusted as per the non-business days applicable to the calculation period following the **_preceding_** business day convention.
        4.  Any day gaps resulting from adjusting the middle dates are filled with the **immediately preceding date**, such that the total number of resulting Adjusted Observation Dates is equal to the number of calendar days in the period.    *   If the Floating Rate Index is a Category of _Calculated Rate_, has a Style of _Compounded Index_, and a Method of _Compounded Index Method_, then the Floating Rate Index is **Compounded Index**. In this case, the first and last date contained in the Unadjusted Observation Dates are returned as the output of this instruction.

Image
Image
Image
Image
Image

### Compatibility

### Release

TBC

### Additional Context

Smart Contract TaskForce

SuperMonanana:
# _Instruction Composition Step 6: Adjust Observation Dates_

_Background_

This release implements Step 6 (Adjust Observation Dates) of the 8-step lifecycle execution engine. Building upon Step 5's unadjusted observation schedule generation, Step 6 applies fixing day offsets, business day conventions, lookback shifts, and non-business day gap-filling logic to establish the precise observation date lists required for fixing rate retrievals. This release delivers essential business day adjustment utilities, date-shifting procedures, and observation schedule gap-expansion logic. For further information, see issue #4414.

_What is being released?_

This release introduces date manipulation functions for business day conventions, recursive business day shifting, list repeating, and daily gap expansion. It also implements the core Step 6 execution function to compute adjusted observation schedules across Single Observation, Compounded Index, and Multiple Observation methodologies, alongside corresponding type and state accumulator updates.

_Functions_

Changes in the cdm.base.datetime namespace:
- `ExpandMissingObservationDates`: A recursive utility function that inspects a sorted list of business dates and fills calendar day gaps by repeating the prior business date for each missing day, preserving correct day-weightings for compound/average rate calculations.
- `BuildRepeatedDateList`: A recursive utility that builds a list repeating a specified date a set number of times.
- `ShiftBusinessDays`: Recursively shifts a baseline date by a specified integer offset of valid business days while skipping non-business dates.

- `Create_AdjustObservationDatesInstruction`: An operational function that evaluates unadjusted observation dates, fixing day offsets, lookback shifts, FRO categories/styles, and non-business day calendars to calculate and return the final list of adjusted observation dates.

- `UpdateResetCompositionState`: (Updated). Modified to implement "Step 6" state overlay logic. Conditionally checks for the presence of the adjustObservationDates payload, incrementally appending adjustedObservationDates to the cumulative state or preserving existing state values if absent.

_Types and Choices_

- `CompositionStepInstructions` (Updated): Expanded to include adjustObservationDates as an active data payload binding for Step 6 processing.

- `AdjustObservationDatesInstruction`: Introduced as the official type for Step 6.
- `ResetInstructionState` (Updated): Expanded to hold its sixth phase state variable (adjustedObservationDates), allowing the cumulative workflow to track adjusted parameters over its lifecycle.


SuperMonanana:
This PR is technically independent, however it reuses some changes contributed in Step 5 (https://github.com/finos/common-domain-model/pull/5069). To minimise dependencies between the PRs, the Rune code is duplicated here, while the corresponding Java tests are only included in the Step 5 PR.

Duplicate functions are:
`GenerateCalendarDateList` in `cdm.base.datetime` namespace
 `DetermineObservationType` in `cdm.observable.asset.fro` namespace

@ReleaseManagement-BP Ready for review

ReleaseManagement-BP:
Hi @chrisisla @tomhealey-icma @llynhiavu this PR is ready to be reviewed when you have time. Thank you!

dschwartznyc:
This PR seems to introduce new Java native functions that will be used for testing.  Doing so increases the effort of supporting other language implementations and will lead to greater likelihood that there's divergence in behavior between one language and another.

SuperMonanana:
> This PR seems to introduce new Java native functions that will be used for testing.  Doing so increases the effort of supporting other language implementations and will lead to greater likelihood that there's divergence in behavior between one language and another.

Hello @dschwartznyc I'm not sure I understand what you meant about new Java native functions used for testing. There are no Java implementation in this PR, all Java changes are testings of the Rune functions and have no impact on other languages. Could you please elaborate on which functions do you  mean?

dschwartznyc:
@SuperMonanana to ensure that generated code performs to expectations, other languages will need to replicate tests that are implemented in Java.  that expands the support demand for those other languages and creates greater likelihood that behavior in for CDM implemented in one language does not match another.

SuperMonanana:
@dschwartznyc Do you mean that we shouldn’t contribute any Java tests at all? These tests verify that the code developed in Rune behaves as expected, which has also been the approach in the other already merged Reset PRs we’ve raised as there is currently no way to write tests directly in Rune we are aware of.
Not sure if I understand it correctly, but with multi-language support, I would have thought it’s especially important to have tests in each language to ensure CDM behaviour doesn’t diverge between implementations. Yes having tests increases maintenance cost, but wouldn't that actually help prevent differences in the behaviour of generated code from going unnoticed? Or do you mean that tests can be written in a language agnostic way and automatically generated for each supported language?
Could you please point me to an example of the generated code you were referring to?

dschwartznyc:
@SuperMonanana — as discussed in TAWG and elsewhere (see Issue #3287), the goal should be that CDM functionality and its tests are specified in Rune wherever possible, rather than natively in Java or any other target language. The reasoning: once Rune is the source of truth, every generated-language implementation (Java, Python, etc.) inherits the same logic automatically, assuming the generation process is correct. Native tests written in one language don't extend to the others, so they end up reimplemented per language — that's duplicated effort, and it opens the door to behavior drifting apart between implementations without anyone noticing.

To your specific questions:

- **Should we not contribute any Java tests at all?** To the fullest extent possible, yes — new test coverage should move toward being expressed in Rune rather than added natively in Java, so that it's portable to every generated language rather than tied to one.
- **There's currently no way to write tests directly in Rune.** Agreed, and that's a real gap. We've opened #5081 proposing an approach that wouldn't require any grammar changes — using [codeImplementation] functions with file:-prefixed sample JSON as inputs/expected outputs, and condition blocks doing the assertion, with a harness that resolves the file, calls the real function, and checks the result. Since conditions already generate to every target language, the assertion logic itself stays language-agnostic.
- **Tests written language-agnostically and generated per language** — yes, that's the direction as described from the harness side in Issue 5081.

On the generated-code question: the reference to `generators` is the tooling that produces each language's implementation from the Rune definitions — e.g. the [Python generator](https://github.com/finos/rune-python-generator) uses CDM's Rune definitions to create a [Python implementation](https://pypi.org/project/finos-cdm/). Apologies for the ambiguity there.

We're not asking this PR be held up over this — flagging the direction so it's on record. Would welcome input on Issue #5081 in the TAWG.

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
```diff
@@ -460,3 +460,120 @@ func AdjustDateToModfollowingBusinessDay: <"Applies the 'Following' convention,
...
+
+    alias finalRecursiveList: <"Append the current last date to the partially built list, ensuring inclusion of the current endpoint.">
+        if listSize <= 1
+        then []
+        else AppendDateToList(partiallyExpandedDates, lastDate)
+
+    add filledDates: <"Return the original list if only 1 input date, else return the recursively constructed list with gaps filled.">
+        if listSize <= 1 then inputDates else finalRecursiveList
+
+func BuildRepeatedDateList: <"Creates a list containing the same date repeated a specified number of times.">
```
llynhiavu, line 535: can you please provide a simple business example with a handful of dates. unclear why we have the function in the first place.

e.g.
input:
dateToRepeat Fri August 28
numberOfRepeats 3

output:
repeatedDate: {Aug 28, Aug 28, Aug 28)

?

but why?

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
```diff
@@ -460,3 +460,120 @@ func AdjustDateToModfollowingBusinessDay: <"Applies the 'Following' convention,
...
+
+    alias extendedList: <"Append one instance of dateToRepeat to the smaller list.">
+        AppendDateToList(shorterList, dateToRepeat)
+
+    condition PositiveNumberOfRepeats: <"The number of repeats must be greater than zero.">
+        numberOfRepeats > 0
+
+    add repeatedDate: extendedList
+
+func ShiftBusinessDays: <"Recursively shifts a date by a specific number of business days, skipping dates in the provided non business date list.">
```
llynhiavu, line 555: if I understand correctly:

e.g. 
input:
startDate = Friday August 28
offset = +1
nonBusinessDates = Monday August 31 (holiday in UK), Monday Sept 7 (holiday in US)

output:
shiftedDate = Tuesday September 1

right?

and if offset = +2, shiftedDate = Wed September 2
and if offset = -1, shiftedDate = Thu August 27 ?

SuperMonanana:
@dschwartznyc Thanks a lot! Agreed with Rune-native test approach, and thanks for flagging the direction. Issue https://github.com/finos/common-domain-model/issues/5081 makes perfect sense, will keep an eye on it. 
For the current reset work(2 PRs to go), we'll continue contributing the java tests if no strong objections, so we don't introduce untested Rune code.

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
lagarciath: `BuildRepeatedDateList `exists to support interest rate weighting over weekends and holidays during overnight rate calculations. The problem is that financial markets are closed on non-business days, so no new rates are published. However, interest still accrues on these calendar days. 

So under standard financial conventions, weekend days reuse the rate published on the preceding business day. `BuildRepeatedDateList `is needed when filling calendar gaps (`ExpandMissingDates`). 

The system creates a repeated array like [Aug 28, Aug 28, Aug 28] for a Friday-through-Sunday period. This allows array-based calculation engines to map Friday’s published rate three times without breaking vector operations downstream.

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
lagarciath: Yes, your understanding is correct.

Here is how the `ShiftBusinessDays` function processes each scenario:
- offset = +1 --> Tue Sep 1: the code steps through Saturday (Aug 29), Sunday (Aug 30), and Monday (Aug 31, holiday). Because all three are non-business days, the offset counter stays at 1 until it hits Tuesday, Sep 1 (the first valid business day)
- offset = +2 --> Wed Sep 2: building on the first shift, the next available business day after Tuesday is Wednesday, Sep 2, where the offset counter reaches 0
- offset = -1 --> Thu Aug 27: Stepping backward 1 day lands on Thursday, Aug 27. Since Thursday is a valid business day, the offset counter returns Aug 27

llynhiavu: I see...

was my nonBusinessDates list inaccurate then?
- nonBusinessDates = Monday August 31 (holiday in UK), Monday Sept 7 (holiday in US)

it should have included all weekends as well (?)
- nonBusinessDates = Sat Aug 29, Sun Aug 30, Mon Aug 31, Sat Sept 5, Sun Sept 6, Mon Sept 7

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
llynhiavu: Thanks, now I understand better. I guess I didn't get the intent from the _**short**_ version func descriptions (at the top of this page.) 

Should the func descriptions in the code itself need a business rationale? (vs being explained here)

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
lagarciath: Yes, so the input of this function is a list of "nonBusinessDates", that contains all the dates that are different than a business day, including then weekends and holidays.

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
lagarciath: This particular function is described in the description next to the name. I know this is a generic description (really describing what the function does but without any context), but I think it makes sense to be like that so that it can be reused elsewhere. You can get the context of the function for our particular use case by going up in the chain and looking at how and where it is being used.

Does it make sense? Do you think we need more descriptions?

llynhiavu: I understand. I won't push for it but deferring to others for opinions on this.

chrisisla: In Lyteck's example, is there any reason why it needs to be the same date repeated i.e. [Aug 28, Aug 28, Aug 28]? I know we're taking the rate from Aug 28 and using it for each day over the weekend (in this example) but would it not make more sense to have the array be chronological i.e. [Aug 28, Aug 29, Aug 30] with the same rate against each? 
This function just seems to be a more complex way of doing what `GenerateDateList` can already do?
I'm querying this because if I saw a list of dates which had the same date multiple times in a row I would most likely think that something had gone wrong in my function.

chrisisla (commented):
I'm struggling with the date processing on this one slightly I'm afraid. Could you provide a quick example of the sort of ranges and dates you are expecting please i.e. are we expecting a sequential range of dates like 23rd, 24th, 25th, 26th, or ranges with missing dates like 2nd, 14th, 26th?

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
lagarciath: Yes, that is a reasonable way to frame it... The design choice comes down to capturing _observation date_ versus _application date_, especially when pairing dates with values in data types like DatedValue (it is part of the next steps):
- **DatedValue.date represents the Observation Date**: Since DatedValue defines the date on which a rate becomes effective or was observed, setting the date to Aug 29 or Aug 30 would be inaccurate, no rate was actually published or observed on those days 
- **Auditability**: Repeating Aug 28 explicitly signals that the August 28th rate is being weighted three times, rather than implying a new observation occurred over the weekend

So while seeing repeated dates in a generic date utility might feel like a bug, in this context it’s intentionally preserving the "source-of-truth" observation date

chrisisla: The end result of this step is the `AdjustObservationDatesInstruction` which has a single attribute of `adjustedObservationDates` which is just a list of dates. I'm not sure where `datedValues` come into it?

Although the `DatedValue` type does not specify that the dates should be unique, the way it is used I think implies this, especially since it states that the list of dates should be in ascending order.

Is `DatedValue` actually used here though - or is this date list to create `datedValues` in a following step? Putting the same date 3 times into a sequential list of dates does not seem correct, even after taking your explanation into account (as why would you observe the same rate 3 times for the same day?).

If this is a transient step, whereby the data is eventually discarded, then it is probably ok. However, I don't think this should be a generic function as I can't think of any instance where you would create multiple `datedValue` with the same date (not until the time is added to `datedValue` anyway!).

Sorry if I'm misinterpreting what you are doing but this still seems a bit  obscure to me.

llynhiavu: Would renaming the function name make it easier to digest? e.g. BuildRepeatedDateList  > BuildRepeatedObservationDateList

lagarciath: Sorry if I wasn't clear enough, but the `DatedValue` is coming in step 7. Its sole purpose is to map rates used during the period to their respective publication dates (filling in non-business days using the last available published rate). Because it’s tailored to this specific process, maybe Lyteck’s renaming proposal to make its scope explicit makes sense.

llynhiavu: I would support that function name change as a compromise (i.e. explains the repeated date, while at the same time having a more explicit name would deter someone not interested in observables.  if not this proposed name perhaps another one that you would prefer @chrisisla ?

llynhiavu (changes requested):

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
chrisisla: Let's go with Lyteck's suggestion of changing the function name to `BuildRepeatedObservationDateList`. This does make the use case of the function explicit 👍

chrisisla: Actually, I think this function name is probably okay, as it is clear we're building a list of repeated dates - there is no real need to put the use case into the function name here, the purpose of the function is pretty clear. I think the `ExpandMissingDates` is where we may need to highlight exactly why it is putting the same date in there?

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
```diff
@@ -488,3 +488,98 @@ func GenerateCalendarDateList: <"Recursively generates a list of all calendar da
 
     add dateList: <"Return the new list if still within range, stops recursion when startDate passes endDate.">
         if active then newList
+
+func ExpandMissingDates: <"Recursively fills gaps between consecutive dates by repeating the previous date for each missing day. Ensures every day gap is represented by repeated instances of the previous business date.">
```
chrisisla: So this is the function name that I think is a bit confusing. `ExpandMissingDates` to me would be filling in a date range e.g. passing in  [28, 31] I would expect to receive back [28, 29, 30, 31]. 
So I think this function should be renamed to `ExpandMissingObservationDates`, and then in the comment we can explain why we repeat previous dates and don't create a sequential date list. 
How does that sound?

chrisisla (changes requested):

rosetta-source/src/main/rosetta/base-datetime-func.rosetta
lagarciath: Sounds good!

SuperMonanana: Renamed!

SuperMonanana:
PR for 7.x.x: https://github.com/finos/common-domain-model/pull/5176

llynhiavu (approved):

chrisisla (approved):

ReleaseManagement-BP:
@SuperMonanana could you please update the branch? Once thats done we can merge this in!

SuperMonanana:
@ReleaseManagement-BP Sorry I'm on holiday from today until 30th September...would you mind press that "Update" button? I believe there are no conflicts (you could also merge directly) Otherwise I'll take care of it when I'm back

ReleaseManagement-BP:
Sorry @SuperMonanana I don't have the option because it's a tokenovate branch, so I don't have the ability to. @lukebanas @iansloyan  is anyone else from Tokenovate able to update this branch? Or @lagarciath would you be able to since you've worked on these steps?
image

PR #5107, opened 2026-09-01
[6.x.x] Update bundle dependency to version 11.132.4

### LINKED ISSUE #5080: Synonym Ingestion: intermittent ingestion test failures / build flake
### Issue Summary

The CDM's FpML synonym ingestion tests check the ingestion engine correctly transforms test pack samples from FpML to CDM.  The synonym ingestion engine is considered legacy (it has been replaced by function ingest), however it is still present on CDM version `6.x.x`, and DRR versions `6.x.x` and `7.x.x`.

CDM specifies its FpML synonym mappings in:

- `rosetta-source/src/main/rosetta/mapping-fpml-confirmation-tradestate-synonym.rosetta`
- `rosetta-source/src/main/rosetta/mapping-fpml-confirmation-workflowstep-synonym.rosetta`

and the Synonym Ingestion Engine is code generated from them. 

Several unit tests exist in CDM and DRR that ingest sample FpML documents through that generated engine and compare the result against a checked-in expectation. In the CDM these are the classes in `tests/src/test/java/com/regnosys/ingest/fpml/` — for example
`Fpml510IncompleteProductIngestionServiceTest`, `Fpml513ProductIngestionServiceTest` and
`Fpml510ProcessesIngestionServiceTest`.

One of those tests, `Fpml510IncompleteProductIngestionServiceTest`, **intermittently fails on mismatched expectations** for equity basket samples. The failure is not reproducible from source: the same commit passes or fails between builds, with no change to the model, the synonym mappings, the sample files or the expectations.

**Important:** This is build time code generation flake.  Once generated, the ingestion engine behaviour is deterministic.

**Frequency.** Intermittent. Based on CDM and DRR releases, ~6% failed and had to be restarted.

**Workaround.** Re-run the build.

**Status.** Cause identified; a fix to the code generator is in progress upstream.

### Steps to Reproduce

1. Check out the branch and build it clean.
2. Run the ingestion tests, for example:
   ```
   mvn -pl tests test
   ```
3. Repeat from clean. The FpML parse handlers are code generated during the test run into `tests/target/DEFAULT/`, so each run generates them afresh.

The failure occasionally appears on some runs and not others. Nothing about the checkout changes between a passing run and a failing one.

### Expected Behaviour

The same commit produces the same ingestion result every time.

### Actual Behaviour

Nine tests fail, all equity basket samples under
`cdm-sample-files/fpml-5-10/incomplete-products/`:

- `equity-swaps/eqs-ex03-index-quanto-long-form.xml`
- `equity-swaps/eqs-ex07-long-form-with-stub.xml`
- `equity-swaps/eqs-ex08-composite-basket-long-form-separate-spreads.xml`
- `equity-swaps/trs-ex01-equity-basket.xml`
- `equity-options/eqd-ex08-basket-long-form.xml`
- `equity-options/eqd-ex15-basket-passthrough-long-form.xml`
- `equity-options/eqd-ex20-nested-basket.xml`
- `equity-options/eqd-ex21-flat-weight-basket.xml`
- `equity-options/eqd-ex26-mixed-asset-basket.xml`

In each one the constituent quantity is absent from the ingested result — the expectation carries it, the actual output does not. From `eqd-ex21-flat-weight-basket.xml`, one constituent of three:

```diff
                       "assetClass" : "Equity"
                     }
-                  },
-                  "quantity" : [ {
-                    "value" : {
-                      "value" : 0.50,
-                      "unit" : {
-                        "financialUnit" : "Weight"
-                      }
-                    }
-                  } ]
+                  }
                 },
```

Only these basket samples are affected; no other test class has been observed failing on a build where these tests go red. 

### Compatibility

None. This is a defect in how the ingestion engine is code generated, not a change to the model, the synonym mappings or the expectations. No CDM artefact changes, so there is no backward compatibility impact for consumers.

### Release

The defect is in the code generation of the Synonym Ingestion Engine rather than in any released CDM artefact, so it is not specific to a CDM release.  However, this issue caused build flake on CDM version `6.x.x`, and DRR versions `6.x.x` and `7.x.x`.

### Additional Context

Historically we found that raising the Codefresh build runtime memory appeared to fix or reduce non-deterministic ingestion test failures, so that was the previous response. Now that we've traced the root cause, here's the more accurate picture.

Raising the memory did fix real failures. An under-sized build runtime gets JVMs killed by the kernel rather than throwing a clean out-of-memory error, and until recently all the test JVMs generated their handlers into one shared directory — so a test JVM killed part-way through generation left a half-written tree that the next one built on top of. More memory genuinely closed that off.

What it couldn't do is change the randomness behind the mapping failures. That comes from Java identity hash codes, which the synonym map was ordered by; they're randomised per run and are the same whatever heap you give the JVM. We've measured that. So the memory increases were treating a real but different problem, and the flake kept happening underneath — it still reproduced about once in twelve full builds on the current branch, after the last increase.

Both causes are now fixed properly (in upstream project): the synonym map no longer depends on hash order, and each JVM generates into a directory of its own and rebuilds it from scratch. Worth keeping the Codefresh runtime memory where it is, though — the build declares 8g of JVM heap inside an 8000Mi runtime, so there's no headroom to give back.

hugohills-regnosys:


hugohills-regnosys:
# _Infrastructure - Dependency Update_

_What is being released?_

This release updates the `Bundle` dependencies:

Version updates include:

- `Bundle` `11.132.4` - CVE fixes.
- `Bundle` `11.132.3` - Test logging fix, synonym fix for clashing attributes.
- `Bundle` `11.132.2` - Synonym build flake fix - see issue [#5080](https://github.com/finos/common-domain-model/issues/5080)
- `Bundle` `11.132.1` - Ingestion test logging improvements.

_Review Directions_

Changes can be reviewed in PR: #5107

JayasriR (approved):

PR #5122, opened 2026-09-03
Ben Page CDM-Ingest-fpml-confirmation-party - Mapping identifierType to fpml schemes

regnosys-prod-user:

PR #5124, opened 2026-09-03
[master] - Ben Page CDM-Ingest-fpml-confirmation-party - Mapping identifierType to fpml schemes (Updated Expectations)

### LINKED ISSUE #5074: Mapping`IdentifierType` to `partyId` schemes in FpML
### Background

In DRR for NoA regimes, `Trade Party 2 - ID Type` can be one of three values:

LEID
NPID
PLID

Image

The logic in DRR to have these types populated is such:

Image

However, in CDM, there is currently no identifierType for a Person ID, so the ingest message will never populate with NPID or PLID

### Proposal

To add `identifierType` to `MapPersonIdentifier`

Image

And to create `MapPersonIdentifierTypeEnum`

Image

Using the suggested FpML schemes on the FpML coding-scheme's list

Image

### Business Justification

To give thorough testing coverage for NoA regimes in DRR

### Compatibility

backward-compatible

### Release

CDM 6.x.x, 7.x.x, 8Dev

regnosys-ops:
Update Expectation for PR #5122


rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
```diff
@@ -539,12 +537,16 @@ func MapNaturalPerson:
...
-                    extract
-                        MapPersonIdentifier(
-                                item,
-                                fpmlPerson -> contactInfo -> address -> country
-                            ),
+            personId: fpmlPerson
+                    extract person [
+                        person -> personId
+                            extract fpmlPersonId [
+                                MapPersonIdentifier(
```
manel-martos, line 544: @BENP-REGNOSYS - can you clarify the rationale for sourcing this mapping from `person -> contactInfo -> address -> country` instead of `person -> country`?

Also, just noting that mappings aren't considering data points within `person -> businessUnitReference` in `MapPersonIdentifier`. Is that intentional?

There's a question currently raised to the FpML working group about country representation for MiFIR too, so we may need to revisit this depending on their conclusion.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-other-func.rosetta
```diff
@@ -2127,6 +2126,19 @@ func MapPartyIdentifierTypeEnum:
             "http://www.fpml.org/coding-scheme/external/mifir/extension-iso10383" then MIC,
             default empty
 
+func MapPersonIdentifierTypeEnum:
```
manel-martos, line 2129: @BENP-REGNOSYS - should we raise a separate DRR review task regarding the current usage of the person identifier enums, or is there an existing follow-up action already tracking this? A quick DRR review shows we are currently only considering NPID and PLID for enum-based filtering.

manel-martos (commented):
This is safe for contribution since it has no broken expectations and the new mapping data points in the test-pack samples look good. Before I proceed with reviewing the other branches, please clarify the pending comments.

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-other-func.rosetta
```diff
@@ -2127,6 +2126,19 @@ func MapPartyIdentifierTypeEnum:
...
+    inputs:
+        value string (0..1)
+    output:
+        result PersonIdentifierTypeEnum (0..1)
+    set result:
+        value switch
+            "http://www.fpml.org/coding-scheme/external/person-id" then NPID,
+            "http://www.fpml.org/coding-scheme/external/person-id/privacy-law-id" then PLID,
```
manel-martos, line 2137: @BENP-REGNOSYS - Can you verify the source of the `privacy-law-id` scheme URI? We aren't seeing it in any Fpm¡ML coding scheme releases. 
Also, I noticed `person-id/mifir/concatenation` is in the issue description but wasn't included in the switch logic, was it intentional?

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-other-func.rosetta
ReleaseManagement-BP: We could do, I'm not sure where the use case for the other enum values or what regimes those other enum values are applicable to. Certainly not applicable for the NoA regimes which this change was made for

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-other-func.rosetta
ReleaseManagement-BP: I created the `privacy-law-id` URI. I spoke with @llynhiavu and he said the URI's are merely suggestions and not anything concrete. When I presented this issue I highlighted this and asked for suggestions, but did not get anything from anyone, so I used one that made sense with alongside the other URI's

If you have found a PLID URI in the fpml coding scheme releases, I am happy to use that.

`person-id/mifir/concatenation`, I did not see any corresponding enum, so did not map anything to it

rosetta-source/src/main/rosetta/ingest-fpml-confirmation-party-func.rosetta
ReleaseManagement-BP: I couldn't find any recorded rationale. It was initially introduced in
[#3996](https://github.com/rosetta-models/common-domain-model/pull/3996)
 (@hugohills-regnosys 2025-09-01) with no comment, and the synonym-based ingest that preceded it had no mapping for PersonIdentifier.country

My guess would be because the identifier types in PersonIdentifierTypeEnum are national-authority documents (NIDN, SOSE, CCPT, NPID), and for those, where someone is domiciled predicts the issuing authority better than where they work. Someone resident in Ireland working in London has an Irish national ID.

I'm guessing based on the FpML WG review, if country representation needed to change, that would be done in a separate PR changing country across the board?

manel-martos (approved):
- Future work is expected on ingestion logic to support MiFIR requirements and businessUnitReference extensions.
- Accepted interpretation for the current `person -> country` mapping logic. It may require upgrades as per FpML or DRR uses.
Approved for merge.

PR #5128, opened 2026-09-03
Reset Step 7: Calculate Reset Value

### LINKED ISSUE #4482: CDM Smart Contract Step 9: Calculate the Reset Value
### Background

The purpose of this issue is to implement **Smart Contract Step 9 - Calculate the Reset Value**. Please refer to previous issues for an explanation of the Smart Contract process.

This is the ninth Smart Contract Step in the Interest Rate Reset Smart Contract process, the purpose of which is to calculate and persist the final Reset Value given the market rates recorded in the eighth step.

By automating this step, downstream processes can function smoothly, as the creation of the definitive Reset Instruction hinges on the correct calculation of the Reset Rate obtained in this step.

### **Objective**

Implement **Smart Contract Step 9 - Calculate the Final Reset Value** as an executable Smart Contract Step within the Reset Instruction Smart Contract. The step must:

*   Take the Market Rate Observation(s) obtained in Step 8
*   If the Floating Rate Index requires **Single Observation**:
    *   There should be **one** Market Rate. This is passed straight through to the output.
*   If the Floating Rate Index requires **Multiple Observations**:
    *   If the Calculation Method is _Compounding_, the Compounding Formula is applied to the list of Market Rate Observations. If rounding is specified, the resulting value is rounded accordingly and returned as the output. If no rounding is specified, the default rounding is to the nearest one hundred-thousandth of a percentage point.
    *   _Note: The _Averaging_ Calculation Method is out of scope and is not currently supported.
*   If the Floating Rate Index requires **Index Level Observation**:
    *   There should be **two** Market Rate Observations corresponding to the start and end dates of the Adjusted Observation Period. The Compounded Index Method Formula is applied to these two Market Rate Observations and the resulting value is rounded accordingly and returned as the output. If no rounding is specified, the default rounding is to the nearest one hundred-thousandth of a percentage point.
*   The calculated Reset Value and Calculation Method are returned as the final output to build the Instruction of this step.
    

### **Scope**
To implement an executable Smart Contract Step.
*   **Step Name:** CalculateResetValue
*   **Instruction Function**: _Create\_CalculateResetValueInstruction_
*   **Purpose:** 
    *   Receive the following:
        *   The list of Market Observations
        *   The list of Adjusted Observation Dates
        *   The Rounding convention applicable to the reset value
        *   The Currency of the Floating Rate Index
        *   The Day Count Fraction applicable
        *   The Category of the Floating Rate Index
        *   The Style of the Floating Rate Index
        *   The Method of the Floating Rate Index from the FRO Matrix, if present
        *   The Method of the Floating Rate Index from the Confirmation, if present
    *   Depending on whether the Floating Rate Index requires Single Observation, Multiple Observation, or Index Level Observation, the calculation applied to the market observations will differ:

For **Single Observation**, there is only **one** market observation. This observation is passed straight through to the output.

For **Multiple Observation**, the Compounding Formula is applied to the list of the Market Observations. Once the final value is calculated, it is rounded appropriately if a Rounding convention is specified. If no rounding convention is specified, the default rounding is to the nearest one hundred-thousandth of a percentage point. The Compounding Formula is displayed below:
       
Image
        

For **Index Level Observation**, there are **two** market observations, corresponding to the rates at the beginning and the end of the Adjusted Observation Period. Given that only the Compounded Index Method is in-scope, the Compounded Index Method Formula is applied to obtain a single value. This value is then rounded appropriately if Rounding is specified. If no rounding convention is specified, the default rounding is to the nearest one hundred-thousandth of a percentage point. The Compounded Index Method Formula is displayed below:

Image

* *   _Note_: the Averaging calculation methodology is_ _**not**_ _in scope._

**Inputs** to _**Create_CalculateResetValueInstruction**_
*   The **Market Observations**, extracted from the Smart Contract History in Step 8 (Record market Rates)
*   The **Adjusted Observation Dates**, extracted from the Smart Contract History in Step 7 (Adjust Observation Dates)
*   The **Rounding** Convention applicable to the final reset value, extracted from the Smart Contract History in Step 2
*   The **Currency** applicable to the final reset value, extracted from the Smart Contract History in Step 2
*   The **Day Count Fraction** used to calculate the Compounding and Compounded Index Method formulas, extracted from the Workflow Step if present, else from the Smart Contract History in Step 2
*   The **Category** of the Floating Rate Index, extracted from the Smart Contract History in Step 2
*   The **Style** of the Floating Rate Index, extracted from the Smart Contract History in Step 2
*   The **Method from the FRO** of the Floating Rate Index, extracted if present from the Smart Contract History in Step 2 (Record FRO Data)
*   The **Method from the Confirmation** of the Floating Rate Index, extracted if present from the Workflow Step
    *   _Note 1_: Should both calculation methods be present, the methodology within the Confirmation takes precedence over the one extracted from the FRO Matrix. This aligns with the ISDA documentation architecture, where terms specified in the Confirmation take precedence over terms set out in a Definitions book (e.g. the ISDA 2021 Interest Rate Definitions)_. 
    *   _Note 2_: If the Style is Overnight Rate, the calculation method from the confirmation should be present._

### **Process**
1.  The instruction is created using the _Create\_CalculateResetValueInstruction_ function with the below input:
    *   marketRates number (1..\*)
    *   adjustedObservationDates date (1..\*)
    *   rounding Rounding (0..1)
    *   currency string (1..1)
    *   dayCountFraction DayCountFractionEnum (0..1)
    *   category FloatingRateIndexCategoryEnum (1..1)
    *   style FloatingRateIndexStyleEnum (1..1)
    *   calculationMethodologyFromFRO FloatingRateIndexCalculationMethodEnum (0..1)
    *   calculationMethodologyFromConfirmation CalculationMethodEnum (0..1)
2.  To discern between Single Observation, Multiple Observation, and Index Level Observation approaches, the Category, Style, and Method are observed:
    *   If the Floating Rate Index is a Category of _Screen Rate_ and does **not** have a Style of _Overnight_, and the Method (both from FRO and Confirmation) is **absent,** then the Floating Rate Index requires a Single Observation. In this case, the **Market Rate** **is returned as output** of this instruction.
    *   If the Floating Rate Index is a Category of _Calculated Rate_, has a Style of _Compounded Index_, and a Method of _Compounded Index Method_, then the Floating Rate Index requires Index Level Observation. In this case, the Compounded Index Method formula is applied to the **two market rates** (corresponding to the start and end rates of the Adjusted Observation Dates) and the **resulting (rounded) value is** **returned as the output** of this instruction.
    *   If the Floating Rate Index is **(1)** a Category of _Calculated Rate_, it has a Style of _Average FRO_, _Compounded FRO_, and a Method of _OIS Compounding_ or _Overnight Averaging_, or **(2)** a Category of _Screen Rate_ and a Style of _OvernightRate_, then the Floating Rate Index requires **Multiple Observation**. 
        1.  If the Calculation Method (the Confirmation method takes precedence if present, else the FRO method is considered) is _Compounding_ then the Compounding Formula is applied to the list of Market Observations and the **resulting (rounded) value is returned as the output** of this instruction.
        2.  The Calculation Method of _Averaging_ is currently out-of-scope and not supported.
3.  Once the Reset Value has been calculated, the CalculateResetValueInstruction is built, consisting of the _resetValue_ and the _calculationMethod_. A couple of considerations:
    *   The _resetValue_ in CDM is of type _Price_, which requires the currency to be specified. This explains the reason for _Currency_ being an input to the creation of the instruction in this step
    *   The _calculationMethod_ returned as part of the output of this Step 9 is of type _CalculationMethodEnum_, the same type as the _methodFromConfirmation_ input to this step. Given this, the output _calculationMethod_ represents the “merging” of the _methodFromFRO_ and _methodFromConfirmation_ enumerations into one:
        1.  If the _methodFromConfirmation_ is present, its value is simply passed to the output. It can take one of these values: {_Averaging, Compounding, CompoundedIndex_}
        2.  If the _methodFromFRO_ is present **and** the _methodFromConfirmation_ is absent, it can take one of these values: {_OISCompound, Average, Compounded, AllInCompounded_}, which are translated as follows:
            1.  _OISCompound → Compounding_
            2.  _Average_ → Not translated, out of scope
            3.  _Compounded → CompoundedIndex_
            4.  _AllInCompounded_ → Not translated, out of scope
4.  The generated instruction is then used to create the Smart Contract Step, which records the final Reset Value and the Calculation Method in the Smart Contract History. The Calculation Method is a necessary data point to generate the definitive Reset Instruction in the subsequent step.
    

**Outputs**
*   Smart Contract containing:
    *   A New Smart Contract Step
        *   Smart Contract Step Identifier
        *   The CalculateResetValue Instruction
    *   Trade Identifier
    *   Smart Contract Type
    *   Smart Contract Identifier
    *   Smart Contract History
        *   Reset Value
        *   Calculation Method

### Compatibility

### Release

TBD

### Additional Context

CDM Smart Contract Task Force

SuperMonanana:
# _Instruction Composition Framework Functions_

_Background_

This release implements Step 7 (Calculate Reset Value & Market Rate Integration) of the 8-step lifecycle execution engine. Following the determination and adjustment of observation schedules in preceding steps, Step 7 fetches observation market rates and computes the final reset rate or fixing value using applicable calculation methodologies, such as compounded index formulas and rate transformation mappings. This release delivers essential mathematical repetition utilities, rate formula functions, market rate retrievals, and calculation state updates. For further information, see issue #4482.

_What is being released?_

This release introduces mathematical rate helper utilities, compounding formula functions, and enum mapping logic. It implements the primary Step 7 execution function to retrieve market rates and compute final reset values, updates the global instruction step containers, and completes the state overlay logic for the reset accumulator.

_Functions_

Changes in the cdm.base.math namespace:
- `RepeatNumber`: A utility function designed to repeat a numerical value across a sequence, supporting array construction for compound rate weighting calculations.

Changes in the cdm.observable.asset.calculatedrate namespace:
- `ApplyCompoundedIndexFormula`: Implements the mathematical formula to compute a compounded rate fixing from start and end compounded index observations over a calculation period.

Changes in the cdm.product.asset.floatingrate namespace:
- `MapFloatingRateIndexCalculationMethodEnumToCalculationMethodEnum`: A mapping function that aligns Floating Rate Option calculation methods from FRO Matrix standards to the general calculation method enumerations used across product specifications.

- `GetMarketRates`: An operational interface function that retrieves published market benchmark rates or index values for the list of adjusted observation dates.
- `Create_CalculateResetValueInstruction`: An operational function that takes observation dates, fetched market rates, and index parameters to compute the final reset rate value, returning the result in a structured instruction.

- `UpdateResetCompositionState`: (Updated). Modified to implement "Step 7" state overlay logic. 

_Types and Choices_

- `CompositionStepInstructions` (Updated): Expanded to include calculateResetValue as an active data payload binding for Step 7 processing.

- `CalculateResetValueInstruction`: Introduced as the official type for Step 7.
- `ResetInstructionState` (Updated): Expanded to hold its seventh phase state variable.


ReleaseManagement-BP:
@chrisisla @llynhiavu @tomhealey-icma This step is now ready to be reviewed, please have a look when you get time, dependent if you want steps 5 and 6 merged before reviewing step 7.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
```diff
@@ -117,6 +123,130 @@ func Create_AdjustDateInstruction: <"This function generates the Instruction Com
             adjustedDate: adjustedResetDate
         }
 
+// Function to fetch the market rates and observation dates key-value pairs
+func GetMarketRates: <"This function, given a floating rate index and a list of observation dates, fetches the corresponding market rate observations and returns the market rates and related observation dates in a key-value pair format.">
+    [codeImplementation]
+    inputs:
+        floatingRateIndex cdm.observable.asset.FloatingRateIndex (1..1) <"The floating rate index.">
```
chrisisla, line 130: Any reason why this type needs the namespace prefix and you have not imported it instead?

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
```diff
@@ -117,6 +123,130 @@ func Create_AdjustDateInstruction: <"This function generates the Instruction Com
...
+        observationDates date (1..*) <"The list of observation dates.">
+    output:
+        observedRateDates DatedValue (1..*) <"The list of market rate and observation dates key-value pairs.">
```
chrisisla, line 133: Are we expecting this to be a list of unique dates (related to my questions on Step 6).
This output is also set as mandatory. If there is no code implementation of this function what will happen if the function is called?

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
```diff
@@ -117,6 +123,130 @@ func Create_AdjustDateInstruction: <"This function generates the Instruction Com
...
+                    dayCountFraction,
+                    DateDifference(
+                            adjustedCalculationPeriod -> adjustedStartDate,
+                            adjustedCalculationPeriod -> adjustedEndDate
+                        ) + 1
+                )
+    // (4) --- MULTIPLE OBSERVATION ---
+    alias calendarDays: <"The number of calendar days contained in the observation period for a multiple observation FRO.">
+        marketRates count
+    alias weights: <"A list of ones repeated by the number of calendar days in the period.">
```
chrisisla, line 179: Why?

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
```diff
@@ -117,6 +123,130 @@ func Create_AdjustDateInstruction: <"This function generates the Instruction Com
...
+func GetMarketRates: <"This function, given a floating rate index and a list of observation dates, fetches the corresponding market rate observations and returns the market rates and related observation dates in a key-value pair format.">
+    [codeImplementation]
+    inputs:
+        floatingRateIndex cdm.observable.asset.FloatingRateIndex (1..1) <"The floating rate index.">
+        observationDates date (1..*) <"The list of observation dates.">
+    output:
+        observedRateDates DatedValue (1..*) <"The list of market rate and observation dates key-value pairs.">
+
+// Instruction Composition STEP 7 - Calculate Reset Value
+func Create_CalculateResetValueInstruction: <"This function generates the Instruction Composition step instruction necessary to calculate the final reset value and applicable calculation methodology, adding the resulting reset rate to the CalculateResetValueInstruction.">
```
chrisisla, line 136: General note on this function... it is very long! Can we put some blank lines between things like aliases and conditions to make it more readable please? It will appear even longer but will be more user friendly. 
A lot of the computation seems to differ based upon the observation type. Wondering if splitting the function out based upon the observation type might make it more efficient and also easier to navigate?

rosetta-source/src/test/java/cdm/event/instructioncomposition/reset/functions/UpdateResetCompositionStateTest.java
```diff
@@ -236,4 +243,78 @@ void shouldProduceNullAdjustedResetDateWhenAdjustDateAndCurrentStateAreBothAbsen
...
+    @DisplayName("Step 7 - Final Calculation Results")
+    class CalculateResetValueTest {
+
+        @Test
+        @DisplayName("Sets resetValue, calculationMethod and observedRateDates from the step instruction when calculateResetValue is present")
+        void shouldSetResetValueCalculationMethodAndObservedRateDatesFromStepWhenCalculateResetValueIsPresent() {
+            Price resetValue = Price.builder()
+                    .setValue(new BigDecimal("0.055"))
+                    .build();
+            DatedValue obs1 = DatedValue.builder().setDate(Date.of(2024, 1, 1)).setValue(new BigDecimal("0.055")).build();
```
chrisisla, line 257: Based upon the fact that your test is using a `DatedValue` array that is using sequential unique dates can I assume that this will be the norm i.e. no repeated dates as in the Step 6 list of dates?

chrisisla (commented):
A few comments to start off with...

rosetta-source/src/test/java/cdm/event/instructioncomposition/reset/functions/UpdateResetCompositionStateTest.java
SuperMonanana: No...this test is just for the high level function `UpdateResetCompositionState`, and it's just verifying that given an instruction with `calculateResetValue`, the state has been updated accordingly with whatever information in the instruction.

`Create_CalculateResetValueInstructionTest` has full coverage of the different observation patterns. And for multiple observation, if there are nonBusinessDates, there will be repeated dates with same rate as in the Step 6

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-type.rosetta
```diff
@@ -30,6 +38,10 @@ type ResetInstructionState: <"The definitive state object for the Reset process.
     adjustedCalculationPeriod CalculationPeriodBase (0..1) <"The calculation period after adjusting for non-business days and relevant financial calendars.">
     // Step 4 - AdjustDateInstruction
     adjustedResetDate date (0..1) <"The final, legally binding reset date after applying business day conventions.">
+    // Step 7 - RecordMarketRatesInstruction
+    resetValue Price (0..1) <"The final calculated rate or price derived from the observed market data.">
+    calculationMethod calculatedrate.CalculationMethodEnum (0..1) <"The specific compounding or averaging formula used to arrive at the final reset value.">
```
llynhiavu, line 43: is that standard rune syntax calculatedrate.Cal...?

(similar comment as Chris' in the .reset.func about namespace)

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
```diff
@@ -117,6 +123,130 @@ func Create_AdjustDateInstruction: <"This function generates the Instruction Com
...
+        else if singleResetRate exists
+        then resetRate
+    alias resetPrice: <"Builds the Price type consisting of the rounded reset value just calculated, its priceType, and the currency.">
+        Price {
+            value: roundedResetRate,
+            priceType: InterestRate,
+            unit:
+                UnitType {
+                    currency: currency,
+                    ...
```
llynhiavu, line 219: what's the meaning of the 3 dots? ...

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
SuperMonanana: Maybe @lagarciath can expand on this, but explicit path is necessary here because there is another `FloatingRateIndex` type used in this namespace. Without the prefix, it doesn't compile

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
SuperMonanana: Yes! I've added blank lines. I'll wait for @lagarciath to decide about splitting the function

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-type.rosetta
SuperMonanana: The reason is the same as the other case in .reset.func about namespace. There are two types of `CalculationMethodEnum` being used in this namespace, they have the same name but in different namespace. So we have to add the namespace prefix to one type to clarify, otherwise it won't compile

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: Yes, the reason is that this wasn't compiling without it

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: We are expecting to send as an input the list of dates from the previous step so let's assume we had [Aug 24, Aug 25, Aug 26 , Aug 26, Aug 26, Aug 29]. And the output is the same list with the observed market rates next to each. Meaning if the observed rates are:
- Aug 24 = 0.039
- Aug 25 = 0.040
- Aug 26 = 0.041
- Aug 29 = 0.037

The expected output of the function is:

- [date = Aug 24, rate = 0.039]
- [date = Aug 25, rate = 0.040]
- [date = Aug 26, rate = 0.041]
- [date = Aug 26, rate = 0.041]
- [date = Aug 26, rate = 0.041]
- [date = Aug 29, rate = 0.037]

This helper function is optional and provided to simplify internal rate lookups (as agreed in one of the initial calls we had in the TF). If used, the integrating system must supply the underlying lookup implementation. If skipped, the caller must manually map rates to dates to construct the required input format for the subsequent function.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: Why do we need weights? It is part of the formula extracted from the 2021 defs for applying the Compounding method.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: Sure! Let me work on a proposal and we will contribute the changes.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: The 3 dots mean "ignore all remaining fields". It tells the compiler that the UnitType struct has other fields besides currency, but you only care about populating the currency and want to explicitly ignore the rest without listing them all individually.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
chrisisla: Sorry, I'll be a bit clearer 😅 Why are they all set to 1? I couldn't see anywhere that updates them, is this as expected?

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
chrisisla: Thanks Laura 🙂

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
chrisisla: Ahhh, okay, so yes, we have `FloatingRateIndex` as a type and as a codelist:
 - cdm.base.staticdata.codelist
 - cdm.observable.asset

Not an issue for this PR to resolve but I know there are currently complications with code generators where we have types with the same name but in different namespaces. I might ask @hugohills-regnosys to chime on this.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-type.rosetta
chrisisla: What are the namespaces where `CalculationMethodEnum` is held please? I can see it in the cdm.observable.asset.calculatedrate namespace but nowhere else in the CDM (version 8.0.0-dev.5). 
It is in an fpml namespace though (fpml.consolidated.fpmlenum), is that the conflict?

SuperMonanana: Yes you are right it's in `fpml.consolidated.fpmlenum`, I just tried again without the `calculatedrate` here, it actually works (but FloatingRateIndex still requires explicit namespace prefix) I'll remove the prefix here

SuperMonanana: Fixed

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
SuperMonanana: Done, the function has been split based on observation type

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: Yes! The behavior comes down to how ISDA documentation defines the process versus how we implemented it in code. While the implementation differs slightly, the mathematical output is identical. Let me try to exaplain it:

**Standard ISDA logic:**
The standard compounding formula calculates the product of ( 1 + (rate * weight) / basis ), where the **weight** is the number of days a single rate applies. For example, if the rate observed on Aug 28 is 0.039 and it applies across 3 days, the weight is = 3.

**Our implementation:**
Rather than identifying a single rate and calculating its multi-day weight by subtracting dates, our input array explicitly expands the data to provide an individual (date, rate) pair for every single day (as we have been discussing in other comments).

Because every daily pair is passed explicitly, each observation naturally carries a weight of 1. In the code, the weights parameter is structured as an array of 1s with the exact same length as the number of observation rates. This allows us to perform vector multiplication in subsequent processes downstream.

Compounding three separate daily observations of 0.039 (weight = 1 each) yields the exact same mathematical result as compounding one observation of 0.039 with a weight of 3. We chose this approach because it keeps the code cleaner and enables vectorized calculations while avoiding date-range subtraction logic.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
llynhiavu: I have seen the split functions but don't remember where in github (got the link from one of the automated emails by chance). please paste here (or however you were planning to introduce it to us)

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
llynhiavu: so in the example, you assumed that Aug 26 is a Friday. ok.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
SuperMonanana: Hi @llynhiavu the split functions are already pushed to this PR, if you refresh and review these changes again, you should be able to see them. 

The main function of this step starts at https://github.com/finos/common-domain-model/pull/5128/changes#diff-822e5aca93ffacfc4159a640b063c3067a2574217784f0d6afe080408c8df4a5R180 and it calls the three helper functions for each observation type. The helper functions start at https://github.com/finos/common-domain-model/pull/5128/changes#diff-822e5aca93ffacfc4159a640b063c3067a2574217784f0d6afe080408c8df4a5R253

rosetta-source/src/test/java/cdm/base/math/functions/RepeatNumberTest.java
```diff
@@ -0,0 +1,49 @@
+package cdm.base.math.functions;
```
llynhiavu, line 1: (general question) please educate me, how do the different java functions bundled with the contribution get used?

SuperMonanana: All Java contributions in this PR are just tests that verify the Rune functions generate correct output and behave as expected. They call the generated Java code for the contributed functions and run automatically as part of the build process.

llynhiavu: ok thanks. it'd be nice to have a walkthrough when everything is in CDM 7, e.g. a recorded one that can be used as demo to implementers interested in this feature.

llynhiavu (approved):

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
chrisisla: So is this another reason why you create your date array with the replicated observation dates in order to keep the weights as 1? So rather than:
- [ date = Aug 28th, rate = 0.039, weight = 3] 

where 29th and 30th are a weekend so the weight is 3 as the rate on 28th is used for 3 days, you created:

- [ date = Aug 28th, rate = 0.039, weight = 1]
- [ date = Aug 28th, rate = 0.039, weight = 1]
- [ date = Aug 28th, rate = 0.039, weight = 1]

From what you described the approach that fits the ISDA documentation is the first one, so that implementation would probably be preferable (it makes more sense to me as the weight shows me that the rate is to be used for 3 days, whereas replicating the same date 3 times with a weight of 1 is less obvious). 

If the results of using your replicated array are the same then I have no problem with it, the description you have put above will be very useful to put into the documentation though.

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
```diff
@@ -172,6 +177,182 @@ func Create_DetermineUnadjustedObservationDatesInstruction: <"This function gene
...
+                    rounding -> precision,
+                    rounding -> roundingDirection,
+                    True
+                )
+        else RoundToPrecision(unroundedRate, 7, Nearest, True) // Default 7 decimal rounding for calculated rates
+
+func CalculateMultipleObservationResetRate: <"Calculates the unrounded reset value for Multiple Observation FROs using compounding methodology, then applies standard rounding.">
+    inputs:
+        observedRateDates DatedValue (1..*) <"The list of daily observed rates for the calculation period.">
+        dayCountFraction DayCountFractionEnum (0..1) <"The day count fraction applied to determine the year fraction per day.">
```
chrisisla, line 316: Does this need to be mandatory?

chrisisla (commented):
One quick question but otherwise I think this is looking good. That big function does look better now it has been split out 🙂

rosetta-source/src/main/rosetta/event-instructioncomposition-reset-func.rosetta
lagarciath: Should be, yes! We'll change it

PR #5150, opened 2026-09-09
Chris Rayner Product Model - Prevent mandatory inclusion of a Cash Settlement Centre in AssetPayout

### LINKED ISSUE #5071: AssetPayout Settlement Centre issues
### Issue Summary

In 7.0.0 several new conditions and attributes were added which now enforce the inclusion of `securitySettlementCentre` and `cashSettlementCentre`. There are two main problems with this:

**Forced to provide a `cashSettlementCentre` when there is no Cash**
For a securities lending trade the `AssetPayout` is used to represent the cashflows associated to the security on loan, not the cash - thus you should not be forced to enter a `cashSettlementCentre`. 
If a securities lending trade is against cash then the `Collateral` will be defined as a `TransferableProduct` that will include an `InterestRatePayout`. This is where the cash settlement details would be provided, not in the `AssetPayout`.
At the moment securities lending trades cannot be modelled correctly.

**Actual setting of `cashSettlementCentre` and `securitySettlementCentre` seems extranaous**
`SettlementBase` has conditions that force these two attributes to be set to "CashSettlementCentre" and "SecuritySettlementCentre". This does not seem to be required? 

These issues are preventing securities lending use cases as the payouts now are not correct.

### Steps to Reproduce

A securities lending Execution event now fails unless you specify the two settlement centre attributes:

Image

As described above, this is incorrect behaviour for a trade against non-cash (as there is no cash) and a trade against cash (as the `InterestRatePayout` under `collateral` is where the cash is represented, not in the `AssetPayout`).

Image

Image

Image

As can be seen from the condition in `SettlementBase` these attributes must always be set to enum values that are identical to the names of the attributes themselves which does not seem to be of any value.

### Expected Behaviour

Securities lending trades should be able to continue to use the `AssetPayout` for the security being lent only - i.e. this can specify a `securitySettlementCentre` but not a `cashSettlementCentre`.

The need to add `"securitySettlementCentre": "SecuritySettlementCentre"` as enforced by the conditions in `SettlementBase` also does not seem correct. 

My expectation would be that:
- if required for other use cases, `cashSettlementCentre` can be included in `AssetPayout` but must be optional
- the value of these new settlement centre attributes should be improved, as at the moment they are not helpful
- although a security will need a settlement centre, this may want to be optional as well to facilitate other use cases (although I think this attribute makes more sense, just the value that `SettlementBase` is forcing it to be set to does not).

### Actual Behaviour

Forced to add both a `securitySettlementCentre` and a `cashSettlementCentre` into `AssetPayout` in order to pass validation tests for an execution. This is invalid for both trades against cash and against non-cash.

Setting the value of these attributes to the same name as the attributes themselves also seems unhelpful. 

### Compatibility

Issue only observed post 7.0.0. This would be a backwards incompatible change to fix it. 

### Release

Next 7.x.x release as this is currently blocking all securities lending use cases.

### Additional Context

_No response_

regnosys-prod-user:


regnosys-prod-user:
# *Product Model - Prevent mandatory inclusion of a Cash Settlement Centre in AssetPayout*

_Background_

Securities lending trades use the `AssetPayout` to represent the cash flows for the securities being lent. Any cash being used as collateral is not represented here, thus it should not be mandatory to provide a Cash Settlement Centre.

_What is being released?_

The `SettlementCentreRequired` condition in `AssetPayout` has been updated to now only require the setting of a `securitySettlementCentre`.

_Review Directions_

Changes can be reviewed in PR: [#5150](https://github.com/finos/common-domain-model/pull/5150)

> [!NOTE]

regnosys-ops:
No Expectations generated for this PR. Make sure build was run with tests and has passed before merging.

PR #5180, opened 2026-09-17
CI/CD Security Update

Chuks-Ajeh:
Codefresh federates into GCP with workload identity federation and exports a short-lived `GCP_ACCESS_TOKEN`, replacing the long-lived `ARTIFACT_REGISTRY_SA_KEY` that `settings.xml` used as its Basic auth password. The maven-central mirror now authenticates as `oauth2accesstoken` with that token.

No pom change: this build publishes to OSSRH rather than to Artifact Registry, so it has no `artifactregistry://` URLs and no wagon extension. The credential was still doing real work, though — `settings.xml` mirrors all of Maven Central through the `maven-central` proxy, so every dependency resolved here is an authenticated read.

---

### Infrastructure - CI/CD Security Update

**What is being released?**

This is a security update to our Codefresh setup (Codefresh is the CI/CD platform used to build and publish releases). The build now signs in with a short-lived access key that expires after an hour, instead of a long-lived key that stayed valid indefinitely. If a credential is ever exposed, the window in which it is useful is now very small.

**Why is this being changed?**

The previous credential was a permanent key held as a build setting. It never expired, it was readable by anyone who could reach the build, and withdrawing access for one project meant replacing it everywhere at once. The replacement is issued fresh for each build, expires on its own, and can be granted or withdrawn per project.

**What changes?**

Only build configuration — how the build signs in when it downloads and publishes artefacts. There are no changes to the model, to any generated code, or to how published artefacts are consumed.

**Backwards compatibility**

Nothing is affected. There are no model, schema or API changes, so nothing downstream needs to change, and the published artefacts are the same as before.

**Review Directions**

Changes can be reviewed in PR: https://github.com/finos/common-domain-model/pull/5180 (branch `master`)


hugohills-regnosys:
# _Infrastructure - CI/CD Security Update_

_What is being released?_

This is a security update to our Codefresh setup (Codefresh is the CI/CD platform we use to build and publish releases). We've switched to using short-lived access keys instead of long-lived ones, so credentials expire quickly instead of staying valid indefinitely — reducing the impact if one is ever leaked.

_Review Directions_

Changes can be reviewed in PR: https://github.com/finos/common-domain-model/pull/5180

hugohills-regnosys (approved):

